PR #18, opened 2017-10-30
Increase Width of to to_char Template String for PID

jbylund:


lfittl:
Hi @jbylund - thanks for the contribution!

Looking over this, I'm actually wondering if you didn't run into another issue, since PIDs should say 5 digits on most systems I know.

In the latest published release (0.9.17) we update this logic, but didn't accommodate permission errors correctly.

This has been fixed in https://github.com/pganalyze/collector/commit/fded085a9cf9aa626ce42859cf52f47c9b313e4c

Could you check if the following query works for you correctly, when run as the pganalyze user:

```
SELECT (extract(epoch from COALESCE(backend_start, pg_postmaster_start_time()))::int::text || to_char(pid, 'FM00000'))::bigint,
  datid, datname, usesysid, usename, pid, application_name, client_addr::text, client_port,
  backend_start, xact_start, query_start, state_change, state, query
  FROM pg_stat_activity
  WHERE pid IS NOT NULL
```

(note that we also fixed a bug where this part of the collector is not using the `get_stat_activity` helper correctly, which also fixes the permission issue)

jbylund:
```
~ > ps -eo pid | sort -n | tail -n 1
126186
~ > cat /etc/issue
Ubuntu 16.04.3 LTS \n \l
```

```
third_party=# SELECT (extract(epoch from COALESCE(backend_start, pg_postmaster_start_time()))::int::text || to_char(pid, 'FM00000'))::bigint,
third_party-#   datid, datname, usesysid, usename, pid, application_name, client_addr::text, client_port,
third_party-#   backend_start, xact_start, query_start, state_change, state, query
third_party-#   FROM pg_stat_activity
third_party-#   WHERE pid IS NOT NULL;
ERROR:  invalid input syntax for integer: "1509391952#####"
```

Note: I had to loop for a while before I got my pid high enough to cause this crash.

lfittl:
@jbylund I stand corrected - thanks for looking that up!

Will merge shortly :)

PR #19, opened 2018-03-29
Update README.md

peletiah:
public.pg_stat_statements(showtext) does not exist in 9.3, added a hint to use public.pg_stat_statements() instead

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,10 @@ CREATE SCHEMA pganalyze;
 
 CREATE EXTENSION IF NOT EXISTS pg_stat_statements;
 
+/* HINT: Postgres 9.3 does not know the showtext-flag in public.pg_stat_statements, so you need to replace with with: */
+/* SELECT * FROM public.pg_stat_statements(); */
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,6 +53,10 @@ CREATE SCHEMA pganalyze;
 
 CREATE EXTENSION IF NOT EXISTS pg_stat_statements;
 
+/* HINT: Postgres 9.3 does not know the showtext-flag in public.pg_stat_statements, so you need to replace with with: */
+/* SELECT * FROM public.pg_stat_statements(); */
+/* in the following block, if you use 9.3 (The showtext-flag has been added in 9.4) */
```
lfittl: Thanks, good catch!

I think this may be better as a comment following the section (i.e. similar to how "If you are on Postgres 9.6 and use activity snapshots" is described), given 9.3 will be EOL from Postgres side this year.

Separately I'll apply a fix to the installation docs.

lfittl:
Re-worded and fixed by https://github.com/pganalyze/collector/commit/0e0e400c12d6a065a7caab81946172b6a29b499a

Thanks for the contribution!

PR #26, opened 2018-10-15
FIX: Incorrect specification of memory limit & restart event

Dombo:


contrib/systemd/pganalyze-collector.service
```diff
@@ -10,8 +10,8 @@ ProtectSystem=full
 ProtectHome=true
 CapabilityBoundingSet=
 NoNewPrivileges=true
-MemoryLimit=256MB
-Restart=Always
+MemoryLimit=256m
```
lfittl: Most documentation I can find has the suffixes uppercase - should we do the same? (i.e. `256M`)

PR #35, opened 2019-07-16
Fix ecs credentials

sj26:
I got an interesting error while trying to setup the pganalyze collector in an aws ecs service backed by fargate:

```
2019/07/16 13:24:33 E [default] Rds/Logs: Could not find RDS instance: NoCredentialProviders: no valid providers in chain
caused by: EnvAccessKeyNotFound: failed to find credentials in the environment.
SharedCredsLoad: failed to load profile, .
CredentialsEndpointError: failed to load credentials
caused by: RequestError: send request failed
caused by: Get http://169.254.170.2/v2/credentials/abcdef12-3456-789a-bcde-f1234567: Unencrypted connection is not permitted by pganalyze configuration
```

ECS tasks using a slightly different metadata endpoint which should probably also be whitelisted. I've pushed a custom docker image with this new code and it works great. Sorry, I didn't look too hard at tests yet. 😅

### DIFF
--- a/util/awsutil/amazon.go
+++ b/util/awsutil/amazon.go
@@ -14,7 +14,12 @@ func GetAwsSession(config config.ServerConfig) (*session.Session, error) {
 		creds = credentials.NewStaticCredentials(config.AwsAccessKeyID, config.AwsSecretAccessKey, "")
 	}
 
-	return session.NewSession(&aws.Config{Credentials: creds, Region: aws.String(config.AwsRegion), HTTPClient: config.HTTPClient})
+	return session.NewSession(&aws.Config{
+		Credentials:                   creds,
+		CredentialsChainVerboseErrors: aws.Bool(true),
+		Region:                        aws.String(config.AwsRegion),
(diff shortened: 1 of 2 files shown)


util/awsutil/amazon.go
```diff
@@ -14,7 +14,12 @@ func GetAwsSession(config config.ServerConfig) (*session.Session, error) {
 		creds = credentials.NewStaticCredentials(config.AwsAccessKeyID, config.AwsSecretAccessKey, "")
 	}
 
-	return session.NewSession(&aws.Config{Credentials: creds, Region: aws.String(config.AwsRegion), HTTPClient: config.HTTPClient})
+	return session.NewSession(&aws.Config{
+		Credentials:                   creds,
+		CredentialsChainVerboseErrors: aws.Bool(true),
```
lfittl: Just to verify, this outputs nothing in the case where there are no errors?

lfittl: Verified that this does not output anything in the case of no errors.

PR #38, opened 2019-12-29
Heroku: Use log line number for sorting log lines before processing

lfittl:
This avoids issues where Logplex sends out lines out-of-order, leading
to lines being truncated.


lfittl:
Fixed via #44

PR #48, opened 2020-05-24
Add support for Postgres 13

lfittl:


integration_test/Makefile
```diff
@@ -132,7 +132,7 @@ pg11:
 	diff -Naur pg11.expected pg11.out && echo 'success'
 
 pg12:
-	docker build -f Dockerfile.test-pg11 $(DOCKER_BUILD_OPTS)
+	docker build -f Dockerfile.test-pg12 $(DOCKER_BUILD_OPTS)
```
uhoh-itsmaciek, line 135: :grimacing:

integration_test/Makefile
```diff
@@ -148,3 +148,21 @@ pg12:
...
+
+pg13:
+	docker build -f Dockerfile.test-pg13 $(DOCKER_BUILD_OPTS)
+	$(DOCKER_RUN_CMD)
+	sleep 10
+	docker exec pganalyze-collector-test pgbench -U postgres -i
+	docker exec pganalyze-collector-test pgbench -U postgres
+	docker exec pganalyze-collector-test sh -c "DB_USERNAME=postgres PGA_API_KEY=dummy PGA_DISABLE_ACTIVITY=1 pganalyze-collector --dry-run --force-state-update -v" > pg13_1.json
+	docker exec pganalyze-collector-test pgbench -U postgres
+	docker exec pganalyze-collector-test sh -c "DB_USERNAME=postgres PGA_API_KEY=dummy PGA_DISABLE_ACTIVITY=1 pganalyze-collector --dry-run --force-state-update -v" > pg13_2.json
```
uhoh-itsmaciek, line 160: Why do we run the test twice?

lfittl: In order to get the query statistics - the first run only initializes the state file, and doesn't send any statistics.

Technically we could be doing `> /dev/null` for the first run, but if there are errors it's helpful to have that output around.

PR #61, opened 2020-07-12
Track relation parents as well as partition boundary (if any)

lfittl:
Depends on https://github.com/pganalyze/collector-snapshot/pull/1


input/postgres/relations.go
```diff
@@ -28,6 +30,8 @@ const relationsSQL string = `
 				c.reltoastrelid IS NOT NULL AS relation_has_toast,
 				c.relfrozenxid AS relation_frozen_xid,
 				%s,
+				COALESCE((SELECT inhparent FROM pg_inherits WHERE inhrelid = c.oid ORDER BY inhseqno LIMIT 1), 0) AS parent_relid,
```
uhoh-itsmaciek, line 33: I'm assuming you gave the multiple-parent situation some thought--does it not make sense to support that?

output/transform/postgres_relations.go
```diff
@@ -57,7 +73,7 @@ func transformPostgresRelations(s snapshot.FullSnapshot, newState state.Persiste
 				ForeignMatchType:  constraint.ForeignMatchType,
 			}
 			if constraint.ForeignOid != 0 {
-				sConstraint.ForeignRelationIdx = -1 // FIXME, need to look this up
+				sConstraint.ForeignRelationIdx = relationOidToIdx[constraint.ForeignOid]
```
uhoh-itsmaciek, line 76: 🎉

input/postgres/relations.go
lfittl: Yeah, its a good question - I thought some about it, and did it this way to keep things simple and performant on the pganalyze application side.

I think cascading hierarchies (i.e. having a parent that also has a parent) are much more common compared to having multiple inheritance. For one, multiple inheritance wouldn't be supported with declarative partitioning, i.e. its only a use case when you use the old-style table inheritance. We have one such use case in our own codebase, but that is for legacy reasons, and not because we actually use it.

Supporting multiple inheritance end-to-end (i.e. including in our own internal code) is certainly more expensive, and I think not warranted at this point, based on anecdotal evidence.

Therefore, I'll leave this as is for now - we can always adjust the data model here in the future if there is sufficient demand.

PR #62, opened 2020-07-16
Add ignore_schema_regexp

uhoh-itsmaciek:
This is like `ignore_table_pattern`, but pushed down into the actual
stats-gathering queries to improve performance. This should work much
better on very large schemas.

We use a regular expression instead of the current glob-like matching
since the former is natively supported in Postgres.

Open items:

 * [x] (Updated to `ignore_schema_regexp` per discussion) ~Confirm the setting name (right now we filter schema name, table name, and function name, so `ignore_table_regexp` is not super accurate)~
 * [x] Filter the Citus-specific stats in `relation_stats_ext.go`
 * [x] Settle on handling of vacuum progress and stats
 * [x] Test. I don't believe `ignore_table_pattern` is tested right now, but its implementation is much simpler--it'd be great if we could exercise the filter in integration tests.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -104,12 +104,12 @@ type ServerConfig struct {
 	// development and debugging. The value needs to be the name of the container.
 	LogDockerTail string `ini:"db_log_docker_tail"`
 
-	// **DEPRECATED**: Please use ignore_schema_regexp instead, since that uses an
-	// optimized code path in the collector and can avoid long-running queries.
-	//
 	// Specifies a table pattern to ignore - no statistics will be collected for
 	// tables that match the name. This uses Golang's filepath.Match function for
 	// comparison, so you can e.g. use "*" for wildcard matching.
+	//
+	// Deprecated: Please use ignore_schema_regexp instead, since that uses an
+	// optimized code path in the collector and can avoid long-running queries.
 	IgnoreTablePattern string `ini:"ignore_table_pattern"`
 
 	// Specifies a regular expression to ignore - no statistics will be collected for
--- a/config/config.go
+++ b/config/config.go
@@ -104,10 +104,18 @@ type ServerConfig struct {
 	// development and debugging. The value needs to be the name of the container.
 	LogDockerTail string `ini:"db_log_docker_tail"`
 
+	// **DEPRECATED**: Please use ignore_table_regexp instead, since that uses an
+	// optimized code path in the collector and can avoid long-running queries.
+	//
 	// Specifies a table pattern to ignore - no statistics will be collected for
 	// tables that match the name. This uses Golang's filepath.Match function for
 	// comparison, so you can e.g. use "*" for wildcard matching.
 	IgnoreTablePattern string `ini:"ignore_table_pattern"`
+
+	// Specifies a regular expression to ignore - no statistics will be collected for
+	// tables or schemas that match the name. E.g., to ignore tables that start with
+	// "ignored_", set this to "^ignored_".
+	IgnoreTableRegexp string `ini:"ignore_table_regexp"`
 
 	// Specifies the frequency of query statistics collection in seconds
 	//
--- a/config/config.go
+++ b/config/config.go
@@ -104,7 +104,7 @@ type ServerConfig struct {
 	// development and debugging. The value needs to be the name of the container.
 	LogDockerTail string `ini:"db_log_docker_tail"`
 
-	// **DEPRECATED**: Please use ignore_table_regexp instead, since that uses an
+	// **DEPRECATED**: Please use ignore_schema_regexp instead, since that uses an
 	// optimized code path in the collector and can avoid long-running queries.
 	//
 	// Specifies a table pattern to ignore - no statistics will be collected for
@@ -113,9 +113,12 @@ type ServerConfig struct {
 	IgnoreTablePattern string `ini:"ignore_table_pattern"`
 
 	// Specifies a regular expression to ignore - no statistics will be collected for
-	// tables or schemas that match the name. E.g., to ignore tables that start with
-	// "ignored_", set this to "^ignored_".
-	IgnoreTableRegexp string `ini:"ignore_table_regexp"`
+	// tables, views, functions, or schemas that match the name. Note that the match
+	// is applied to the '.'-joined concantenation of schema name and object name.
+	// E.g., to ignore tables that start with "ignored_", set this to "^ignored_". To
+	// ignore table "foo" only in the public schema, set to "^public\.foo$" (N.B.: you
+	// should escape the dot since that has special meaning in a regexp).
+	IgnoreSchemaRegexp string `ini:"ignore_schema_regexp"`
 
 	// Specifies the frequency of query statistics collection in seconds
 	//
--- a/input/postgres/buffercache.go
+++ b/input/postgres/buffercache.go
@@ -23,7 +23,7 @@ SELECT block_count * pg_catalog.current_setting('block_size')::int, d.datname, n
   JOIN pg_catalog.pg_database d ON (d.oid = reldatabase)
        LEFT JOIN pg_catalog.pg_class c ON (b.relfilenode = pg_catalog.pg_relation_filenode(c.oid) AND (b.reldatabase = 0 OR d.datname = pg_catalog.current_database()))
        LEFT JOIN pg_catalog.pg_namespace n ON (n.oid = c.relnamespace)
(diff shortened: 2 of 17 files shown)


integration_test/Makefile
```diff
@@ -155,14 +195,19 @@ pg13:
...
 	docker rmi pganalyze-collector-test
 	if [ "`jq '.collectorErrors' pg13_2.json | tr -d ' \n\r\t '`" != "null" ]; then \
 		jq '.collectorErrors' pg13_2.json; \
 		exit 1; \
 	fi
+	if jq '.relationReferences' pg13_2.json | grep --quiet pgbench_history; then \
+                echo "failed to ignore table according to IGNORE_SCHEMA_REGEXP=pgbench_history" \
+		jq '.relationReferences' pg13_2.json; \
+		exit 1; \
+	fi
```
lfittl, line 211: Neat, good way to test this :)

config/config.go
```diff
@@ -104,11 +104,22 @@ type ServerConfig struct {
 	// development and debugging. The value needs to be the name of the container.
 	LogDockerTail string `ini:"db_log_docker_tail"`
 
+	// **DEPRECATED**: Please use ignore_schema_regexp instead, since that uses an
+	// optimized code path in the collector and can avoid long-running queries.
+	//
```
lfittl, line 109: We could consider outputting a warning message during collector test runs that tells customers to upgrade to the new setting - otherwise its not very discoverable that the previous setting is deprecated.

uhoh-itsmaciek: Sure, I'll add that.

input/postgres/vacuum_progress.go
```diff
@@ -13,7 +13,18 @@ SELECT (EXTRACT(epoch FROM a.query_start)::int::text || pg_catalog.to_char(pid,
...
+				   (COALESCE(n.nspname,
+					  (SELECT pg_catalog.regexp_matches(query, 'autovacuum: VACUUM (ANALYZE )?([^\.]+).([^\(]+)( \(to prevent wraparound\))?'))[2])
+					  || '.' ||
+					  COALESCE(c.relname,
+					  (SELECT pg_catalog.regexp_matches(query, 'autovacuum: VACUUM (ANALYZE )?([^\.]+).([^\(]+)( \(to prevent wraparound\))?'))[3])) !~* $1)
+			   THEN COALESCE(c.relname,
+				   (SELECT pg_catalog.regexp_matches(query, 'autovacuum: VACUUM (ANALYZE )?([^\.]+).([^\(]+)( \(to prevent wraparound\))?'))[3])
+			   ELSE
+				   ''
+		   END AS relname,
```
uhoh-itsmaciek, line 27: Unfortunately this one gets pretty gnarly. I can try rewriting with a CTE to avoid duplication, or we can cross our fingers that we don't have to touch this much. Thoughts?

input/postgres/relation_stats.go
```diff
@@ -65,10 +66,11 @@ SELECT s.indexrelid,
 			 COALESCE(sio.idx_blks_hit, 0)
 	FROM pg_catalog.pg_stat_user_indexes s
 			 LEFT JOIN pg_catalog.pg_statio_user_indexes sio USING (indexrelid)
-	WHERE s.indexrelid NOT IN (SELECT relid FROM locked_relids)
+ WHERE s.indexrelid NOT IN (SELECT relid FROM locked_relids)
+			 AND ($1 = '' OR (schemaname || '.' || relname) !~* $1)
```
lfittl, line 70:
```suggestion
			 AND ($1 = '' OR (s.schemaname || '.' || s.relname) !~* $1)
```

Looks like this needs to be schema qualified, since both `pg_stat_user_indexes` and `pg_statio_user_indexes` have these columns

input/postgres/relation_stats_ext.go
```diff
@@ -10,10 +10,13 @@ import (
...
        pg_catalog.citus_table_size(logicalrelid)
-  FROM pg_catalog.pg_dist_partition
+	FROM pg_catalog.pg_dist_partition dp
+			 INNER JOIN pg_catalog.pg_class c ON (dp.logicalrelid::oid == c.oid)
+			 INNER JOIN pg_catalog.pg_namespace n ON (c.relnamespace = n.oid)
+ WHERE ($1 = '' OR (n.nspname || '.' || c.relname) !~* $1)
 `
 
-func handleRelationStatsExt(db *sql.DB, relStats state.PostgresRelationStatsMap, postgresVersion state.PostgresVersion) (state.PostgresRelationStatsMap, error) {
+func handleRelationStatsExt(db *sql.DB, relStats state.PostgresRelationStatsMap, postgresVersion state.PostgresVersion, ignoreRegexp string) (state.PostgresRelationStatsMap, error) {
```
lfittl, line 19: I think this `ignoreRegexp` needs to be passed to the stmt.Query() call at line 27

input/postgres/buffercache.go
```diff
@@ -23,6 +23,7 @@ SELECT block_count * pg_catalog.current_setting('block_size')::int, d.datname, n
   JOIN pg_catalog.pg_database d ON (d.oid = reldatabase)
        LEFT JOIN pg_catalog.pg_class c ON (b.relfilenode = pg_catalog.pg_relation_filenode(c.oid) AND (b.reldatabase = 0 OR d.datname = pg_catalog.current_database()))
        LEFT JOIN pg_catalog.pg_namespace n ON (n.oid = c.relnamespace)
+ WHERE ($1 = '' OR (coalesce(n.nspname, '') || '.' || coalesce(c.relname, '')) !~* $1)
```
lfittl, line 26: Generally I would uppercase `COALESCE`, since thats how its done in the Postgres docs: https://www.postgresql.org/docs/12/functions-conditional.html#FUNCTIONS-COALESCE-NVL-IFNULL

input/postgres/vacuum_progress.go
lfittl: Hmm, yeah, this becomes very difficult to read due to the regexp. Doing this in a CTE would certainly read easier.

Since we are dealing with a low volume scenario (i.e. there is at most a handful of vacuums running), we could also do the filtering in the Go code, if we felt that was easier to do - after all, one motivation for picking regexp as a format was that its (mostly) portable between Postgres, Go and other languages.

uhoh-itsmaciek: I think I'd like to stick to SQL where possible just for consistency. Let me try refactoring this.

input/postgres/vacuum_progress.go
```diff
@@ -13,7 +13,18 @@ SELECT (EXTRACT(epoch FROM a.query_start)::int::text || pg_catalog.to_char(pid,
 			 (EXTRACT(epoch FROM COALESCE(backend_start, pg_catalog.pg_postmaster_start_time()))::int::text || pg_catalog.to_char(pid, 'FM0000000'))::bigint AS backend_identity,
 			 a.datname,
 			 (SELECT pg_catalog.regexp_matches(query, 'autovacuum: VACUUM (ANALYZE )?([^\.]+).([^\(]+)( \(to prevent wraparound\))?'))[2] AS nspname,
-			 (SELECT pg_catalog.regexp_matches(query, 'autovacuum: VACUUM (ANALYZE )?([^\.]+).([^\(]+)( \(to prevent wraparound\))?'))[3] AS relname,
```
uhoh-itsmaciek, line 16: Is it intentional that this does *not* coalesce `nspname` like we do [here](https://github.com/pganalyze/collector/pull/62/files#diff-1f9a634167b0b4f0ff14633f304b9fb6R46)?

uhoh-itsmaciek: Ah yes, because we don't have that table.

input/postgres/vacuum_progress.go
uhoh-itsmaciek: Refactored and I think this is much more readable. Good call :+1:.

input/postgres/vacuum_progress.go
```diff
@@ -70,7 +102,7 @@ func GetVacuumProgress(logger *util.Logger, db *sql.DB, postgresVersion state.Po
 		} else {
 			vacuumSourceTable = "pg_catalog.pg_stat_progress_vacuum"
 		}
-		sql = fmt.Sprintf(vacuumProgressSQLDefault, vacuumSourceTable, activitySourceTable)
+		sql = fmt.Sprintf(vacuumProgressSQLDefault, activitySourceTable, vacuumSourceTable)
```
uhoh-itsmaciek, line 105: Note that because of the CTE, these switched places.

PR #64, opened 2020-07-29
Warn on too many tables

uhoh-itsmaciek:


input/postgres/schema.go
```diff
@@ -56,6 +56,9 @@ func collectSchemaData(collectionOpts state.CollectionOpts, logger *util.Logger,
 			logger.PrintWarning("Skipping table/index data for database \"%s\", due to error: %s", databaseName, err)
 			return ps
 		}
+		if relCount := len(newRelations); relCount > 5000 {
+			logger.PrintWarning("Too many tables: got %d for database \"%s\", but only 5000 can be monitored; use ignore_schema_regexp config setting to filter", relCount, databaseName)
+		}
```
lfittl, line 61: Technically this limit is for the whole server, not per database - maybe we can move this check to happen after line 47 in this file? (and checking length of `ps.Relations`)

PR #65, opened 2020-07-31
Add partitioned table partitioning columns

uhoh-itsmaciek:


input/postgres/relations.go
```diff
@@ -32,6 +38,9 @@ const relationsSQL string = `
 				%s,
 				COALESCE((SELECT inhparent FROM pg_inherits WHERE inhrelid = c.oid ORDER BY inhseqno LIMIT 1), 0) AS parent_relid,
 				%s,
+				%s,
+				%s,
+				%s,
```
uhoh-itsmaciek, line 43: Should we keep all the fields together to avoid the interpolation proliferation? I saw that's done in some places but not others...

lfittl: I think this is okay for now, but open to adjusting as well if you think there is a better way to structure it.

output/transform/postgres_relations.go
```diff
@@ -30,6 +30,18 @@ func transformPostgresRelations(s snapshot.FullSnapshot, newState state.Persiste
 			hasParentRelation = true
 		}
 
+		var partStrat snapshot.RelationInformation_PartitionStrategy
+		switch relation.PartitionedBy {
```
lfittl, line 34: I think this should be `relation.PartitionStrategy`

uhoh-itsmaciek: Good catch

PR #67, opened 2020-08-11
Add basic CI via GitHub Actions

uhoh-itsmaciek:
@lfittl thoughts? This is really just a smoke test, but it's a start. Anything else on the shortlist for adding to CI?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Set up Go 1.x
       uses: actions/setup-go@v2
       with:
-        go-version: ^1.13
+        go-version: 1.14.3
       id: go
 
     - name: Set up protoc
@@ -32,4 +32,7 @@ jobs:
       run: git submodule update --init
 
     - name: Runs tests
-      run: make
+      run: |
+        make build OUTFILE=pganalyze-collector-linux-amd64
+        make test
+        make integration_test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,5 +28,8 @@ jobs:
     - name: Check out code
       uses: actions/checkout@v2
 
+    - name: Initialize submodules
+      run: git submodule update --init
+
     - name: Runs tests
       run: make
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,8 +23,7 @@ jobs:
     - name: Set up protoc
       uses: arduino/setup-protoc@v1.1.0
       with:
-        # Version to use. Example: 3.9.1
-        version: ^3.12
+        version: 3.x
 
     - name: Check out code
       uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,12 @@ jobs:
         go-version: ^1.13
       id: go
 
+    - name: Set up protoc
+      uses: arduino/setup-protoc@v1.1.0
+      with:
+        # Version to use. Example: 3.9.1
+        version: ^3.12
+
     - name: Check out code
       uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


uhoh-itsmaciek:
Ah, I can see if i can get protoc working, and that the generated files from the protoc run actually match what's checked in.

uhoh-itsmaciek:
Failing now with

```
mkdir -p /home/runner/work/collector/collector/bin
GOBIN=/home/runner/work/collector/collector/bin go install github.com/golang/protobuf/protoc-gen-go
PATH=/home/runner/work/collector/collector/bin:/home/runner/go/bin:/opt/hostedtoolcache/protoc/3.12.4/x64/bin:/opt/hostedtoolcache/go/1.14.6/x64/bin:/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:/usr/share/rust/.cargo/bin:/home/runner/.config/composer/vendor/bin:/home/runner/.dotnet/tools:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games protoc --go_out=Mgoogle/protobuf/timestamp.proto=github.com/golang/protobuf/ptypes/timestamp:output/pganalyze_collector -I protobuf  
Missing input file.
make: *** [output/pganalyze_collector/snapshot.pb.go] Error 1
Makefile:37: recipe for target 'output/pganalyze_collector/snapshot.pb.go' failed
##[error]Process completed with exit code 2.
```
Any ideas?

uhoh-itsmaciek:
Ah it looks like `$(PROTOBUF_FILES)` is empty--I think I need to `git submodule update`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
...
+        version: 3.x
+
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Initialize submodules
+      run: git submodule update --init
+
+    - name: Runs tests
+      run: make
```
lfittl, line 35: I think we should also run `make integration_test` here (as we used to do on Travis)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.13
```
lfittl, line 20: I think we should specify 1.14 here, since thats what we use across the board (soon to be updated to 1.15 I expect)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
+name: CI
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
lfittl, line 9: Is that correct? (does this mean PRs that are intended to be merged into `master`?)

uhoh-itsmaciek: I think so. [The docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags) are not super clear, but that seems to be what's happening (based on the fact that the action *is* executing on this PR). Or rather, strictly speaking, pull requests against the master branch.

.github/workflows/ci.yml
uhoh-itsmaciek: I didn't realize we used to have a Travis setup. I'll add `make integration_test`--we might need additional Docker setup for that, but there should be existing actions to support that.

I also see a `deploy` configuration in `.travis.yml`--is that something we should use actions for as well?

lfittl: Yeah, we should also move the deploy configuration, but lets do that separately - its what's responsible for creating a binary in the releases list.

PR #68, opened 2020-08-12
Enable log explain for all databases

uhoh-itsmaciek:
Totally untested so far but it builds and doesn't break existing tests.

Fixes #910

### DIFF
--- a/input/postgres/explain.go
+++ b/input/postgres/explain.go
@@ -26,12 +26,15 @@ func RunExplain(server state.Server, inputs []state.PostgresQuerySample, collect
 	for dbName, dbSamples := range samplesByDb {
 		db, err := EstablishConnection(server, logger, collectionOpts, dbName)
 
-		if err == nil {
-			dbOutputs := runDbExplain(db, dbSamples)
-			outputs = append(outputs, dbOutputs...)
-
-			db.Close()
+		if err != nil {
+			logger.PrintVerbose("Could not connect to %s to run explain: %s; skipping", dbName, err)
+			continue
 		}
+
+		dbOutputs := runDbExplain(db, dbSamples)
+		db.Close()
+
+		outputs = append(outputs, dbOutputs...)
 	}
 	return
 }
--- a/input/postgres/explain.go
+++ b/input/postgres/explain.go
@@ -17,7 +17,8 @@ func RunExplain(server state.Server, inputs []state.PostgresQuerySample, collect
 	var samplesByDb = make(map[string]([]state.PostgresQuerySample))
 
 	skip := func(sample state.PostgresQuerySample) bool {
-		monitoredDb := sample.Database == server.Config.GetDbName() || server.Config.DbAllNames || contains(server.Config.DbExtraNames, sample.Database)
+		monitoredDb := sample.Database == "" || sample.Database == server.Config.GetDbName() ||
+			server.Config.DbAllNames || contains(server.Config.DbExtraNames, sample.Database)
 
 		return !monitoredDb ||
 			// EXPLAIN was already collected, e.g. from auto_explain
--- a/input/postgres/explain.go
+++ b/input/postgres/explain.go
@@ -10,9 +10,33 @@ import (
 	"github.com/lib/pq"
 	"github.com/pganalyze/collector/output/pganalyze_collector"
 	"github.com/pganalyze/collector/state"
+	"github.com/pganalyze/collector/util"
 )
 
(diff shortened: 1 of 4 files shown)


input/postgres/explain.go
```diff
@@ -77,3 +95,12 @@ func RunExplain(db *sql.DB, connectedDbName string, inputs []state.PostgresQuery
 
 	return
 }
+
+func contains(strs []string, val string) bool {
```
uhoh-itsmaciek, line 99: Is there a built-in way to do this that I missed? I expected `strings` maybe to have something but I didn't see anything.

input/postgres/explain.go
```diff
@@ -10,9 +10,33 @@ import (
...
+	}
+
+	for dbName, dbSamples := range samplesByDb {
+		db, err := EstablishConnection(server, logger, collectionOpts, dbName)
+
+		if err == nil {
+			dbOutputs := runDbExplain(db, dbSamples)
+			outputs = append(outputs, dbOutputs...)
+
+			db.Close()
```
uhoh-itsmaciek, line 33: I did this because that's what `stream.go` was doing, but we should be calling `db.Close` regardless of whether or not there is an error, no? Both the `database/sql` and the `lib/pq` docs are vague about this, but if memory serves, `pq` could certainly fail in setup after actually connecting...

lfittl: I think thats a good idea to explicitly close - especially since we enforce a connection limit of 5 connections by the pganalyze collector across the server. Which means if we keep these open and there is multiple DBs, we would quickly exhaust that limit.

uhoh-itsmaciek: Oh sure, I agree it should definitely be closed explicitly, I'm just saying I think we should try to close it explicitly regardless of whether or not `EstablishConnection` returns an error (i.e., just because `EstablishConnection` returned an error, that doesn't mean the connection was not at least partly established).

uhoh-itsmaciek: Also, looking at this again, in [runner/logs.go](https://github.com/pganalyze/collector/blob/master/runner/logs.go#L37-L44), we log and bail if we can't establish a connection--we should probably log a warning here at least, right?

uhoh-itsmaciek: (and the connection argument to `DownloadLogs` is no longer necessary--I'll remove it)

uhoh-itsmaciek: >I'm just saying I think we should try to close it explicitly regardless of whether or not EstablishConnection returns an error

I realize now I'm wrong about this because EstablishConnection explicitly returns `nil` for the connection if `err` is set (which may not be the right approach, but that's a separate issue).

uhoh-itsmaciek:
Tested in a live self-managed setup and seems to work.

input/postgres/explain.go
```diff
@@ -10,9 +10,36 @@ import (
...
 
-func RunExplain(db *sql.DB, connectedDbName string, inputs []state.PostgresQuerySample) (outputs []state.PostgresQuerySample) {
+func RunExplain(server state.Server, inputs []state.PostgresQuerySample, collectionOpts state.CollectionOpts, logger *util.Logger) (outputs []state.PostgresQuerySample) {
+	var samplesByDb = make(map[string]([]state.PostgresQuerySample))
+
+	for _, sample := range inputs {
+		if sample.Database != server.Config.GetDbName() && !contains(server.Config.DbExtraNames, sample.Database) && !server.Config.DbAllNames {
+			continue
+		}
+		samplesByDb[sample.Database] = append(samplesByDb[sample.Database], sample)
```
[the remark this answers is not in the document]
uhoh-itsmaciek: Good call. I'm pulling up all the checks except the parse. It's awkward to do that without re-parsing, and I doubt that's worth the overhead or common enough to really make a difference.

input/postgres/explain.go
```diff
@@ -16,8 +16,21 @@ import (
 func RunExplain(server state.Server, inputs []state.PostgresQuerySample, collectionOpts state.CollectionOpts, logger *util.Logger) (outputs []state.PostgresQuerySample) {
 	var samplesByDb = make(map[string]([]state.PostgresQuerySample))
 
+	skip := func(sample state.PostgresQuerySample) bool {
+		monitoredDb := sample.Database == server.Config.GetDbName() || server.Config.DbAllNames || contains(server.Config.DbExtraNames, sample.Database)
```
uhoh-itsmaciek, line 20: @lfittl hmm, from our conversation earlier, if `sample.Database` is the empty string, this would be `false`, no? And then we'd skip based on the first condition below?

lfittl: Good point. I think we should add `sample.Database == "" || ...` here

PR #71, opened 2020-08-14
EXPLAIN plans / Query samples: Support log line prefix without %d and %u

lfittl:
Whilst not recommended, in some scenarios changing the log_line_prefix
is difficult, and we want to make it easy to get EXPLAIN data even in
those scenarios.

In case the log_line_prefix is missing the database (%d) and/or the user
(%u), we simply use the user and database of the collector connection.

### DIFF
--- a/output/transform/logs.go
+++ b/output/transform/logs.go
@@ -9,10 +9,10 @@ import (
 	uuid "github.com/satori/go.uuid"
 )
 
-func LogStateToLogSnapshot(logState state.TransientLogState) (snapshot.CompactLogSnapshot, snapshot.CompactSnapshot_BaseRefs) {
+func LogStateToLogSnapshot(server state.Server, logState state.TransientLogState) (snapshot.CompactLogSnapshot, snapshot.CompactSnapshot_BaseRefs) {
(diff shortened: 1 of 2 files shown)


output/transform/logs.go
```diff
@@ -62,14 +62,22 @@ func upsertRelationReference(refs []*snapshot.RelationReference, databaseIdx int
...
 			continue
 		}
 
+		if sampleIn.Username == "" {
+			sampleIn.Username = server.Config.GetDbUsername()
+		}
+
+		if sampleIn.Database == "" {
+			sampleIn.Database = server.Config.GetDbName()
+		}
```
uhoh-itsmaciek, line 79: Should we warn in these situations? I think this is a great idea in general, since it'll allow more configurations and it's likely to be correct, but it *could* be wrong. Should we point that out? (Though I guess the trick would be to point that out once and not for every sample.)

lfittl: As discussed, since that warning wouldn't show during a test run (since we don't really exercise this code in that situation), I would not warn - but rather have a check up or something in-app that tells you about a log_line_prefix thats missing %u and %d.

PR #74, opened 2020-08-19
Use helper function for log-based EXPLAIN access if available

uhoh-itsmaciek:
This works with the big caveat that more queries may contain semicolons than we expected: I was testing by submitting queries in psql, and things there need to end in a semicolon. That's probably not a common use case, but a lot of larger heredoc-style queries embedded in apps often end in semicolons out of habit. Should we strip trailing semicolons? Trailing semicolons and whitespace? ~Or is that all stripped out via `pg_query.Parse`?~ I realize it's *not* stripped out because that's exactly what I ran into it. I'll try to regexp_replace it.

This will also need docs, but this part should probably come first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,47 +90,6 @@ but from here onwards you can use the `pganalyze` user instead.
 The collector will automatically use the helper methods
 if they exist in the `pganalyze` schema - otherwise data will be fetched directly.
 
-If you use `enable_log_explain`, create the pganalyze schema and this function on each
-database where EXPLAIN should run:
-
-```
-CREATE OR REPLACE FUNCTION pganalyze.explain(query text, params text[]) RETURNS text AS
-$$
-DECLARE
-  prepared_query text;
-  prepared_params text;
-  result text;
-BEGIN
-  SELECT regexp_replace(query, ';+\s*\Z', '') INTO prepared_query;
-  IF prepared_query LIKE '%;%' THEN
-    RAISE EXCEPTION 'cannot run EXPLAIN when query contains semicolon';
-  END IF;
-
-  IF array_length(params, 1) > 0 THEN
-    SELECT string_agg(quote_literal(param) || '::unknown', ',') FROM unnest(params) p(param) INTO prepared_params;
-
-    EXECUTE 'PREPARE pganalyze_explain AS ' || prepared_query;
-    BEGIN
-      EXECUTE 'EXPLAIN (VERBOSE, FORMAT JSON) EXECUTE pganalyze_explain(' || prepared_params || ')' INTO STRICT result;
-    EXCEPTION WHEN OTHERS THEN
-      DEALLOCATE pganalyze_explain;
-      RAISE;
-    END;
-    DEALLOCATE pganalyze_explain;
-  ELSE
-    EXECUTE 'EXPLAIN (VERBOSE, FORMAT JSON) ' || prepared_query INTO STRICT result;
-  END IF;
-
-  RETURN result;
-END
-$$ LANGUAGE plpgsql VOLATILE SECURITY DEFINER;
-```
-
-Note that this function contains a check for semicolons in the query
-text. This is to minimize collector access to your data: it ensures
-that the collector cannot piggyback other queries that could
-exfiltrate data.
-
 If you are on Postgres 9.6 and use activity snapshots:
 
 ```
@@ -196,6 +155,55 @@ $$
   /* pganalyze-collector */ SELECT * FROM public.pg_stat_statements_reset();
 $$ LANGUAGE sql VOLATILE SECURITY DEFINER;
 ```
+
+Setting up Log EXPLAIN helper
+-----------------------------
+
+This section is relevant only if you use the `enable_log_explain`
+setting; if you use the recommended autoexplain extension, or if you
+do not plan to use EXPLAIN plan collection, you can skip it.
+
+If you use `enable_log_explain`, use a superuser connection to create
+the pganalyze schema and this function on each database where EXPLAIN
+should run:
+
+```
+CREATE OR REPLACE FUNCTION pganalyze.explain(query text, params text[]) RETURNS text AS
+$$
+DECLARE
+  prepared_query text;
+  prepared_params text;
+  result text;
+BEGIN
+  SELECT regexp_replace(query, ';+\s*\Z', '') INTO prepared_query;
+  IF prepared_query LIKE '%;%' THEN
+    RAISE EXCEPTION 'cannot run EXPLAIN when query contains semicolon';
+  END IF;
+
+  IF array_length(params, 1) > 0 THEN
+    SELECT string_agg(quote_literal(param) || '::unknown', ',') FROM unnest(params) p(param) INTO prepared_params;
+
+    EXECUTE 'PREPARE pganalyze_explain AS ' || prepared_query;
+    BEGIN
+      EXECUTE 'EXPLAIN (VERBOSE, FORMAT JSON) EXECUTE pganalyze_explain(' || prepared_params || ')' INTO STRICT result;
+    EXCEPTION WHEN OTHERS THEN
+      DEALLOCATE pganalyze_explain;
+      RAISE;
+    END;
+    DEALLOCATE pganalyze_explain;
+  ELSE
+    EXECUTE 'EXPLAIN (VERBOSE, FORMAT JSON) ' || prepared_query INTO STRICT result;
+  END IF;
+
+  RETURN result;
+END
+$$ LANGUAGE plpgsql VOLATILE SECURITY DEFINER;
+```
+
+Note that this function contains a check for semicolons in the query
+text. This is to minimize collector access to your data: it ensures
+that the collector cannot piggyback other queries that could
+exfiltrate data.
 
 
 Example output
--- a/README.md
+++ b/README.md
@@ -90,6 +90,33 @@ but from here onwards you can use the `pganalyze` user instead.
 The collector will automatically use the helper methods
 if they exist in the `pganalyze` schema - otherwise data will be fetched directly.
 
+If you use `enable_log_explain`:
+
+```
+CREATE OR REPLACE FUNCTION pganalyze.explain(query text, params text) RETURNS text AS
+$$
+DECLARE
+  result text;
+BEGIN
+  -- this is a check to minimize collector access to your data: it ensures that the
+  -- collector cannot piggyback other queries that could exfiltrate data
+  IF query LIKE '%;%' OR params LIKE '%;%' THEN
+    RAISE EXCEPTION 'cannot run EXPLAIN when query or parameters contain semicolon';
+  END IF;
+
+  IF length(params) > 0 THEN
+    EXECUTE '/* pganalyze-collector */ PREPARE pganalyze_explain AS ' || query;
+    EXECUTE '/* pganalyze-collector */ EXPLAIN (VERBOSE, FORMAT JSON) EXECUTE pganalyze_explain(' || params || ')' INTO STRICT result;
+    /* pganalyze-collector */ DEALLOCATE pganalyze_explain;
+  ELSE
+    EXECUTE '/* pganalyze-collector */ EXPLAIN (VERBOSE, FORMAT JSON) ' || query INTO STRICT result;
+  END IF;
+
+  RETURN result;
+END
+$$ LANGUAGE plpgsql VOLATILE SECURITY DEFINER;
+```
+
 If you are on Postgres 9.6 and use activity snapshots:
 
 ```
--- a/README.md
+++ b/README.md
@@ -90,25 +90,27 @@ but from here onwards you can use the `pganalyze` user instead.
 The collector will automatically use the helper methods
 if they exist in the `pganalyze` schema - otherwise data will be fetched directly.
 
-If you use `enable_log_explain`:
+If you use `enable_log_explain`, create the pganalyze schema and this function on each
+database where EXPLAIN should run:
 
 ```
-CREATE OR REPLACE FUNCTION pganalyze.explain(query text, params text) RETURNS text AS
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -90,6 +90,33 @@ but from here onwards you can use the `pganalyze` user instead.
...
+```
+CREATE OR REPLACE FUNCTION pganalyze.explain(query text, params text) RETURNS text AS
+$$
+DECLARE
+  result text;
+BEGIN
+  -- this is a check to minimize collector access to your data: it ensures that the
+  -- collector cannot piggyback other queries that could exfiltrate data
+  IF query LIKE '%;%' OR params LIKE '%;%' THEN
+    RAISE EXCEPTION 'cannot run EXPLAIN when query or parameters contain semicolon';
```
uhoh-itsmaciek, line 104: We may want to special-case this error in the UI like we are special-casing permissions issues.

README.md
```diff
@@ -90,6 +90,35 @@ but from here onwards you can use the `pganalyze` user instead.
...
+  SELECT regexp_replace(query, ';+\s*\Z', '') INTO prepared_query;
+  -- this is a check to minimize collector access to your data: it ensures that the
+  -- collector cannot piggyback other queries that could exfiltrate data
+  IF prepared_query LIKE '%;%' OR params LIKE '%;%' THEN
+    RAISE EXCEPTION 'cannot run EXPLAIN when query or parameters contain semicolon';
+  END IF;
+
+  IF length(params) > 0 THEN
+    EXECUTE 'PREPARE pganalyze_explain AS ' || prepared_query;
+    EXECUTE 'EXPLAIN (VERBOSE, FORMAT JSON) EXECUTE pganalyze_explain(' || params || ')' INTO STRICT result;
```
lfittl, line 111: I've been thinking if there is a way to avoid checking the `params` value for semicolons, because thats where the restriction is going to be most annoying (assuming you're using prepared statements, the query text itself shouldn't contain semicolons - but as discussed a string used in the query very well might).

I think there may be a way to achieve this, if we run `quote_ident` inside this function, and accept a variadic parameter list (instead of doing the escaping on the Go clientside). Its a bit more PL/pgSQL code, but would allow us to lift the restriction on checking semicolons in `params`. What do you think?

input/postgres/explain.go
```diff
@@ -44,40 +44,54 @@ func RunExplain(server state.Server, inputs []state.PostgresQuerySample, collect
 			logger.PrintVerbose("Could not connect to %s to run explain: %s; skipping", dbName, err)
 			continue
 		}
+		useHelper := statsHelperExists(db, "explain")
+		if useHelper {
+			logger.PrintVerbose("Found pganalyze.explain() stats helper")
+		} else {
+			logger.PrintInfo("Warning: pganalyze.explain() helper function not found. Please set up" +
```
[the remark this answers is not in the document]
lfittl: +1 maybe we should not even attempt to run the EXPLAIN if the helper doesn't exist and we're not superuser? (e.g. always return an error in the query explain error text, but not causing an unnecessary database log entry saying we couldn't run the EXPLAIN)

README.md
```diff
@@ -90,6 +90,35 @@ but from here onwards you can use the `pganalyze` user instead.
...
+```
+CREATE OR REPLACE FUNCTION pganalyze.explain(query text, params text) RETURNS text AS
+$$
+DECLARE
+  prepared_query text;
+  result text;
+BEGIN
+  SELECT regexp_replace(query, ';+\s*\Z', '') INTO prepared_query;
+  -- this is a check to minimize collector access to your data: it ensures that the
+  -- collector cannot piggyback other queries that could exfiltrate data
```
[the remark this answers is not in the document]
uhoh-itsmaciek: Yeah, the latter is exactly what I was thinking. But maybe we should still move it out. That makes it less clear what it's for and users may be less likely to change that, but maybe that's fine: you shouldn't be bypassing this unless you know what you're doing.

README.md
uhoh-itsmaciek: I was thinking about that, too, and had convinced myself that we couldn't do that because we would be forcing some bogus type information (i.e., claiming the params are all `text`). I've definitely run into issues in the past where `unknown` goes through implicit casts more easily more easily than text.

But I realized we will run into issues in similar cases even with the quote-in-collector approach, and we could explicitly cast these parameters to `unknown` in the new approach to help avoid that as much as possible. We don't know the actual parameter types in our query samples, right?

I think we'll still run into some issues, but this should be rare enough that we can treat it as an edge case:

```
maciek=# prepare add(integer, integer) as select $1 + $2;
PREPARE
maciek=# execute add(1,2);
 ?column?
----------
        3
(1 row)

maciek=# select pganalyze.explain('select $1 + $2', '''1'', ''2''');
ERROR:  operator is not unique: unknown + unknown
LINE 1: PREPARE pganalyze_explain AS select $1 + $2
                                               ^
HINT:  Could not choose a best candidate operator. You might need to add explicit type casts.
QUERY:  PREPARE pganalyze_explain AS select $1 + $2
CONTEXT:  PL/pgSQL function pganalyze.explain(text,text) line 14 at EXECUTE
```

(or with the updated param-array explain helper):

```
maciek=# select pganalyze.explain('select $1 + $2', array['1', '2']::text[]);
ERROR:  operator is not unique: unknown + unknown
LINE 1: PREPARE pganalyze_explain AS select $1 + $2
                                               ^
HINT:  Could not choose a best candidate operator. You might need to add explicit type casts.
QUERY:  PREPARE pganalyze_explain AS select $1 + $2
CONTEXT:  PL/pgSQL function pganalyze.explain(text,text[]) line 15 at EXECUTE
```
I don't think there's really much we can do about these cases. We should probably catch "Could not choose a best candidate operator" cases and special-case the UI error as something like "Sorry, this explain doesn't work in log-based explain right now; try autoexplain if possible".

Also, I was thinking about other situations that could break log-based EXPLAIN, and I think `search_path` is a big one. Not sure there's an easy fix (at least for complicated cases), but curious if you've given that any thought.

input/postgres/explain.go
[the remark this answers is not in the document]
uhoh-itsmaciek: I kinda feel like granting the pganalyze user SELECT permissions directly is a reasonable alternative approach: it's pretty clear that you're giving us read access in that case, so if that's what you want to do, I don't see a reason to get in your way. Thoughts?

README.md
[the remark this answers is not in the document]
uhoh-itsmaciek: Ended up moving it as a follow-up note after the function body.

uhoh-itsmaciek:
Addressed feedback except https://github.com/pganalyze/collector/pull/74#discussion_r476262167 -- let me know what you think.

input/postgres/explain.go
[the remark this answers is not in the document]
lfittl: Ah, yeah, thats a fair point - should we only display the warning in the case of a permission error then?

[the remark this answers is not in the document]
lfittl: Also I just realized, since this is a warning we can use the `PrintWarning` helper, instead of `PrintInfo` (doesn't make much of a difference in practice)

README.md
lfittl: > We don't know the actual parameter types in our query samples, right?

I think thats correct - interestingly enough, Postgres seems to generally output most parameters as string values (even when the data type is integer)

> I don't think there's really much we can do about these cases. We should probably catch "Could not choose a best candidate operator"

Yeah, I think for now we can just make sure that gets stored as an EXPLAIN error - its unclear how common this problem would really be (AFAIK when a column is involved, e.g. in an expression, Postgres can infer the type of the parameter from the column type)

> Also, I was thinking about other situations that could break log-based EXPLAIN, and I think search_path is a big one. Not sure there's an easy fix (at least for complicated cases), but curious if you've given that any thought.

I don't think we can do anything about search path, unfortunately. I think the assumption we make with the code right now is that the search_path thats active for the collector user when connected to the DB, is the same as that which was used for the original query. When someone sets the search path on a per-connection basis, this would break. If necessary we could provide more functionality here, but its unclear if thats really needed.

As a side note, I also briefly thought whether the recent `search_path` security issues with Postgres are of concern here (where the attacker tricks you to call a function in an attacker-controlled schema, due to references in the query not being fully qualified), but since we strictly pass everything through `EXPLAIN`, this wouldn't be of a concern.

input/postgres/explain.go
```diff
@@ -44,40 +44,53 @@ func RunExplain(server state.Server, inputs []state.PostgresQuerySample, collect
...
 						continue
 					}
 
-					params := []string{}
-					for i := 0; i < len(sample.Parameters); i++ {
-						params = append(params, pq.QuoteLiteral(sample.Parameters[i]))
-					}
-					err = db.QueryRow(QueryMarkerSQL + "EXPLAIN (VERBOSE, FORMAT JSON) EXECUTE pganalyze_explain(" + strings.Join(params, ", ") + ")").Scan(&sample.ExplainOutput)
+					paramStr := getQuotedParamsStr(sample.Parameters)
+					err = db.QueryRow(QueryMarkerSQL + "EXPLAIN (VERBOSE, FORMAT JSON) EXECUTE pganalyze_explain(" + paramStr + ")").Scan(&sample.ExplainOutput)
```
lfittl, line 93: I think we should be able to pass this as a query parameter now, without escaping it - correct?

```
err = db.QueryRow(QueryMarkerSQL + "EXPLAIN (VERBOSE, FORMAT JSON) EXECUTE pganalyze_explain($1)", sample.Parameters).Scan(&sample.ExplainOutput)
```

lfittl: Nevermind, I was looking at the wrong place - please disregard this comment :)

input/postgres/explain.go
```diff
@@ -44,40 +44,53 @@ func RunExplain(server state.Server, inputs []state.PostgresQuerySample, collect
...
+		}
+		stmt := parsetree.Statements[0].(pg_query_nodes.RawStmt).Stmt
+		switch stmt.(type) {
+		case pg_query_nodes.SelectStmt, pg_query_nodes.InsertStmt, pg_query_nodes.UpdateStmt, pg_query_nodes.DeleteStmt:
+			sample.HasExplain = true
+			sample.ExplainSource = pganalyze_collector.QuerySample_STATEMENT_LOG_EXPLAIN_SOURCE
+			sample.ExplainFormat = pganalyze_collector.QuerySample_JSON_EXPLAIN_FORMAT
+
+			if useHelper {
+				err = db.QueryRow(QueryMarkerSQL+"SELECT pganalyze.explain($1, $2)", sample.Query, sample.Parameters).Scan(&sample.ExplainOutput)
```
lfittl, line 80: Looks like this needs `pq.Array`, otherwise we get an error (`sql: converting argument $2 type: unsupported type []string, a slice of string`)

```diff
diff --git a/input/postgres/explain.go b/input/postgres/explain.go
index 22d7ca7..91a20fb 100644
--- a/input/postgres/explain.go
+++ b/input/postgres/explain.go
@@ -77,7 +77,7 @@ func runDbExplain(db *sql.DB, inputs []state.PostgresQuerySample, useHelper bool
                        sample.ExplainFormat = pganalyze_collector.QuerySample_JSON_EXPLAIN_FORMAT
 
                        if useHelper {
-                               err = db.QueryRow(QueryMarkerSQL+"SELECT pganalyze.explain($1, $2)", sample.Query, sample.Parameters).Scan(&sample.ExplainOutput)
+                               err = db.QueryRow(QueryMarkerSQL+"SELECT pganalyze.explain($1, $2)", sample.Query, pq.Array(sample.Parameters)).Scan(&sample.ExplainOutput)
                                if err != nil {
                                        sample.ExplainError = fmt.Sprintf("%s", err)
                                }
```

lfittl:
> Addressed feedback except [#74 (comment)](https://github.com/pganalyze/collector/pull/74#discussion_r476262167) -- let me know what you think.

Found one more issue (`pq.Array` being required), otherwise I think we can ship it.

Re: the question in that comment, I think the code as is works, but we can also consider showing the warning only when we actually get a permission error (to support the use case of doing a `GRANT SELECT ON ...` instead of creating the helper function)

input/postgres/explain.go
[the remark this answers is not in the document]
uhoh-itsmaciek: >...should we only display the warning in the case of a permission error then?

Yes, good call.

>Also I just realized, since this is a warning we can use the PrintWarning helper, instead of PrintInfo

I think I got `PrintInfo` from [here](https://github.com/pganalyze/collector/blob/7efb8e0095cde47913511528439c75384d3dba1b/input/postgres/buffercache.go#L80), but makes sense, I'll update.

README.md
uhoh-itsmaciek: Makes sense. Fun fact: I found two `search_path` vulnerabilities in heroku hobby DBs using that mechanism that could give you superuser on the cluster (thus full access to ~2k other hobby DBs).

input/postgres/explain.go
uhoh-itsmaciek: Nice catch, I hadn't actually tried that change yet.

input/postgres/explain.go
```diff
@@ -47,15 +47,24 @@ func RunExplain(server state.Server, inputs []state.PostgresQuerySample, collect
...
 		db.Close()
 
+		hasPermErr := false
+		for _, sample := range dbOutputs {
+			if strings.HasPrefix(sample.ExplainError, "pq: permission denied") {
+				hasPermErr = true
+				break
+			}
+		}
+		if hasPermErr && !connectedAsSuperUser(db, server.Config.SystemType) {
```
uhoh-itsmaciek, line 62: @lfittl come to think of it, we could drop the superuser check here, no?

lfittl: I think we can leave this here, for the case of cloud providers that provide "almost superuser" (which the `connectedAsSuperUser` check returns true for). In those cases we don't want to report the permission error, because the user can't do anything about it.

PR #78, opened 2020-08-27
Fix oid aliasing when constructing snapshot

uhoh-itsmaciek:
Relation oids are not necessarily unique across databases, but we
assume they are, which can cause issues where we clobber some relation
references in our mapping.

Track the mapping per-database instead to avoid this.

~We may need to take a similar approach to other database objects, but I
think a similar pattern should work.~ (I don't think any other code paths are
affected--see below).

I've now tried this in dev and I can't think of a good way to test the issue this
is intended to fix, but it does not break anything.

### DIFF
--- a/output/transform/postgres_relations.go
+++ b/output/transform/postgres_relations.go
@@ -101,7 +101,7 @@ func transformPostgresRelations(s snapshot.FullSnapshot, newState state.Persiste
 		s.RelationInformations = append(s.RelationInformations, &info)
 
 		// Statistic
-		stats, exists := diffState.RelationStats[relation.Oid]
+		stats, exists := diffState.DBStats[relation.DatabaseOid].RelationStats[relation.Oid]
 		if exists {
 			statistic := snapshot.RelationStatistic{
 				RelationIdx:    relationIdx,
@@ -168,7 +168,7 @@ func transformPostgresRelations(s snapshot.FullSnapshot, newState state.Persiste
 			s.IndexInformations = append(s.IndexInformations, &indexInfo)
 
 			// Statistic
-			indexStats, exists := diffState.IndexStats[index.IndexOid]
+			indexStats, exists := diffState.DBStats[relation.DatabaseOid].IndexStats[index.IndexOid]
 			if exists {
 				statistic := snapshot.IndexStatistic{
 					IndexIdx:    indexIdx,
--- a/output/transform/postgres_relations.go
+++ b/output/transform/postgres_relations.go
@@ -7,28 +7,8 @@ import (
 	"github.com/pganalyze/collector/state"
 )
 
-type OidMapByDB map[state.Oid](map[state.Oid]int32)
-
-func MakeOidMap() OidMapByDB {
-	return make(map[state.Oid](map[state.Oid]int32))
-}
-
-func (m OidMapByDB) Put(dbOid, objOid state.Oid, idx int32) {
-	if _, ok := m[dbOid]; !ok {
-		m[dbOid] = make(map[state.Oid]int32)
-	}
-	m[dbOid][objOid] = idx
-}
-
-func (m OidMapByDB) Get(dbOid, objOid state.Oid) int32 {
-	if _, ok := m[dbOid]; !ok {
-		return 0
-	}
-	return m[dbOid][objOid]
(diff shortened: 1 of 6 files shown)


uhoh-itsmaciek:
Looking through the code for other `map[state.Oid]...`, I think the only other place we may need to make changes is `reports/sequence.go`.

output/transform/postgres_relations.go
```diff
@@ -7,8 +7,28 @@ import (
 	"github.com/pganalyze/collector/state"
 )
 
+type OidMapByDB map[state.Oid](map[state.Oid]int32)
```
lfittl, line 10: We could name this `OidToIdxMapByDB` or `OidToIdxMap`

uhoh-itsmaciek:
@lfittl the sequence report is not an issue because it's per-DB. I've updated the others and am testing now.

uhoh-itsmaciek:
@lfittl could you take a look at the last two commits to confirm the logic? This seems to be working, but it required more changes than I anticipated and wanted to make sure I didn't miss anything.

state/state.go
```diff
@@ -8,14 +8,18 @@ import (
...
+
 // PersistedState - State thats kept across collector runs to be used for diffs
 type PersistedState struct {
 	CollectedAt time.Time
 
 	StatementStats PostgresStatementStatsMap
-	RelationStats  PostgresRelationStatsMap
-	IndexStats     PostgresIndexStatsMap
-	FunctionStats  PostgresFunctionStatsMap
+	DBStats        map[Oid]*DBStats
```
lfittl, line 22: We should bump the `StateOnDiskFormatVersion` further down in this file, so if an old state file is read it gets ignored.

state/state.go
```diff
@@ -8,14 +8,18 @@ import (
 	"github.com/pganalyze/collector/config"
 )
 
+type DBStats struct {
```
lfittl, line 11: Maybe `SchemaStats` is a better name? (there are other statistics about the database that are not inside this struct, e.g. statements could also be considered to be part of a database)

state/util.go
```diff
@@ -0,0 +1,21 @@
+package state
```
lfittl, line 1: I wonder if we shouldn't put this into something like`util/oid_map.go` instead, since there might be use cases for this outside of the direct state logic (though `Oid` is defined in the `state` package right now)

runner/diff.go
```diff
@@ -7,8 +7,24 @@ import (
...
+		newDbStats := newState.DBStats[dbOid]
+		prevDbStats := prevState.DBStats[dbOid]
+		var prevRelStats state.PostgresRelationStatsMap
+		var prevIdxStats state.PostgresIndexStatsMap
+		if prevDbStats != nil {
+			prevRelStats = prevDbStats.RelationStats
+			prevIdxStats = prevDbStats.IndexStats
+		} else {
+			prevRelStats = make(state.PostgresRelationStatsMap)
+			prevIdxStats = make(state.PostgresIndexStatsMap)
```
lfittl, line 21: Do we actually need to do this initialization?

My understanding is that its safe to access a key in an uninitialized map: https://play.golang.org/p/TKm5W9UmsIT

lfittl:
> @lfittl could you take a look at the last two commits to confirm the logic? This seems to be working, but it required more changes than I anticipated and wanted to make sure I didn't miss anything.

Thanks for checking - I think this looks good overall, besides some stylistic comments, the only thing we need to do before we ship it is make sure we bump up the state file version.

state/util.go
uhoh-itsmaciek: That makes sense but I feel like in a smaller codebase with a limited number of active contributors, it's fine to leave refactoring like that until you need it: I don't think either of us are likely to forget that this helper exists and accidentally reimplement it somewhere else. If we do need it elsewhere, we can move it (and Oid) before tackling that without any more trouble.

state/state.go
uhoh-itsmaciek: Interesting--I had wondered about state compatibility but I tried running it with a statefile from master specifically to test that and did not have any problems so I figured it Just Works™ somehow. But versioning makes sense; I'll bump it.

PR #81, opened 2020-09-04
Use GitHub Actions for releases

uhoh-itsmaciek:
This may not work right now since some of the documentation is vague, but it should be close. In order to avoid conflicts with Travis releases, we'll need to remove the Travis config, but we obviously don't want to do that until we're sure this is working. We could try providing a different upload file name (.e.g, `__gha_test_pganalyze-collector-linux-amd64`) and merge, and then drop Travis once we confirm the new mechanism for generating the release build works. I'm also not sure if we can do two separate jobs like this--I'm not sure if the second one has access to the build artifacts of the second one or if the workspace is cleaned in between. I think we should definitely continue to run tests before release and release the same build we actually test, so if this does not work, I think we can remove the release job and move its steps to the end of build with the same condition (conditions can be set on jobs or steps).

Thoughts?


uhoh-itsmaciek:
@lfittl this is working now and I confirmed with a dummy release (I've deleted it to avoid confusion but you can see the Workflow execution [here](https://github.com/pganalyze/collector/runs/1088898646?check_suite_focus=true)--we can create and push another dummy tag to create a new release). I ended up adding a separate workflow for releases because that seems most idiomatic. It's not great that we need to manually keep code in sync between CI and the release process, but I think it's less bad than trying to manually maintain workflow dependencies.

.travis.yml
```diff
@@ -1,24 +0,0 @@
...
-script:
-  - make build OUTFILE=pganalyze-collector-linux-amd64
-  - make test
-  - make integration_test
-deploy:
-  skip_cleanup: true
-  provider: releases
-  api_key:
-    secure: wOikWTqlZBnQvy8leNdHCyKroOus4/jo7mlQMyNl0/czTIWSBi/KQNewKEB4bCZxNJUiccl7Ts5JwWMC9nEgHsS4Cd5BXvPfOEh3hR/LQGL833ts8Z4c8axP1g4YSXndTW0nHWu4WZJ4PrPlGmiIiRWCpSSKHC1CmmMJfTzgdX3e34w/kEqUeIrmgOrNZcJnwz36DbS/CL+DWUq8fXX/2OcBLKIJPS+react4guBe/CpGEkbiOsqFD+4R5Ew3k5U1rXXnXrQyiGZx0XPReHO7dnhI/uYL48bsk2RF7wdD9hbrmCDeszJiJaUqxoh1ycpq06MNImWMRInB8U7+DwWcNNA3wOQBE6dQh1qQK9Dfi74pytm589xtIk9C9Q1pnR5E2nM2yAiJpwNL7j6L+TcuCsekm6jpipWY+QhVzCcMlJY9neeKAhTymUu41dxiZw5j4kbdrv2/fzlVTf5xCh91bzDXqUV+IsPyOZLweOjxsKEcwSIqWb3MaoGNiXlzXA7BOAqYHEay/dwXruv4gLx/QKvkXsZQj0vcmRfR9Oym8tj0SmNtafbjIrW5YkzZFIJ6csCN+aUnlCrENaZ0GcIV+STvUo6p+MVKpjrmSh7+RJukGvUP8uO2bCeIv8CBLCjQBarzzjsFyriudEPFLOyBwIuV1AV/PKoVfLSs5HdVa8=
-  file: pganalyze-collector-linux-amd64
```
uhoh-itsmaciek, line 20: We need to avoid duplicate releases/release artifacts so I figured at this point we can just drop Travis.

.github/workflows/ci.yml
```diff
@@ -27,9 +26,8 @@ jobs:
 
     - name: Check out code
       uses: actions/checkout@v2
-
-    - name: Initialize submodules
-      run: git submodule update --init
+      with:
+        submodules: true
```
uhoh-itsmaciek, line 30: Found that there's a built-in way to do this

lfittl (approved):
Looks good!

PR #88, opened 2020-09-22
Add support for assuming instance roles

lfittl:
Set the role to be assumed using the new "aws_assume_role" / AWS_ASSUME_ROLE
configuration setting. This is useful when the collector runs in a different
AWS account than your database.

See https://docs.aws.amazon.com/IAM/latest/UserGuide/tutorial_cross-account-with-roles.html


uhoh-itsmaciek (approved):
Looks great, but what's the actual value here? The full ARN of the role to assume?

lfittl:
> Looks great, but what's the actual value here? The full ARN of the role to assume?

Yeah, exactly - the ARN of the role to assume. Its the same you would pass in to CLI commands/etc when using the assume role functionality.

PR #89, opened 2020-10-06
Hoist log_line_prefix check in logs test

uhoh-itsmaciek:
This lets us ensure we have a valid prefix on all platforms.


lfittl (approved):
:+1:

PR #90, opened 2020-10-09
log about gcp workload identity;  update gcp pubsub lib dependencies

yashbhutwala:
- [X] log about gcp workload identity
- [X] update gcp pubsub lib dependencies


lfittl:
@yashbhutwala Thanks for the contribution! I've edited this a bit and merged it in #91

PR #92, opened 2020-10-14
Drop extraneous KernelVersion lookup

uhoh-itsmaciek:


lfittl (approved):
Good catch - no idea how that got duplicated.

PR #93, opened 2020-10-14
Report collector config in full snapshot

uhoh-itsmaciek:
Add information about collector config and platform to full snapshot.
This will better allow us to monitor some integration and operational
issues.

This is a draft, to sketch out the solution and vet the approach.


input/collector_config.go
```diff
@@ -0,0 +1,53 @@
...
+		DbLogLocation:           c.LogLocation,
+		DbLogDockerTail:         c.LogDockerTail,
+		IgnoreTablePattern:      c.IgnoreTablePattern,
+		IgnoreSchemaRegexp:      c.IgnoreSchemaRegexp,
+		QueryStatsInterval:      int32(c.QueryStatsInterval),
+		MaxCollectorConnections: int32(c.MaxCollectorConnections),
+		FilterLogSecret:         c.FilterLogSecret,
+		FilterQuerySample:       c.FilterQuerySample,
+		HasProxy:                c.HTTPProxy != "" || c.HTTPSProxy != "",
+		ConfigFromEnv:           os.Getenv("PGA_API_KEY") != "",
```
uhoh-itsmaciek, line 51: Is there a better way to check this?

input/collector_config.go
```diff
@@ -0,0 +1,53 @@
...
+		GcpProjectId:            c.GcpProjectID,
+		ApiSystemId:             c.SystemID,
+		ApiSystemType:           c.SystemType,
+		ApiSystemScope:          c.SystemScope,
+		DbLogLocation:           c.LogLocation,
+		DbLogDockerTail:         c.LogDockerTail,
+		IgnoreTablePattern:      c.IgnoreTablePattern,
+		IgnoreSchemaRegexp:      c.IgnoreSchemaRegexp,
+		QueryStatsInterval:      int32(c.QueryStatsInterval),
+		MaxCollectorConnections: int32(c.MaxCollectorConnections),
```
uhoh-itsmaciek, line 47: Was not sure if I should have the int32 casts here or in `transform`--it looks like `state` is using both `int` and `int32` for different things.

input/collector_platform.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"runtime"
+	"time"
+
+	"github.com/pganalyze/collector/state"
+	"github.com/shirou/gopsutil/host"
+)
+
+// TODO: is there a better place to initialize this?
+var collectorStartTime = time.Now()
```
uhoh-itsmaciek, line 12: Also, this causes a test failure in the transform test because it introduces non-determinism. We can potentially just zero it out before comparing, but not sure if this is the right approach in the first place.

input/collector_platform.go
```diff
@@ -0,0 +1,36 @@
...
+
+func getCollectorPlatform() state.CollectorPlatform {
+	hostInfo, err := host.Info()
+	if err != nil {
+		// TODO: log this? return error?
```
uhoh-itsmaciek, line 17: What should we do with errors here?

output/pganalyze_collector/bloat_report.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.25.0
-// 	protoc        v3.12.3
+// 	protoc        v3.13.0
```
uhoh-itsmaciek, line 4: I apparently have a newer `protoc`--should I roll this back and regenerate?

output/transform/collector_platform.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"github.com/golang/protobuf/ptypes"
+	snapshot "github.com/pganalyze/collector/output/pganalyze_collector"
+	"github.com/pganalyze/collector/state"
+)
+
+func transformCollectorPlatform(s snapshot.FullSnapshot, state state.TransientState) snapshot.FullSnapshot {
+	p := state.CollectorPlatform
+	startTs, _ := ptypes.TimestampProto(p.StartedAt)
```
uhoh-itsmaciek, line 11: I noticed we're ignoring the `ptypes.TimestampProto` error elsewhere, so doing that here, too--lmk if we should be doing something with this.

uhoh-itsmaciek: I *think* we're okay unless something is totally off the rails: https://github.com/golang/protobuf/blob/master/ptypes/timestamp.go#L80-L88

lfittl: Yeah, its safe to ignore these in my opinion

input/collector_config.go
lfittl: Either or works - no strong preference one way or another. We could probably adjust the type for these directly in the config settings as well, so we use `int32` throughout. Though I'll note that technically these could be unsigned int32.

input/collector_config.go
lfittl: I think this is fine - `PGA_API_KEY` is the most reliable indicator that the config comes from the environment.

input/collector_platform.go
lfittl: Maybe `logger.PrintVerbose` (and possibly only with test runs?), so we could debug this if we're surprised by this not working in some case.

input/collector_platform.go
lfittl: We could put this into the `state.CollectionOpts` and initialize this in the main function?

(and for tests, passing the collection opts would allow us to set a specific start time and then verify it gets passed through)

lfittl (approved):
Nice - I think this looks great, and I've also double checked that no sensitive configuration values are passed along.

output/pganalyze_collector/bloat_report.pb.go
lfittl: I think its fine to update them

input/collector_config.go
uhoh-itsmaciek: Okay, I'll leave it for now and we can adjust types later.

input/collector_platform.go
uhoh-itsmaciek: That makes sense :+1:

PR #94, opened 2020-10-15
Self-managed server log tail: Speed improvements, handle multi-line logs correctly

lfittl:
```
Self-managed server: Process logs every 3 seconds, instead of on-demand

This reduces the amount of overall log snapshots generated, and also
helps with the overall efficiency of the code if there are many lines
that are not ready yet (the typical case on a busy system).
```

```
Log stream: Optimize stitching logic to allocate one large string

Previously we kept resizing the string, leading to many small allocations.
Since we have all the strings that we need already allocated, instead do
one new allocation once, for the final size of the string.

Note that there is further optimization potential here, but this aims to
reduce the biggest current problem point.
```

```
Log stream: Handle streams with missing PID/timestamps/etc correctly

Previously we would sort lines by PID, even when some of the lines do
not have a PID assigned, such as multi-line logs from when we process a
log tail from a self-managed server.

In those situations that lead to incorrect output. Instead, only consider
these characteristics when everything in the stream has the PID/timestamp
assigned.
```

### DIFF
--- a/logs/stream/stream.go
+++ b/logs/stream/stream.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/pganalyze/collector/grant"
@@ -130,13 +131,36 @@ func handleLogAnalysis(analyzableLogLines []state.LogLine) ([]state.LogLine, []s
 }
 
 func stitchLogLines(readyLogLines []state.LogLine) (analyzableLogLines []state.LogLine) {
-	for _, logLine := range readyLogLines {
-		if logLine.LogLevel != pganalyze_collector.LogLineInformation_UNKNOWN {
+	var linesToAppend []int
+	var linesToAppendLenSum int
+	var b strings.Builder
+	for idx, logLine := range readyLogLines {
+		if logLine.LogLevel == pganalyze_collector.LogLineInformation_UNKNOWN {
+			if len(analyzableLogLines) > 0 {
+				linesToAppend = append(linesToAppend, idx)
+				linesToAppendLenSum = linesToAppendLenSum + len(logLine.Content)
+			}
+		} else {
+			if linesToAppendLenSum > 0 {
+				b.Grow(linesToAppendLenSum)
+				for _, logIdx := range linesToAppend {
+					b.WriteString(readyLogLines[logIdx].Content)
+				}
+				analyzableLogLines[len(analyzableLogLines)-1].Content += b.String()
+				b.Reset()
+				linesToAppend = nil
+				linesToAppendLenSum = 0
+			}
 			analyzableLogLines = append(analyzableLogLines, logLine)
-		} else if len(analyzableLogLines) > 0 {
-			analyzableLogLines[len(analyzableLogLines)-1].Content += logLine.Content
-			analyzableLogLines[len(analyzableLogLines)-1].ByteEnd += int64(len(logLine.Content))
 		}
+	}
+	if linesToAppendLenSum > 0 {
+		b.Grow(linesToAppendLenSum)
+		for _, logIdx := range linesToAppend {
+			b.WriteString(readyLogLines[logIdx].Content)
+		}
+		analyzableLogLines[len(analyzableLogLines)-1].Content += b.String()
+		b.Reset()
 	}
 	return
 }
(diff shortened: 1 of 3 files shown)


logs/stream/stream.go
```diff
@@ -130,14 +132,37 @@ func handleLogAnalysis(analyzableLogLines []state.LogLine) ([]state.LogLine, []s
...
-		if logLine.LogLevel != pganalyze_collector.LogLineInformation_UNKNOWN {
+	var linesToAppend []int
+	var linesToAppendLenSum int
+	var b strings.Builder
+	for idx, logLine := range readyLogLines {
+		if logLine.LogLevel == pganalyze_collector.LogLineInformation_UNKNOWN {
+			if len(analyzableLogLines) > 0 {
+				linesToAppend = append(linesToAppend, idx)
+				linesToAppendLenSum = linesToAppendLenSum + len(logLine.Content)
+			}
```
uhoh-itsmaciek, line 143: Is the "else" here a "should never happen" case? Or an actual "cannot happen" case? (Or are we intentionally dropping these?) Stepping through the logic, I think it's the first/third case, but I'd like to make sure I understand this correctly.

logs/stream/stream.go
```diff
@@ -151,14 +176,28 @@ func AnalyzeStreamInGroups(logLines []state.LogLine) (state.TransientLogState, s
...
 		}
-		if logLines[i].LogLineNumber != 0 && logLines[j].LogLineNumber != 0 && logLines[i].LogLineNumber != logLines[j].LogLineNumber {
+		if allLinesHaveLogLineNumber && logLines[i].LogLineNumber != logLines[j].LogLineNumber {
 			return logLines[i].LogLineNumber < logLines[j].LogLineNumber
 		}
-		if !logLines[i].OccurredAt.IsZero() && !logLines[j].OccurredAt.IsZero() {
+		if allLinesHaveOccurredAt {
 			return logLines[i].OccurredAt.Sub(logLines[j].OccurredAt) < 0
 		}
 		return false // Keep initial order
```
uhoh-itsmaciek, line 203: So we keep initial order if at least one line is missing each of the above, right? Basically, "sort by pid if all lines have it, then sort by line number if all lines have it, then sort by occurred_at if all lines have it, otherwise keep original sort order"? My naïve approach would be "sort by pids if both lines have pids, then sort by line number if both lines have line numbers, then sort by occurred_at if both lines have occurred_at, otherwise keep original sort order", but I'm guessing that doesn't do the right thing somehow? (Since it's simpler, I imagine you considered it.)

logs/stream/stream.go
```diff
@@ -169,17 +208,17 @@ func AnalyzeStreamInGroups(logLines []state.LogLine) (state.TransientLogState, s
...
-		return state.TransientLogState{}, state.LogFile{}, logLines, err
-	}
-
 	// Ensure that log lines that span multiple lines are already concated together before passing them to analyze
 	//
 	// Since we already sorted by PID earlier, it is safe for us to concatenate lines before grouping. In fact,
 	// this is required for cases where unknown log lines don't have PIDs associated
 	analyzableLogLines := stitchLogLines(readyLogLines)
 
+	logFile, err := writeTmpLogFile(analyzableLogLines)
```
uhoh-itsmaciek, line 217: Is the `tmpLogFile` the log equivalent of persistent state?

uhoh-itsmaciek (approved):
Some questions, but looks good.

logs/stream/stream.go
lfittl: It's a temporary file thats used to encrypt the log data before transmission - it's easier to do so on disk, vs in memory. The file is deleted at the end of each upload, i.e. does not persist between calls of this function.

logs/stream/stream.go
lfittl: @uhoh-itsmaciek If I read your comment right, thats actually exactly the logic thats being replaced here. i.e. before we were only checking on a per-line basis whether both lines had a value (and then we sorted by it), and now we only sort by it if all the lines have it.

The problem with sorting by e.g. PID for some lines, but not others, is that multi-line statements are not handled correctly. e.g. in the self-managed case, EXPLAIN output would usually be multi-line, with only the first line having a PID. Thus, that line would be compared with other lines (and moved around), but everything in the multi-line statement that came after it, would not be sorted. Thus the output is buggy

logs/stream/stream.go
lfittl: We are intentionally dropping these. The situation where we somehow got the tail end of a multi-line statement, we can't actually parse it (because we're missing the primary log line for it). Thus it gets discarded here. I could add a comment to clarify this behavior.

PR #95, opened 2020-10-16
Check and report problematic log collection settings

uhoh-itsmaciek:
We actually need to do this somewhere else where we can enable and
disable log collection based on the setting changing at runtime,
but this is a start.


input/collector.go
```diff
@@ -0,0 +1,151 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/pganalyze/collector/config"
+	"github.com/pganalyze/collector/state"
+	"github.com/pganalyze/collector/util"
+	"github.com/shirou/gopsutil/host"
+	"github.com/shirou/gopsutil/process"
+)
+
```
uhoh-itsmaciek, line 16: Consolidated all the input-related collector stuff in this file for now.

input/full.go
```diff
@@ -116,5 +116,9 @@ func CollectFull(server state.Server, connection *sql.DB, globalCollectionOpts s
 	ts.CollectorConfig = getCollectorConfig(server.Config)
 	ts.CollectorPlatform = getCollectorPlatform(globalCollectionOpts, logger)
 
+	logSnapshotDisabled, logSnapshotDisabledReason := getCollectorLogSnapshotStatus(server.Config, globalCollectionOpts, ts.Settings)
+	ts.CollectorLogSnapshotDisabled = logSnapshotDisabled
+	ts.CollectorLogSnapshotDisabledReason = logSnapshotDisabledReason
```
uhoh-itsmaciek, line 121: So this is not the right place to put this because from here we can only really report this, we can't actually control log processing. We need to be able to turn on/off the setup that happens [here](https://github.com/pganalyze/collector/blob/stop-log-processing-if-using-bad-log-settings/main.go#L211-L253) when this value changes, right?

lfittl: The code part you reference only runs once at start (or SIGHUP) time, and I'm thinking we might need something that can respond to changes in the config without requiring a collector reload.

My first idea was to handle it later on in the grant handling, which is where the server-side deactivation would take effect:

https://github.com/pganalyze/collector/blob/stop-log-processing-if-using-bad-log-settings/logs/stream/stream.go#L259

However, its probably better to handle this in a way that also avoids the overhead on the collector side (i.e. don't do any log parsing in such situations), which means we should track this in the persisted state (or similar) and then take action based on that in the log handling. The deactivation logic could go early in the `ProcessLogStream` and `DownloadLogsFromAllServers` functions.

Re: state - There is actually already the `PersistedLogState` struct, which would be a good fit for this I think. We just need to make sure to lock the `LogStateMutex` when we modify/access that struct. Note that we currently don't yet lock this mutex for stream-based logs (only for RDS log file downloads), so we'll need to add that logic to `ProcessLogStream`.

uhoh-itsmaciek: >The code part you reference only runs once at start (or SIGHUP) time, and I'm thinking we might need something that can respond to changes in the config without requiring a collector reload.

Sure, I just meant that we'd need to re-run that subscription logic every time this changed--but I guess you're suggesting we keep the subscriptions active and just discard data, which should be reasonably cheap and sure is simpler. I sketched out doing this in the patch I just pushed. Still needs some cleanup and fixes (I think the way I ended up doing this the subscriptions are just naturally consumed as log collection is forced off, but for AWS I think we'd need to clear the marker--I'll look into that), and I broke actually reporting the state for now, but I think this approach makes sense?

uhoh-itsmaciek: (I also refactored a couple of things to make it easier to follow; sorry if that confuses things)

runner/logs.go
```diff
@@ -16,12 +16,16 @@ import (
 	"github.com/pkg/errors"
 )
 
-func downloadLogsForServer(server state.Server, globalCollectionOpts state.CollectionOpts, logger *util.Logger) (state.PersistedLogState, bool, error) {
+func downloadLogsForServer(server state.Server, globalCollectionOpts state.CollectionOpts, logger *util.Logger) (state.PersistedLogState, error) {
 	newLogState := server.LogPrevState
 
+	if server.LogPrevState.LogSnapshotDisabled {
+		return newLogState, errors.New(server.LogPrevState.LogSnapshotDisabledReason)
```
lfittl, line 23: I wonder if we shouldn't do `PrintWarning` here as well? (to match the stream-based log processing)

runner/full.go
```diff
@@ -17,6 +19,55 @@ import (
 	"github.com/pganalyze/collector/util"
 )
 
+const MinSupportedLogMinDurationStatement = 100
+
+func enforceLogCollectionLimits(server state.Server, transientState state.TransientState) {
```
lfittl, line 24: I wonder if in addition to calling this from the full snapshots, we should also call it once when the collector reloads/restarts (as you had this initially). We'd have to pull in the settings with an extra query for that, but I think thats okay on startup.

That would ensure that you can always clear this behavior by doing that action, vs having to wait up to 10 minutes to get a full snapshot in.

runner/logs.go
uhoh-itsmaciek: Instead of or in addition to returning the error?

runner/full.go
```diff
@@ -17,6 +19,55 @@ import (
...
+		}
+	}
+
+	server.LogStateMutex.Lock()
+	reason := strings.Join(disabledReasons, "; ")
+	if server.LogPrevState.LogSnapshotDisabled != disabled ||
+		server.LogPrevState.LogSnapshotDisabledReason != reason {
+		var newState = server.LogPrevState
+		newState.LogSnapshotDisabled = disabled
+		newState.LogSnapshotDisabledReason = reason
```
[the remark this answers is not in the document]
uhoh-itsmaciek: Ah, I hadn't tried this yet--just wanted to sanity-check the approach. I definitely missed the assignment, but I'm surprised that's not enough :thinking:.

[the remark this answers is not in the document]
uhoh-itsmaciek: No, looking at this again, it makes sense: we're passing the server by value everywhere, so any local modifications to its state are only going to be to our local copy. We'd need to pass a new log state all the way up to CollectAllServers to use this approach and assign it there (and keep the log mutex locked there as well). Unfortunately, this sort of change does not fit nicely into the current design. I'll see what else I can see.

runner/logs.go
lfittl: Instead of, i.e. only output warning and not an error.

runner/full.go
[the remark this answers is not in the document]
uhoh-itsmaciek: @lfittl I just pushed a different version tracking this on Server directly (again, haven't tried it--just wanted to vet the approach). It's more in line with the existing handling--I introduced a CollectionStatus on Server with its own Mutex that tracks the logs_disabled / logs_disabled_reason, and pass that around. The main difference to the other states is that like transient states it's generated anew each run, but like persistent states it's passed back to the orchestrator function to actually update Server. This means if we ever want to expand this to track things that need context from snapshot to snapshot, we'll probably want to refactor, but for now I think it's a reasonable compromise. Let me know what you think.

runner/full.go
uhoh-itsmaciek: @lfittl reading through `main.go`, the [server array is rebuilt](https://github.com/pganalyze/collector/blob/master/main.go#L75) every reload, no? So this new approach, of storing that log-disabled state on the server object itself, rather than the persisted objects solves this problem, right? (Assuming we're okay with <=10m of high log volume after every restart; we could still try to pull up the check to try to avoid that.)

runner/full.go
```diff
@@ -83,33 +132,32 @@ func processDatabase(server state.Server, globalCollectionOpts state.CollectionO
...
 			} else {
-				return state.PersistedState{}, state.Grant{}, err
+				return state.PersistedState{}, state.Grant{}, state.CollectionStatus{}, err
 			}
 		} else {
 			server.Grant = newGrant
 		}
 	}
 
-	transientState := state.TransientState{}
```
uhoh-itsmaciek, line 93: There doesn't seem to be a good reason to actually have this state around? Or am I missing something?

runner/logs.go
```diff
@@ -16,12 +16,12 @@ import (
 	"github.com/pkg/errors"
 )
 
-func downloadLogsForServer(server state.Server, globalCollectionOpts state.CollectionOpts, logger *util.Logger) (state.PersistedLogState, bool, error) {
```
uhoh-itsmaciek, line 19: I believe the `success` `bool` had the same code paths as `error == nil` so I don't think we need it.

uhoh-itsmaciek: Nope never mind, just caught where this was different. Restoring it.

runner/full.go
uhoh-itsmaciek: We do, however, need to avoid logging so frequently on this:

```
2020/10/19 16:09:39 I [server1] Testing log collection (local)...
2020/10/19 16:09:39 W [server1] Skipping logs: log_statement is set to unsupported value 'all'
2020/10/19 16:09:39 W [server1] Skipping logs: log_statement is set to unsupported value 'all'
2020/10/19 16:09:39 W [server1] Skipping logs: log_statement is set to unsupported value 'all'
2020/10/19 16:09:39 W [server1] Skipping logs: log_statement is set to unsupported value 'all'
2020/10/19 16:09:39 W [server1] Skipping logs: log_statement is set to unsupported value 'all'
2020/10/19 16:09:39 W [server1] Skipping logs: log_statement is set to unsupported value 'all'
2020/10/19 16:09:39 W [server1] Skipping logs: log_statement is set to unsupported value 'all'
2020/10/19 16:09:39 W [server1] Skipping logs: log_statement is set to unsupported value 'all'
2020/10/19 16:09:39 W [server1] Skipping logs: log_statement is set to unsupported value 'all'
2020/10/19 16:09:39 W [server1] Skipping logs: log_statement is set to unsupported value 'all'
2020/10/19 16:09:39 W [server1] Skipping logs: log_statement is set to unsupported value 'all'
2020/10/19 16:09:49 E [server1] ERROR - Could not tail logs for server: Timeout
```
(And fix that tail issue.) Looking into this.

uhoh-itsmaciek: Ok, with properly handling this in `TestLogsForAllServers`, we avoid the warnings in the test code path and this becomes:

```
2020/10/19 16:39:21 I [server1] Testing statistics collection...
2020/10/19 16:39:24 I [server1]   Test submission successful (645 KB received, server 6ayj3psrubag7gwuzv7sbwnq7a)
2020/10/19 16:39:24 I [server1] Testing activity snapshots...
2020/10/19 16:39:24 I [server1]   Test submission successful (5.07 KB received, server 6ayj3psrubag7gwuzv7sbwnq7a)
2020/10/19 16:39:24 W [server1] WARNING - Configuration issue: log_statement is set to unsupported value 'all'
2020/10/19 16:39:24 W [server1]   Log collection will be disabled for this server
```
Still looking at avoiding that warning in an actual run.

uhoh-itsmaciek: Log tails seem to be a third place where we need to check this, right? That doesn't seem to be handled by either the log stream subscription handling or the file downloads. Also, I asked this before but didn't follow up: should we be clearing the AWS log marker when we enable collection after it has been disabled? I think without that we'd process all the log files that were generated while the naughty settings were in place, right?

uhoh-itsmaciek: Ok, so in 0259620, I tried to bite the bullet and change all the pass-by-value to pass-by reference. The log-disabling now seems to work. I think according to [this SO discussion](https://stackoverflow.com/questions/23542989/pointers-vs-values-in-parameters-and-return-values) this is a reasonable change (pretty sure the `Server` struct is too big for there to be any performance benefit to pass-by-value, and while avoiding modification by callees is nice to have, I'm not that worried about that). Thoughts?

uhoh-itsmaciek: Oh come to think of it, we can just warn when we determine whether logs are disabled based on settings--that will just happens once per full snapshot which seems reasonable. I'll do that.

uhoh-itsmaciek: Ended up doing this; we could easily warn just once per collector reload as well, but warning once per full snapshot doesn't seem too verbose?

lfittl: > Ended up doing this; we could easily warn just once per collector reload as well, but warning once per full snapshot doesn't seem too verbose?

Yeah, that seems fine - if we get complaints because this is too verbose we could adjust, but I think err-ing on the side of warning frequently is a good thing (to make it clear why that part of the product isn't working).

runner/full.go
```diff
@@ -17,25 +19,71 @@ import (
 	"github.com/pganalyze/collector/util"
 )
 
-func collectDiffAndSubmit(server state.Server, globalCollectionOpts state.CollectionOpts, logger *util.Logger) (state.PersistedState, error) {
+const MinSupportedLogMinDurationStatement = 100
+
+func validateLogCollectionConfig(server *state.Server, settings []state.PostgresSetting) (bool, string) {
```
lfittl, line 24: I'm wondering if we should move this to something like `logs/validate_config.go` instead, to keep it colocated with most of the other log handling code?

runner/full.go
lfittl: Yeah I think this was a leftover from a prior iteration where the sentry client was passed around (which is no longer the case)

runner/logs.go
```diff
@@ -51,8 +51,36 @@ func downloadLogsForServer(server state.Server, globalCollectionOpts state.Colle
...
+
+	server.LogStateMutex.Lock()
+	newLogState, success, err := downloadLogsForServer(server, globalCollectionOpts, prefixedLogger)
+	if err != nil {
+		server.LogStateMutex.Unlock()
+		prefixedLogger.PrintError("Could not collect logs for server: %s", err)
+		if server.Config.ErrorCallback != "" {
+			go runCompletionCallback("error", server.Config.ErrorCallback, server.Config.SectionName, "logs", err, prefixedLogger)
+		}
+	} else if success {
```
lfittl, line 73: It might be better to change the code like this, fixing a problem with the previous code I just discovered whilst reading this (mutex wasn't unlocked when `err == nil && !success`)

```golang
} else {
  server.LogPrevState = newLogState
  server.LogStateMutex.Unlock()
  if success && server.Config.SuccessCallback != "" {
    go runCompletionCallback("success", server.Config.SuccessCallback, server.Config.SectionName, "logs", nil, prefixedLogger)
  }
}
```

lfittl (approved):
This looks excellent, and the refactoring to use a reference for server makes a lot of sense.

Whilst I haven't discovered any logic issues in the code, we should let this bake in one of our systems for a day or so, to check against any unexpected problems. I think we can do so after the merge into master, to keep things simple.

runner/full.go
uhoh-itsmaciek: Hmm, should we integrate the `log_line_prefix` checks into this validation?

uhoh-itsmaciek: On second thought, I think I'm going to leave this for now--that check uses a different mechanism, and I think this patch is probably big enough as is. We may want to revisit this but let's land this first. I did move the check logic to the logs package as you suggest.

PR #98, opened 2020-10-26
Change panic stack trace logging from Verbose to Warning

uhoh-itsmaciek:
These panics are by definitions things that should not happen, so
having more context as to why they happened is always useful.


lfittl (approved):
:+1:

PR #101, opened 2020-11-02
Fix log tailing error printing logic

uhoh-itsmaciek:
Right now, we print errors only after the loop completes, but the loop
never completes because the existing control flow prevents that.

Instead, break from the loop when the context closes and print the
error, if any. Also avoid redundant Close calls that already happen
with the defer we have in place.

Noticed this when investigating a support issue.


lfittl (approved):
:+1:

PR #102, opened 2020-11-02
Disable log collectiong with setting log_error_verbosity=verbose

uhoh-itsmaciek:
We do not parse logs correctly when this is configured, and there's
little reason to have this since Log Insights does this for you.


lfittl (approved):
:+1:

PR #103, opened 2020-11-03
Log file cleanup: Only delete temporary file if it was created

lfittl:
Due to the change in dc4a65638c2d169af90c224cf9ee2300fa88f491 we may
enter cleanup in cases where the TmpFile wasn't actually initialized.

In such cases, we should just do nothing in cleanup.

### DIFF
--- a/state/logs.go
+++ b/state/logs.go
@@ -163,8 +163,10 @@ type LogLine struct {
 }
 
 func (logFile LogFile) Cleanup() {
-	logFile.TmpFile.Close()
-	os.Remove(logFile.TmpFile.Name())
+	if logFile.TmpFile != nil {
+		logFile.TmpFile.Close()
+		os.Remove(logFile.TmpFile.Name())
+	}
 }
 
 func (ls TransientLogState) Cleanup() {


uhoh-itsmaciek (approved):

uhoh-itsmaciek:
CI failure is the `stream_test.go` issue we've seen before; restarted the build for now.

uhoh-itsmaciek:
Restarted for the CI failure a few times now--I think this might be because in [handleLogAnalysis](https://github.com/pganalyze/collector/blob/master/logs/stream/stream.go#L112) we arrange the log lines by backend in a map, and then traverse that (unordered) map to build the result. If I understand this correctly, this is correct but not deterministic. I'm not sure how to solve this, though: attempting to sort in `handleLogAnalysis` just for determinism seems like overkill, and I don't know if there's a reasonable way to sort this in the test for verifying test behavior. Thoughts?

lfittl:
@uhoh-itsmaciek I think we need to add an explicit sort in `handleLogAnalysis`, can't think of a better way to solve this unfortunately. I can look into that.

(merge now since CI did end up being green after a while)

PR #104, opened 2020-11-04
Fix unstable log stream test

lfittl:


uhoh-itsmaciek (approved):
Nice, I didn't realize this would be this simple. And this does not inadvertently correct any sorting issues that should be failures here, right? That's mostly what I was worried about re: sorting in the test, but I guess I misunderstood the mechanism we're testing.

lfittl:
> Nice, I didn't realize this would be this simple. And this does not inadvertently correct any sorting issues that should be failures here, right? That's mostly what I was worried about re: sorting in the test, but I guess I misunderstood the mechanism we're testing.

Generally the output doesn't need to be sorted in a certain way, since what matters here is the concatenation/detection of log levels/etc. Thus we can safely sort both arrays and check the contents.

PR #105, opened 2020-11-06
Introduce API system scope fallback

lfittl:
This fallback is intended to allow changing the API scope, either based
on user configuration (e.g. moving the collector between different
cloud provider accounts), or because of changes in the collector identify
system logic.

The new "api_system_scope_fallback" / PGA_API_SYSTEM_SCOPE_FALLBACK config
variable is intended to be set to the old value of the scope. When the
pganalyze backend receives a snapshot with a fallback scope set, and there
is no server created with the regular scope, it will first search the
servers with the fallback scope. If found, that server's scope will be
updated to the (new) regular scope. If not found, a new server will be
created with the regular scope. The main goal of the fallback scope is to
avoid creating a duplicate server when changing the scope value.


uhoh-itsmaciek (approved):

config/identify_system.go
```diff
@@ -6,16 +6,18 @@ import (
...
 	systemType = config.SystemType
 	systemScope = config.SystemScope
+	systemScopeFallback = config.SystemScopeFallback
 	systemID = config.SystemID
 
 	if config.AwsDbInstanceID != "" || systemType == "amazon_rds" {
 		systemType = "amazon_rds"
 		if systemScope == "" {
-			systemScope = config.AwsRegion
+			systemScope = config.AwsAccountID + "." + config.AwsRegion
```
uhoh-itsmaciek, line 19: Come to think of it, if `AwsAccountId` is unset (empty), we should probably avoid prepending the period, right? And in that case we don't need to set the fallback scope at all?

lfittl: Yeah, good point - I'll adjust this.

PR #107, opened 2020-11-07
Add redacted db_url to collector config reporting

uhoh-itsmaciek:


config/config.go
```diff
@@ -293,6 +293,20 @@ func (config ServerConfig) GetDbHost() string {
...
+	}
+
+	u, err := url.Parse(config.DbURL)
+	if err != nil {
+		return "<unparsable>"
+	}
+
+	u.User = url.User(u.User.Username())
+	return u.String()
+}
```
lfittl, line 308: A test would be good to make sure we don't regress on this.

state/state.go
```diff
@@ -82,6 +82,7 @@ type CollectorConfig struct {
 	DbHasSslkey             bool
 	DbExtraNames            []string
 	DbAllNames              bool
+	DbURL                   string
```
lfittl, line 85: It would be good to use `DbURLRedacted` (or similar) as the name, so its clear that this is not the same as the original config field.

lfittl (approved):
Looks good, two minor comments.

config/config_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+type testItem struct {
+	input    string
+	expected string
+}
+
+var tests = []testItem{
+	{"postgres://user:password@example.com", "postgres://user@example.com"},
+	{"postgres://user@example.com", "postgres://user@example.com"},
+	{string([]byte{0x7f}), "<unparsable>"},
```
uhoh-itsmaciek, line 17: @lfittl so this is insanely contrived, but I literally could not find another string to parse that returned an error from `net/url.Parse`. It seems to be designed for also parsing relative URL and so just URL-encodes everything it doesn't recognize: https://golang.org/pkg/net/url/#Parse

Thoughts?

lfittl: Hah. Yeah, I think this works - really the most important cases are the first two.

I wonder if we should have a test case for something that is like `postgres://user:pass:word@example.com`?

Also probably good to have a longer example URL that includes things like `?sslmode=verify-full` and similar.

uhoh-itsmaciek:
Rebased and confirmed locally that URLs are correctly redacted.

PR #108, opened 2020-11-07
Fix protobuf reference

uhoh-itsmaciek:
Unless we create a merge commit, GitHub	squashes or rebases on merge
of a pull request even when a fast-forward merge is possible, losing
the original SHA. If we then delete the original branch of the
submodule pull request, we may end up with a broken submodule
reference.

In this	case, the pull request referenced the unmerged SHA which no
longer exists. Fix by referencing master instead.


lfittl (approved):
👍

lfittl:
@uhoh-itsmaciek FYI, I did an update of the protobuf folder in #105, which also has this change now - so I think we can close this.

PR #112, opened 2020-11-13
Collector setup wizard

uhoh-itsmaciek:
Still in draft, but I ran through this on a fresh Ubuntu 20.04 VM with stock Postgres installed and the basic setup works.

TODO:
 * [x] ~make sure log rotation is accounted for in log file detection~ re-reading the code, I think this is actually happening correctly when the log collector is used; when it is not used, we rely on stdout being redirected to the log file (which is the case on Ubuntu/Debian)
 * [x] add explain helper function in each database
 * [x] decide on initial set of supported environments
   - right now only self-hosted local Postgres 10+ (only 12 tested so far) on Ubuntu 20.04 works
 * [x] hide systematic log output of steps that can be done automatically by default, as discussed
 * [x] review overall flow
 * [x] tweak copy
   - ~right now to make it clear whenever we make system modifications, I use "(will be saved to Postgres)" or "(will be saved to collector config)" in the prompts, but this is awkward~ awkward but clear; maybe we should consider "write to Postgres" / "write to collector config" in the future?
 * [x] testing (this could use more manual and ideally automated testing)
   - ~automated testing here will be tricky due to the inherent system dependencies~ added scripted mode and we run a simple end-to-end test in Docker
 * [x] prompt to monitor other/all databases
 * [x] apply connection limit to monitoring user
 * [x] ~install contrib packages if necessary~ not necessary on initially-supported systems
 * [x] warn that configuring auto_explain may require a Postgres restart if one is not otherwise required (i.e., pg_stat_statements was already in shared_preload_librarires)
   - right now, we indicate a restart will be required, but this should be clearer
 * [x] ~handle config file permission issue~ this may be a non-issue, this seems to be how the file starts and the collector *can* read it
   - file ends up being owned by root:pganalyze
 * [x] ~handle permissions issues when configuring logs~ use a separate PR to add pganalyze user to adm group for log access
   - we do not check that the pganalyze user has access to the logs--we document this but should decide on a concrete recommened setup for the automated setup process to implement https://pganalyze.com/docs/log-insights/setup/self-managed#collector-test-error-permission-denied


lfittl:
> * [ ]  decide on initial set of supported environments
>   
>   * right now only self-hosted local Postgres 10+ (only 12 tested so far) on Ubuntu 20.04 works

As discussed, it works if we focus on Postgres 10+ and Ubuntu/Debian for now. Note that we should at least test Ubuntu 18.04 and Debian 10.6 as well (since there would probably be no version choice). Additionally, we should support both PGDG and Distribution-provided packages.

Separately, we should do some basic version checks and reject very old versions (e.g. Ubuntu 14.04 was using upstart, which we shouldn't bother with here). For newer versions, if we can support them (e.g. Ubuntu 20.10) then it would be great, but doing a version check if necessary could also work.

> * [ ]  review overall flow
> * [ ]  tweak copy
>   * right now to make it clear whenever we make system modifications, I use "(will be saved to Postgres)" or "(will be saved to collector config)" in the prompts, but this is awkward

=> Lukas to look at these async and review

> * [ ]  testing (this could use more manual and ideally automated testing)
>   * automated testing here will be tricky due to the inherent system dependencies

Ways we could do it:

* Boot an EC2 instance from GitHub actions
* Run a virtual machine some other way
* Figure out how to test this in a Docker container (challenge: expected that both Postgres and the collector run as a service, i.e. that systemd is working)
  * Could we use an environment variable to indicate Postgres was started directly? (not through systemd)
  * => This approach would be best, e.g. to integrate with our `integration_test` logic

=> In any of these cases we need a way to pass in the inputs (e.g. through an inputs file / JSON string / etc) -- that also has utility for other use cases (e.g. large customer automating setup)

> * [ ]  install contrib packages if necessary

=> Maciek to check with Postgres mailinglist on what the situation here is (is contrib still required/in-use these days?)

> * [ ]  handle permissions issues when configuring logs
>   
>   * we do not check that the pganalyze user has access to the logs--we document this but should decide on a concrete recommened setup for the automated setup process to implement https://pganalyze.com/docs/log-insights/setup/self-managed#collector-test-error-permission-denied

For PGDG .deb packages, as discussed, its likely enough that we run `usermod -a -G adm pganalyze`, so that we can access the log files in `/var/log/postgresql` (which already have group read permissions). This stackoverflow has a good description of what the adm group is supposed to do (which fits our use case exactly): https://askubuntu.com/questions/612751/what-is-the-difference-between-the-groups-adm-and-admin#:~:text=adm%3A%20Group%20adm%20is%20used,on%20ubuntu%2011.10%20and%20earlier.

We need to double check that the other Ubuntu/Debian situations are covered by that - ideally they are and for now we can keep this simple.

In the future, for RHEL/CentOS packages this will be more complicated. e.g. PDGD packages there put log files into the `log` directory inside the data directory by default, which requires a more complex fix (https://pganalyze.com/docs/log-insights/setup/self-managed#allowing-access-when-using-postgres-built-in-logging)

setup/main.go
```diff
@@ -0,0 +1,1653 @@
...
+
+	setupState.Log(`Welcome to the pganalyze collector installer!
+
+We will go through a series of steps to set up the collector to monitor your
+Postgres database. We will not make any changes to your database or system
+without confirmation.
+
+At a high level, we will:
+
+ 1. Configure a database user and helper functions for the collector, with minimal access
```
lfittl, line 164: Maybe:

"1. Create a dedicated pganalyze database user with monitoring-only access"

(I think configure => create is the most important clarification here)

setup/main.go
```diff
@@ -0,0 +1,1653 @@
...
+ 2. Update the collector configuration file with these settings
+ 3. Set up the pg_stat_statements extension in your database for basic query performance monitoring
```
lfittl, line 166: I think we can drop the `basic` here (I know we use it in the UI, but I think its more clear if we don't have it)

setup/main.go
```diff
@@ -0,0 +1,1653 @@
...
+ 4. (Optional) Make log-related configuration settings changes to enable our Log Insights feature
```
lfittl, line 167: `Change log-related configuration settings to enable the pganalyze Log Insights feature`

setup/main.go
```diff
@@ -0,0 +1,1653 @@
...
+ 5. (Optional) Set up EXPLAIN plan collection to enable our Automated EXPLAIN feature
```
lfittl, line 168: `Set up EXPLAIN plan collection to enable the pganalyze Automated EXPLAIN feature`

(I think its a bit weird to use "our" in this context)

setup/main.go
```diff
@@ -0,0 +1,1653 @@
...
+	}
+}
+
+func doStep(setupState *SetupState, step *Step) error {
+	if step.Check == nil {
+		panic("step missing completion check")
+	}
+	setupState.Logger.StartStep(step.Description)
+	defer setupState.Logger.EndStep()
+	done, err := step.Check(setupState)
```
lfittl, line 210: When you're running this as non-root, it will currently fail like this:

```
? Continue with setup? Yes
  ✗ failed to check status: open /etc/pganalyze-collector.conf: permission denied
```

It might be good to do an explicit permission check, and say that you should re-run with sudo if the user is not root?

setup/main.go
```diff
@@ -0,0 +1,1653 @@
...
+			return true, nil
+		}
+		return state.CurrentSection.HasKey("enable_log_explain"), nil
+	},
+	Run: func(state *SetupState) error {
+		var optIdx int
+		err := survey.AskOne(&survey.Select{
+			Message: "Select automated EXPLAIN mechanism to use (will be saved to collector config):",
+			Help:    "Learn more about the options at https://pganalyze.com/docs/explain/setup",
+			Options: []string{"auto_explain (recommended)", "log-based EXPLAIN"},
```
lfittl, line 1250: I would use `Log-based EXPLAIN` here (i.e. uppercase `L`), for consistency with the docs

setup/main.go
```diff
@@ -0,0 +1,1653 @@
...
+		}
+		var pendingSettings []string
+		for _, row := range rows {
+			pendingSettings = append(pendingSettings, row.GetString(0))
+		}
+
+		pendingList := getConjuctionList(pendingSettings)
+		var restartNow bool
+		err = survey.AskOne(&survey.Confirm{
+			Message: fmt.Sprintf("Postgres must be restarted for changes to %s to take effect; restart Postgres now?", pendingList),
```
lfittl, line 1445: This one is probably the most impactful step as it would be visible to any applications connected - can we somehow give extra attention to this? (use different color, add exclamation mark, etc)

setup/main.go
```diff
@@ -0,0 +1,1653 @@
...
+		return applyConfigSetting("shared_preload_libraries", newSpl, state.QueryRunner)
+	},
+}
+
+var reloadCollector = &Step{
+	Description: "Reload collector configuration",
+	Check: func(state *SetupState) (bool, error) {
+		// N.B.: there's no way to tell whether the collector actually needs to reload; we
+		// force this once per setup helper invocation (this starts out as false and is set
+		// to true below)
```
lfittl, line 1399: I was wondering why I was prompted again when cancelling the setup steps and re-running, but I think this works for now.

For reference, this is what I ran (I quit the setup whilst on the last question shown here):

```
? Continue with setup? Yes
? Select Postgres superuser to connect as for configuration purposes postgres
? Proceed to configuring optional Log Insights feature? Yes
? Proceed to configuring optional Automated EXPLAIN feature? Yes
? The collector configuration must be reloaded for changes to take effect; reload now? Yes
? Review auto_explain configuration settings? Yes
? Set auto_explain.log_analyze to (will be saved to Postgres):  [Use arrows to move, type to filter, ? for more help]
> on (recommended)
  off
```

setup/main.go
```diff
@@ -0,0 +1,1653 @@
...
+		}
+		if !doReview {
+			state.SkipAutoExplainRecommendedSettings = true
+			return nil
+		}
+
+		var logAnalyzeIdx int
+		var logAnalyzeOpts = []string{"on", "off"}
+		err = survey.AskOne(&survey.Select{
+			Message: "Set auto_explain.log_analyze to (will be saved to Postgres):",
```
lfittl, line 1496: Should we maybe just ask `Use recommended auto_explain configuration?` instead of checking each setting individually?

setup/main.go
```diff
@@ -0,0 +1,1653 @@
...
+		}
+		logFormat := logFormatOpts[logFormatIdx]
+		err = applyConfigSetting("auto_explain.log_format", logFormat, state.QueryRunner)
+		if err != nil {
+			return err
+		}
+
+		var logMinDuration int
+		err = survey.AskOne(&survey.Input{
+			Message: "Set auto_explain.log_min_duration, in milliseconds, to (will be saved to Postgres):",
```
lfittl, line 1593: I wonder if we should just default to 1000 here? (to get a working setup without the person setting up wondering whats a good threshold - which they likely don't know yet, because they could evaluate this based on the query performance data they can only access at a later point)

setup/main.go
```diff
@@ -0,0 +1,1653 @@
...
+	}
+
+	setupState.Verbose("  re-checking...")
+	done, err = step.Check(setupState)
+	if err != nil {
+		setupState.Log("✗ failed to check status: %s", err)
+		return err
+	}
+	if !done {
+		err := errors.New("check still failed after running resolution; please try again")
```
lfittl, line 237: Its not clear to me how one should respond when this happens - I did everything right (I think?) and I keep getting this:

```
? Continue with setup? Yes
? Select Postgres superuser to connect as for configuration purposes postgres
? Proceed to configuring optional Log Insights feature? Yes
? Proceed to configuring optional Automated EXPLAIN feature? Yes
? The collector configuration must be reloaded for changes to take effect; reload now? Yes
? Review auto_explain configuration settings? Yes
? Set auto_explain.log_analyze to (will be saved to Postgres): on (recommended)
? Set auto_explain.log_buffers to (will be saved to Postgres): on (recommended)
? Set auto_explain.log_timing to (will be saved to Postgres): on
? Set auto_explain.log_triggers to (will be saved to Postgres): on (recommended)
? Set auto_explain.log_verbose to (will be saved to Postgres): on (recommended)
? Set auto_explain.log_format to (will be saved to Postgres): text
? Set auto_explain.log_min_duration, in milliseconds, to (will be saved to Postgres): 
? Set auto_explain.log_nested_statements to (will be saved to Postgres): on (recommended)
  ✗ check still failed after running resolution; please try again
```

Can we show a more verbose error here?

setup/main.go
uhoh-itsmaciek: Technically you can use an existing user. Should we still allow that and have "create" in this overview copy? I think that probably makes sense, just wanted to confirm...

setup/main.go
uhoh-itsmaciek: Yeah, makes sense. Just check `os.Geteuid`, right?

setup/main.go
uhoh-itsmaciek: Maybe? I wanted to be really clear about what we're doing at every step. I suppose we could show you all the settings we would change, but if you want to change some of them but not all, that's frustrating. We should probably, though, only change settings that are not already set to the recommended values, right?

setup/main.go
lfittl: Ah, good point. Hmm. Maybe "Configure database access using the pganalyze monitoring user"?

setup/main.go
lfittl: Yeah, I think so

setup/main.go
lfittl: Fair point - yes, it would be good if we limited it to the settings that are not configured correctly yet

setup/main.go
uhoh-itsmaciek: Like, do a two-part question like I'm doing in other places (1000/other, then enter other if that's what you selected) or just do yes/no set-to-1000 (and you can tweak separately if you want)?

lfittl: Maybe 1000/other is good (with 1000 being recommended) - if this is the Nth server someone is setting up they might know exactly which value they'd like to use.

setup/main.go
uhoh-itsmaciek: So, as discussed, there's a bug here. I'll fix that, but more generally, in terms of giving more info for why re-checks fail, I'm not sure there's a good way to add meaningful, actionable feedback here. There's no error per se: we run the check for the step, if that does not pass, we run the resolution (typically requiring user input) and then run the check again. If the check still does not pass, the reason should be obvious from the nature of the check (if not, we may have made the step too big). (If there is an actual error running the check--e.g., we can't connect to the database--we do print that.)

Let's review again once this is working as intended and then figure out how to address this (maybe we should prompt you to contact support in these situations, or re-run with the --verbose or --log flag and then contact support?).

setup/main.go
uhoh-itsmaciek: I suppose we could track whether we've made any changes to the collector config at all in the session--if not, there's no need to restart.

setup/main.go
uhoh-itsmaciek: Thinking about this again, you're right that we should mention "create"--that's clearly something users will want to know about. So maybe "Configure database access and, if necessary, create the pganalyze database user with monitoring-only access"?

setup/main.go
```diff
@@ -0,0 +1,250 @@
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"flag"
```
lfittl, line 6: In the main codebase we use https://github.com/ogier/pflag (aliased to `flag`) - the main difference is that the Go standard library uses a non-standard way of passing arguments (in particular, `-longname` is permitted, instead of only allowing `--longname`)

Probably good to use that here as well for consistency?

setup/main.go
```diff
@@ -0,0 +1,250 @@
...
+		if (step.Kind == state.LogInsightsStep &&
+			setupState.Inputs.SkipLogInsights.Valid &&
+			setupState.Inputs.SkipLogInsights.Bool) ||
+			(step.Kind == state.AutomatedExplainStep &&
+				((setupState.Inputs.SkipLogInsights.Valid &&
+					setupState.Inputs.SkipLogInsights.Bool) ||
+					(setupState.Inputs.SkipAutomatedExplain.Valid &&
+						setupState.Inputs.SkipAutomatedExplain.Bool))) {
+			continue
+		}
```
lfittl, line 191: I wonder if for clarity this wouldn't read better as

```golang
skipLogInsights := setupState.Inputs.SkipLogInsights.Valid && setupState.Inputs.SkipLogInsights.Bool
skipAutomatedExplain := setupState.Inputs.SkipLogInsights.Valid && setupState.Inputs.SkipLogInsights.Bool
if step.Kind == state.LogInsights && skipLogInsights {
  continue
}
if step.Kind == state.AutomatedExplainStep && (skipLogInsights || skipAutomatedExplain) {
  continue
}
```

setup/main.go
```diff
@@ -0,0 +1,250 @@
...
+
+Please run pganalyze-collector --reload to apply these changes.`)
+			}
+			os.Exit(1)
+		}
+	}
+	setupState.Log(`
+Setup complete!
+
+Check for integration instructions in the pganalyze app.`)
```
lfittl, line 206: Maybe: 

```
Collector setup complete!

For next steps, go back to the pganalyze app in your web browser.
```

setup/main.go
```diff
@@ -0,0 +1,250 @@
...
+collector config file or your database. If you prefer, you can instead follow
+the manual collector install instructions: https://pganalyze.com/docs/install`)
+		os.Exit(1)
+	}
+
+	setupState.Log(`Welcome to the pganalyze collector guided setup!
+
+We will go through a series of steps to set up the collector to monitor your
+Postgres server. We will not make any changes to Postgres or your system
+without confirmation.
```
lfittl, line 139: I wonder if we should add a disclaimer somewhere here, that this only supports a certain kind of setup:

```
IMPORTANT: Please note that this setup only works when monitoring a self-managed system, and installing the collector directly on your database server. For other setup types, please check the installation instructions https://pganalyze.com/docs/install
```

(in particular, there are cases where you run the collector on a VM, but you monitor a remote database server, e.g. for unsupported PaaS providers - in case the user ends up in this by accident, we should be very clear on this)

setup/service/service.go
```diff
@@ -0,0 +1,16 @@
+package service
+
+import (
+	"fmt"
+	"os/exec"
+)
+
+func RestartPostgres() error {
+	// TODO: account for different environments; handle error output
```
lfittl, line 9: Do we need to resolve this TODO for the initial release?

setup/state/state.go
```diff
@@ -0,0 +1,194 @@
...
+	"%m [%p] %q[user=%u,db=%d,app=%a] ",
+	"%m [%p] %q[user=%u,db=%d,app=%a,host=%h] ",
+	"%t:%r:%u@%d:[%p]:",
+	"%t [%p-%l] %q%u@%d ",
+	"%t [%p]: [%l-1] user=%u,db=%d - PG-%e ",
+	"%t [%p]: [%l-1] user=%u,db=%d,app=%a,client=%h ",
+	"%t [%p]: [%l-1] [trx_id=%x] user=%u,db=%d ",
+	"%m %r %u %a [%c] [%p] ",
+	"%m [%p][%v] : [%l-1] %q[app=%a] ",
+	"%m [%p] ",
```
lfittl, line 20: Hmm. I wonder if there is a way we could re-use the list from the main collector code here? (not sure if e.g. we could do a `shared` directory that can be used by all binaries, including this one?)

setup/state/state.go
```diff
@@ -0,0 +1,194 @@
...
+	AutoExplainLogBuffers:          null.StringFrom("on"),
+	AutoExplainLogTiming:           null.StringFrom("off"),
+	AutoExplainLogTriggers:         null.StringFrom("on"),
+	AutoExplainLogVerbose:          null.StringFrom("on"),
+	AutoExplainLogFormat:           null.StringFrom("json"),
+	AutoExplainLogMinDuration:      null.IntFrom(1000),
+	AutoExplainLogNestedStatements: null.StringFrom("on"),
+}
+
+type SetupInputs struct {
```
lfittl, line 69: :+1: This is excellent :)

setup/util/reload_posix.go
```diff
@@ -0,0 +1,45 @@
+// +build linux freebsd darwin
```
lfittl, line 1: If we have a way of sharing code with the main binary, we could avoid copying this (not sure if thats straightforward?)

setup/util/util.go
```diff
@@ -0,0 +1,150 @@
...
+	if ansNum < 10 && ansNum != -1 {
+		return errors.New("value must be either -1 to disable or 10 or greater")
+	}
+	return nil
+}
+
+var expectedMd5s = map[string]string{
+	"explain":              "7a0a1784d170975d8538d3b8b38c3fad",
+	"get_stat_replication": "066680efec598232c0245477976a2c3d",
+}
```
lfittl, line 55: Nice, thats a great idea - I wonder if we could also check that the function is security definer, and is owned by a superuser?

lfittl (commented):
I've done a first pass on the code, but didn't go into the individual steps yet. Overall this looks excellent!

setup/service/service.go
uhoh-itsmaciek: That's a good question. For different envs, looks like Debian's been using systemd by default since 8, the oldest supported release, so we're fine there. I'll double-check the error output handling.

setup/main.go
uhoh-itsmaciek: We do have checks later, but you're right, I'll add something up front.

setup/main.go
uhoh-itsmaciek: Oh whoops--I'm pretty sure I was using pflag here initially, but I think I broke this in refactoring when Go imports helpfully imported the default flag package instead. Good catch; will fix.

setup/state/state.go
uhoh-itsmaciek: @lfittl Oh, good point. I just didn't think of that. No need to rearrange the code--we can import the logs package here I think. The only thing is, the list in `parse.go` puts the prefix that we document as recommended as third in the SupportedPrefixes list. The recommended prefix should be first in the list we present to the user. At first glance, it looks like I should just be able to rearrange SupportedPrefixes in `parse.go`--is there any reason not to do that?

setup/util/util.go
uhoh-itsmaciek: Good call. I think for now I can just add that as predicates to the SQL below.

setup/util/reload_posix.go
uhoh-itsmaciek: We can share code--the problem is that code calls `os.Exit` and logs some stuff that doesn't really make sense in this context.

Looking at this again, though, I can refactor to share code here without too much trouble. We do also have similar around `discoverLogLocation`, and that might be tougher, since that's more integrated into other collector processing. (Ideally, though, that functionality will migrate into the setup binary entirely over time, I think?)

setup/state/state.go
lfittl: No reason - seems good to re-arrange there (or have a separate constant with the "recommended prefix" as an index?)

uhoh-itsmaciek: Oh good idea. Having a separate index will be more robust against changes.

setup/steps/apply_monitoring_user_password.go
```diff
@@ -0,0 +1,86 @@
...
+
+	survey "github.com/AlecAivazis/survey/v2"
+	"github.com/lib/pq"
+	"github.com/pganalyze/collector/config"
+	s "github.com/pganalyze/collector/setup/state"
+	mainUtil "github.com/pganalyze/collector/util"
+)
+
+var ApplyMonitoringUserPasswd = &s.Step{
+	Description: "Apply monitoring user password",
```
lfittl, line 19: Would be good to clarify whether that this is modifying the pganalyze collector config (not the database)

setup/steps/apply_monitoring_user_password.go
```diff
@@ -0,0 +1,86 @@
...
+	Check: func(state *s.SetupState) (bool, error) {
+		cfg, err := config.Read(
```
lfittl, line 21: It would be helpful to explain why where are using the main collector code for reading the config here (i.e. to be able to use `GetPqOpenString` below

setup/steps/apply_monitoring_user_password.go
```diff
@@ -0,0 +1,86 @@
...
+			return err
+		}
+		pgaPasswd := pgaPasswdKey.String()
+
+		var doPasswdUpdate bool
+		if state.Inputs.Scripted {
+			if !state.Inputs.UpdateMonitoringPassword.Valid || !state.Inputs.UpdateMonitoringPassword.Bool {
+				return errors.New("update_monitoring_password flag not set and cannot log in with current credentials")
+			}
+			doPasswdUpdate = state.Inputs.UpdateMonitoringPassword.Bool
```
lfittl, line 63: It would be good to clean this up, either by:

1) Only using `doPasswdUpdate` in the `else` branch (and leaving this logic as is)
2) Removing the check for `!state.Inputs.UpdateMonitoringPassword.Bool` in the earlier if condition, and running through the regular failure code path (same as when saying No to prompt) in case of the flag being set to false

setup/steps/apply_monitoring_user_password.go
```diff
@@ -0,0 +1,86 @@
...
+				return err
+			}
+		}
+
+		if !doPasswdUpdate {
+			return nil
+		}
+		err = state.QueryRunner.Exec(
+			fmt.Sprintf(
+				"ALTER USER %s WITH ENCRYPTED PASSWORD %s",
```
lfittl, line 79: It might be good to prefix `SET log_statement = none; ` here to prevent the password from being logged (`log_statement = ddl` is a recommended config for pganalyze)

setup/steps/autoexplain_recommended_settings.go
```diff
@@ -0,0 +1,519 @@
...
+	s "github.com/pganalyze/collector/setup/state"
+	"github.com/pganalyze/collector/setup/util"
+)
+
+// N.B.: this needs to happen *after* the Postgres restart so that ALTER SYSTEM
+// recognizes these as valid configuration settings
+var ConfigureAutoExplain = &s.Step{
+	Kind:        s.AutomatedExplainStep,
+	Description: "Review auto_explain settings",
+	Check: func(state *s.SetupState) (bool, error) {
```
lfittl, line 21: It could be good to check the config settings here in `Check` as well, so we don't enter `Run` if all settings are set to recommended values.

setup/steps/autoexplain_recommended_settings.go
```diff
@@ -0,0 +1,519 @@
...
+		var doReview bool
+		if state.Inputs.Scripted {
+			if state.Inputs.SkipAutoExplainRecommended.Valid {
+				doReview = !state.Inputs.SkipAutoExplainRecommended.Bool
+			}
+		} else {
+			err := survey.AskOne(&survey.Confirm{
+				Message: "Review auto_explain configuration settings?",
+				Default: false,
+				Help:    "Optional, but will ensure best balance of monitoring visibility and performance; review these settings at https://www.postgresql.org/docs/current/auto-explain.html#id-1.11.7.13.5",
```
lfittl, line 43: Probably better to link to https://pganalyze.com/docs/explain/setup/auto_explain instead - that also makes it easier to update in the future.

setup/steps/autoexplain_recommended_settings.go
```diff
@@ -0,0 +1,519 @@
...
+			)
+		}
+		if state.Inputs.GUCS.AutoExplainLogMinDuration.Valid {
+			// N.B.: here we check for exact equality with the setting, rather than
+			// under the threshold, since the semantics of that behavior are more
+			// straightforward when providing a setting from an inputs file
+			predParts = append(
+				predParts,
+				fmt.Sprintf(
+					"(name = 'auto_explain.log_min_duration' AND setting::float <> %d)",
```
lfittl, line 256: This should probably be `setting::integer`

setup/steps/autoexplain_recommended_settings.go
```diff
@@ -0,0 +1,519 @@
...
+
+	} else {
+		predicate = fmt.Sprintf(
+			`(name = 'auto_explain.log_analyze' AND setting <> %s) OR
+(name = 'auto_explain.log_buffers' AND setting <> %s) OR
+(name = 'auto_explain.log_timing' AND setting <> %s) OR
+(name = 'auto_explain.log_triggers' AND setting <> %s) OR
+(name = 'auto_explain.log_verbose' AND setting <> %s) OR
+(name = 'auto_explain.log_format' AND setting <> %s) OR
+(name = 'auto_explain.log_min_duration' AND setting::float < %d) OR
```
lfittl, line 278: `setting::integer`

setup/steps/autoexplain_recommended_settings.go
```diff
@@ -0,0 +1,519 @@
...
+			panic("auto_explain.log_nested_statements setting needs review but was not provided")
+		}
+		return state.Inputs.GUCS.AutoExplainLogNestedStatements.String, nil
+	}
+
+	var logNestedIdx int
+	opts, optLabels := getBooleanOpts(currValue, s.RecommendedGUCS.AutoExplainLogNestedStatements.String)
+	err := survey.AskOne(&survey.Select{
+		Message: fmt.Sprintf("Setting auto_explain.log_nested_statements is currently set to '%s'", currValue),
+		Help:    "Consider statements executed inside functions for logging",
```
lfittl, line 494: Maybe just using this wording from the Postgres docs works a bit better:

`Causes nested statements (statements executed inside a function) to be considered for logging`

setup/steps/check_auto_explain_available.go
```diff
@@ -0,0 +1,33 @@
...
+				return false, nil
+			}
+
+			return false, err
+		}
+		return true, err
+	},
+	Run: func(state *s.SetupState) error {
+		// TODO: install contrib package?
+		return errors.New("module auto_explain is not available")
```
lfittl, line 31: Probably good to clarify as `contrib module auto_explain is not available`

setup/steps/check_auto_explain_available.go
```diff
@@ -0,0 +1,33 @@
...
+			if strings.Contains(err.Error(), "No such file or directory") {
+				return false, nil
+			}
+
+			return false, err
+		}
+		return true, err
+	},
+	Run: func(state *s.SetupState) error {
+		// TODO: install contrib package?
```
lfittl, line 30: As discussed, we should drop that TODO

setup/steps/check_auto_explain_available.go
```diff
@@ -0,0 +1,33 @@
...
+	"errors"
+	"strings"
+
+	s "github.com/pganalyze/collector/setup/state"
+	"github.com/pganalyze/collector/setup/util"
+)
+
+var CheckAutoExplainAvailable = &s.Step{
+	Kind:        s.AutomatedExplainStep,
+	Description: "Prepare for auto_explain install",
```
lfittl, line 13: We should update this description (since we don't actually install anything here, we just check things)

setup/steps/check_pgss_available.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/pganalyze/collector/setup/query"
+	s "github.com/pganalyze/collector/setup/state"
+)
+
+var CheckPgssAvailable = &s.Step{
+	Description: "Prepare for pg_stat_statements install",
```
lfittl, line 12: We should update this description (since we don't actually install anything here, we just check things)

setup/steps/check_pgss_available.go
```diff
@@ -0,0 +1,30 @@
...
+	Check: func(state *s.SetupState) (bool, error) {
+		row, err := state.QueryRunner.QueryRow(
+			fmt.Sprintf(
+				"SELECT true FROM pg_available_extensions WHERE name = 'pg_stat_statements'",
```
lfittl, line 16: We should fully schema-qualify this (and other similar queries), to avoid someone creating an object in the `public` schema (or maybe set search path?)

setup/steps/check_pgss_available.go
```diff
@@ -0,0 +1,30 @@
...
+		)
+		if err == query.ErrNoRows {
+			return false, nil
+		} else if err != nil {
+			return false, err
+		}
+		return row.GetBool(0), nil
+	},
+	Run: func(state *s.SetupState) error {
+		// TODO: install contrib package on systems where packaged separately
```
lfittl, line 27: We can drop this todo

setup/steps/check_pgss_available.go
```diff
@@ -0,0 +1,30 @@
...
+		return errors.New("extension pg_stat_statements is not available")
```
lfittl, line 28: `contrib extension pg_stat_statements is not available`

setup/steps/check_platform.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"errors"
+
+	s "github.com/pganalyze/collector/setup/state"
+	"github.com/shirou/gopsutil/host"
+)
+
+var CheckPlatform = &s.Step{
+	Description: "Check platform",
```
lfittl, line 11: Might be good to clarify what platform means (OS? etc)

setup/steps/check_platform.go
```diff
@@ -0,0 +1,36 @@
...
+		state.Platform = hostInfo.Platform
+		state.PlatformFamily = hostInfo.PlatformFamily
+		state.PlatformVersion = hostInfo.PlatformVersion
+
+		if state.Platform == "ubuntu" {
+			if state.PlatformVersion < "14.04" {
+				return false, errors.New("Ubuntu versions older than 14.04 are not supported")
+			}
+		} else if state.Platform == "debian" {
+			if state.PlatformVersion < "10.0" {
```
lfittl, line 27: This would not sort correctly for version `9.0` (since lexically that would come after `10.0`) - as discussed it might work if we parse these as floats and then compare them.

setup/steps/check_platform.go
```diff
@@ -0,0 +1,36 @@
...
+				return false, errors.New("Debian versions older than 10 are not supported")
+			}
+		} else {
+			return false, errors.New("this distribution is not currently supported; please contact support")
```
lfittl, line 31: Might be helpful if we output the `state.Platform` value in the error as well (so we know the OS right away when someone reports this error)

setup/steps/check_postgres_version.go
```diff
@@ -0,0 +1,29 @@
...
+		state.PGVersionStr = row.GetString(0)
+		state.PGVersionNum = row.GetInt(1)
+
+		if state.PGVersionNum < 100000 {
+			return false, fmt.Errorf("not supported for Postgres versions older than 10; found %s", state.PGVersionStr)
+		}
+
+		if state.PGVersionNum >= 120000 {
+			state.QueryRunner.EnableCSV()
+		}
```
lfittl, line 25: We should make this dependent on the client version, not the server version (and could check directly in query runner, as discussed)

setup/steps/check_replication_status.go
```diff
@@ -0,0 +1,23 @@
...
+	Description: "Check replication status",
+	Check: func(state *s.SetupState) (bool, error) {
+		result, err := state.QueryRunner.QueryRow("SELECT pg_is_in_recovery()")
+		if err != nil {
+			return false, err
+		}
+		isReplicationTarget := result.GetBool(0)
+
+		if isReplicationTarget {
+			return false, errors.New("not supported for replicas")
```
lfittl, line 19: maybe slightly more clear if we word it as `Postgres server is a replica - this is currently not supported in this setup`

setup/steps/check_replication_status.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+var CheckReplicationStatus = &s.Step{
+	Description: "Check replication status",
+	Check: func(state *s.SetupState) (bool, error) {
+		result, err := state.QueryRunner.QueryRow("SELECT pg_is_in_recovery()")
+		if err != nil {
+			return false, err
+		}
+		isReplicationTarget := result.GetBool(0)
```
lfittl, line 16: maybe just `pgIsInRecovery`? (not clear what target is referring to here)

setup/steps/check_restart_needed.go
```diff
@@ -0,0 +1,68 @@
...
+NOTICE: A Postgres restart will be required to set up query performance monitoring.
+A prompt will ask to confirm the restart before this guided setup performs it.
+`,
+			)
+		} else if !hasAutoExplain {
+			state.Log(
+				`
+NOTICE: A Postgres restart will not be required to set up query performance monitoring.
+
+A restart *will* be required for the recommended setup of the Automated EXPLAIN feature,
```
lfittl, line 35: Maybe add a `However,` at the start of this sentence

setup/steps/configure_log_duration.go
```diff
@@ -0,0 +1,50 @@
...
+			(state.Inputs.Scripted && state.Inputs.GUCS.LogDuration.Valid &&
+				state.Inputs.GUCS.LogDuration.String != currValue)
+
+		return !needsUpdate, nil
+	},
+	Run: func(state *s.SetupState) error {
+		var turnOffLogDuration bool
+		if state.Inputs.Scripted {
+			if !state.Inputs.GUCS.LogDuration.Valid {
+				return errors.New("log_error_verbosity value not provided and current value not supported")
```
lfittl, line 32: `log_duration`, not `log_error_verbosity`

setup/steps/configure_log_duration.go
```diff
@@ -0,0 +1,50 @@
...
+			}
+			turnOffLogDuration = state.Inputs.GUCS.LogDuration.String == "off"
```
lfittl, line 34: I think before this we need an explicit check like this:

```golang
if state.Inputs.GUCS.LogDuration.String == "on" {
  return errors.new("log_duration value provided as \"on\" - this is not a supported configuration for the pganalyze collector, please set to \"off\"")
}
```

setup/steps/configure_log_error_verbosity.go
```diff
@@ -0,0 +1,48 @@
...
+				currVal != state.Inputs.GUCS.LogErrorVerbosity.String)
+
+		return !needsUpdate, nil
+	},
+	Run: func(state *s.SetupState) error {
+		var newVal string
+		if state.Inputs.Scripted {
+			if !state.Inputs.GUCS.LogErrorVerbosity.Valid {
+				return errors.New("log_error_verbosity value not provided and current value not supported")
+			}
```
lfittl, line 34: Similar to `log_duration = on`, we'd have to error out here for the unsupported value of `log_error_verbosity = verbose`

setup/steps/configure_log_error_verbosity.go
```diff
@@ -0,0 +1,48 @@
...
+			newVal = state.Inputs.GUCS.LogErrorVerbosity.String
+		} else {
+			err := survey.AskOne(&survey.Select{
+				Message: "Setting 'log_error_verbosity' is set to unsupported value 'all'; select supported value (will be saved to Postgres):",
```
lfittl, line 38: this should be `is set to unsupported value 'verbose'`

setup/steps/configure_log_line_prefix.go
```diff
@@ -0,0 +1,68 @@
...
+
+		return !needsUpdate, nil
+	},
+	Run: func(state *s.SetupState) error {
+		row, err := state.QueryRunner.QueryRow(`SELECT setting FROM pg_settings WHERE name = 'log_line_prefix'`)
+		if err != nil {
+			return err
+		}
+		oldVal := row.GetString(0)
+		var opts []string
```
lfittl, line 36: This can be moved to inside the `else` condition (only needed when scripted)

setup/steps/configure_log_line_prefix.go
```diff
@@ -0,0 +1,68 @@
...
+			}
+			opts = append(opts, opt)
+		}
+
+		var selectedPrefix string
+		if state.Inputs.Scripted {
+			if !state.Inputs.GUCS.LogLinePrefix.Valid {
+				return errors.New("log_line_prefix not provided and current setting is not supported")
+			}
+			selectedPrefix = state.Inputs.GUCS.LogLinePrefix.String
```
lfittl, line 53: We need to validate that the `selectedPrefix` is a supported prefix

setup/steps/configure_log_location.go
```diff
@@ -0,0 +1,163 @@
...
+)
+
+var ConfigureLogLocation = &s.Step{
+	Kind:        state.LogInsightsStep,
+	Description: "Determine log location",
+	Check: func(state *s.SetupState) (bool, error) {
+		return state.CurrentSection.HasKey("db_log_location"), nil
+	},
+	Run: func(state *s.SetupState) error {
+		guessedLogLocation, err := discoverLogLocation(state.CurrentSection, state.QueryRunner)
```
lfittl, line 26: We should not run the guessing logic if an explicit scripted input is set (`state.Inputs.Settings.DBLogLocation.Valid`)

setup/steps/configure_log_location.go
```diff
@@ -0,0 +1,163 @@
...
+		if err != nil {
+			return err
```
lfittl, line 28: We could consider making this a soft failure (i.e. always prompt for an override)

setup/steps/configure_log_location.go
```diff
@@ -0,0 +1,163 @@
...
+		return state.SaveConfig()
+	},
+}
+
+func validatePath(ans interface{}) error {
+	ansStr, ok := ans.(string)
+	if !ok {
+		return errors.New("expected string value")
+	}
+	// TODO: also confirm this is readable by the regular pganalyze user
```
lfittl, line 74: For now I think its fine to not check permissions here - in the future we should have a follow-up step that checks permissions (but we can do this as follow-up work, since as discussed on Ubuntu-based systems the logs should be accessible for the `adm` group)

setup/steps/configure_log_min_duration_statement.go
```diff
@@ -0,0 +1,57 @@
...
+	Kind:        state.LogInsightsStep,
+	Description: "Check log_min_duration_statement",
+	Check: func(state *s.SetupState) (bool, error) {
+		row, err := state.QueryRunner.QueryRow(`SELECT setting FROM pg_settings WHERE name = 'log_min_duration_statement'`)
+		if err != nil {
+			return false, err
+		}
+
+		lmdsVal := row.GetInt(0)
+		needsUpdate := (lmdsVal < 10 && lmdsVal != -1) ||
```
lfittl, line 24: Could we get the constant (`10`) from the main collector instead? (so we don't forget to update it)

(The same probably applies for `auto_explain.log_min_duration`)

setup/steps/configure_log_min_duration_statement.go
```diff
@@ -0,0 +1,57 @@
...
+		if err != nil {
+			return err
+		}
+		oldVal := fmt.Sprintf("%sms", row.GetString(0))
+
+		var newVal string
+		if state.Inputs.Scripted {
+			if !state.Inputs.GUCS.LogMinDurationStatement.Valid {
+				return errors.New("log_min_duration_statement not provided and current value is unsupported")
+			}
```
lfittl, line 41: We should verify that the value is a supported value here again (i.e. not between 0 and 10)

setup/steps/configure_log_min_duration_statement.go
```diff
@@ -0,0 +1,57 @@
...
+			newVal = strconv.Itoa(int(state.Inputs.GUCS.LogMinDurationStatement.Int64))
+		} else {
+			err = survey.AskOne(&survey.Input{
+				Message: fmt.Sprintf(
+					"Setting 'log_min_duration_statement' is set to '%s', below supported threshold of 10ms; enter supported value in ms or 0 to disable (will be saved to Postgres):",
```
lfittl, line 46: `-1 to disable` :-)

setup/steps/configure_log_statement.go
```diff
@@ -0,0 +1,47 @@
...
+				currValue != state.Inputs.GUCS.LogStatement.String)
+
+		return !needsUpdate, nil
+	},
+	Run: func(state *s.SetupState) error {
+		var newVal string
+		if state.Inputs.Scripted {
+			if !state.Inputs.GUCS.LogStatement.Valid {
+				return errors.New("log_statement value not provided and current value not supported")
+			}
```
lfittl, line 33: Want to check this again here as well (that its not `log_statement = all`)

setup/steps/configure_monitoring_user_password.go
```diff
@@ -0,0 +1,69 @@
...
+
+	survey "github.com/AlecAivazis/survey/v2"
+	s "github.com/pganalyze/collector/setup/state"
+)
+
+var ConfigureMonitoringUserPasswd = &s.Step{
+	Description: "Configure monitoring user password",
+	Check: func(state *s.SetupState) (bool, error) {
+		hasPassword := state.CurrentSection.HasKey("db_password")
+		return hasPassword, nil
```
lfittl, line 17: Could probably simplify this into one line (like we've done in other checks)

setup/steps/configure_monitoring_user_password.go
```diff
@@ -0,0 +1,69 @@
...
+				passwordStrategy = 0
+			} else if state.Inputs.Settings.DBPassword.Valid && state.Inputs.Settings.DBPassword.String != "" {
+				passwordStrategy = 1
+			} else {
+				return errors.New("no db_password specified and generate_monitoring_password flag not set")
+			}
+		} else {
+			err := survey.AskOne(&survey.Select{
+				Message: "Select how to set up the collector user password (will be saved to collector config):",
+				Options: []string{"generate random password (recommended)", "enter existing password"},
```
lfittl, line 35: Maybe `enter password` (instead of `enter existing password`, since we can set the password for you)

setup/steps/confirm_auto_explain_setup.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/guregu/null"
+	"github.com/pganalyze/collector/setup/state"
+	s "github.com/pganalyze/collector/setup/state"
+	"github.com/pganalyze/collector/setup/util"
+)
+
+var ConfirmAutoExplainSetup = &s.Step{
+	// N.B.: this step, asking the user whether to set up automated explain, is *not* an AutomatedExplainStep
+	// itself, but it is a state.LogInsightsStep because it depends on log insights
+	Kind:        state.LogInsightsStep,
```
lfittl, line 16: Side note: It could be helpful to prefix the files with the kind of step, so they are grouped together in the `steps` directory.

setup/steps/confirm_auto_explain_setup.go
```diff
@@ -0,0 +1,67 @@
...
+now, or proceed to configuring the optional Automated EXPLAIN feature. Automated
+EXPLAIN will require either setting up the auto_explain module (recommended) or
+creating helper functions in all monitored databases. The auto_explain module has
+minimal impact on most query workloads with our recommended settings; we will review
+these during setup.
+
+Learn more at https://pganalyze.com/postgres-explain
+`)
+		var setUpExplain bool
+		err := survey.AskOne(&survey.Confirm{
```
lfittl, line 56: If doing a scripted install, we should't show this prompt (i.e. have a better error than just the closed stdin)

setup/steps/confirm_log_insights_setup.go
```diff
@@ -0,0 +1,37 @@
...
+configuring the optional Log Insights feature. Log Insights will require
+specifying your database log file (we may be able to detect this), and
+may require changes to some logging-related settings.
+
+Setting up Log Insights is required for the Automated EXPLAIN feature.
+
+Learn more at https://pganalyze.com/log-insights
+`)
+		var setUpLogInsights bool
+		err := survey.AskOne(&survey.Confirm{
```
lfittl, line 26: We should check explicitly for scripted setup

setup/steps/create_log_explain_helper.go
```diff
@@ -0,0 +1,151 @@
...
+			if !isValid || err != nil {
+				return isValid, err
+			}
+		}
+		return true, nil
+	},
+	Run: func(state *s.SetupState) error {
+		var doCreate bool
+		if state.Inputs.Scripted {
+			if !state.Inputs.CreateExplainHelper.Valid || !state.Inputs.CreateHelperFunctions.Bool {
```
lfittl, line 41: `if !state.Inputs.CreateExplainHelper.Valid || !state.Inputs.CreateExplainHelper.Bool {`

setup/steps/create_log_explain_helper.go
```diff
@@ -0,0 +1,151 @@
...
+				return errors.New("create_explain_helper flag not set and helper function does not exist or does not match expected signature")
+			}
+			doCreate = state.Inputs.CreateHelperFunctions.Bool
```
lfittl, line 44: `doCreate = state.Inputs. CreateExplainHelper.Bool`

setup/steps/create_log_explain_helper.go
```diff
@@ -0,0 +1,151 @@
...
+				return isValid, err
+			}
+		}
+		return true, nil
+	},
+	Run: func(state *s.SetupState) error {
+		var doCreate bool
+		if state.Inputs.Scripted {
+			if !state.Inputs.CreateExplainHelper.Valid || !state.Inputs.CreateHelperFunctions.Bool {
+				return errors.New("create_explain_helper flag not set and helper function does not exist or does not match expected signature")
```
lfittl, line 42: `does not exist on all monitored databases`

setup/steps/create_log_explain_helper.go
```diff
@@ -0,0 +1,151 @@
...
+		return err
+	}
+	pgaUser := userKey.String()
+
+	return dbRunner.Exec(
+		fmt.Sprintf(
+			`CREATE SCHEMA IF NOT EXISTS pganalyze; GRANT USAGE ON SCHEMA pganalyze TO %s;`,
+			pq.QuoteIdentifier(pgaUser),
+		) + `
+CREATE OR REPLACE FUNCTION pganalyze.explain(query text, params text[]) RETURNS text AS
```
lfittl, line 92: Might be slightly better if we move this to the same location as the helper checksums (`util.ValidateHelperFunction`)

setup/steps/create_pganalyze_schema.go
```diff
@@ -0,0 +1,84 @@
...
+		if state.Inputs.Scripted {
+			if !state.Inputs.CreateHelperFunctions.Valid || !state.Inputs.CreateHelperFunctions.Bool {
+				return errors.New("create_helper_functions flag not set and pganalyze schema or helper functions do not exist")
+			}
+			doSetup = state.Inputs.CreateHelperFunctions.Bool
+		} else {
+			err := survey.AskOne(&survey.Confirm{
+				Message: "Create pganalyze schema and helper functions (will be saved to Postgres)?",
+				Default: false,
+				// TODO: better link?
```
lfittl, line 58: It would be very good to not link to the collector README here - maybe this is a reason enough to just copy the README contents to our own docs and link to that?

setup/steps/create_pganalyze_schema.go
```diff
@@ -0,0 +1,84 @@
...
+		userKey, err := state.CurrentSection.GetKey("db_username")
+		if err != nil {
+			return err
+		}
+		pgaUser := userKey.String()
+
+		return state.QueryRunner.Exec(fmt.Sprintf(`CREATE SCHEMA IF NOT EXISTS pganalyze;
+GRANT USAGE ON SCHEMA pganalyze TO %s;
+
+CREATE OR REPLACE FUNCTION pganalyze.get_stat_replication() RETURNS SETOF pg_stat_replication AS
```
lfittl, line 79: This helper function definition should also be colocated with the `util.ValidateHelperFunction` checksums

setup/steps/create_pgss.go
```diff
@@ -0,0 +1,54 @@
...
+	Run: func(state *s.SetupState) error {
+		var doCreate bool
+		if state.Inputs.Scripted {
+			if !state.Inputs.CreatePgStatStatements.Valid || !state.Inputs.CreatePgStatStatements.Bool {
+				return errors.New("create_pg_stat_statements flag not set and pg_stat_statements does not exist in primary database")
+			}
+			doCreate = state.Inputs.CreatePgStatStatements.Bool
+		} else {
+			err := survey.AskOne(&survey.Confirm{
+				Message: "Create extension pg_stat_statements in public schema for performance monitoring (will be saved to Postgres)?",
```
lfittl, line 40: maybe `query performance monitoring` for clarity

setup/steps/enable_auto_explain.go
```diff
@@ -0,0 +1,59 @@
...
+	Run: func(state *s.SetupState) error {
+		var doAdd bool
+		if state.Inputs.Scripted {
+			if !state.Inputs.EnableAutoExplain.Valid || !state.Inputs.EnableAutoExplain.Bool {
+				return errors.New("enable_auto_explain flag not set but auto_explain configuration selected")
+			}
+			doAdd = state.Inputs.EnableAutoExplain.Bool
+		} else {
+			err := survey.AskOne(&survey.Confirm{
+				Message: "Add auto_explain to shared_preload_libraries (will be saved to Postgres)?",
```
lfittl, line 35: `Add auto_explain to shared_preload_libraries (will be saved to Postgres - requires a restart)?`

setup/steps/enable_pgss.go
```diff
@@ -0,0 +1,56 @@
...
+	Run: func(state *s.SetupState) error {
+		var doAdd bool
+		if state.Inputs.Scripted {
+			if !state.Inputs.EnablePgStatStatements.Valid || !state.Inputs.EnablePgStatStatements.Bool {
+				return errors.New("enable_pg_stat_statements flag not set but pg_stat_statements not in shared_preload_libraries")
+			}
+			doAdd = state.Inputs.EnablePgStatStatements.Bool
+		} else {
+			err := survey.AskOne(&survey.Confirm{
+				Message: "Add pg_stat_statements to shared_preload_libraries (will be saved to Postgres)?",
```
lfittl, line 31: `Add pg_stat_statements to shared_preload_libraries (will be saved to Postgres - requires a restart)?`

setup/steps/establish_superuser_connection.go
```diff
@@ -0,0 +1,152 @@
...
+					"no Postgres server found listening on %d%s",
+					state.Inputs.PGSetupConnPort.Int64,
+					socketDirStr,
+				)
+			}
+		} else {
+			if len(localPgs) == 0 {
+				return errors.New("failed to find a running local Postgres install")
+			} else if len(localPgs) == 1 {
+				selectedPg = localPgs[0]
```
lfittl, line 58: Might be good if we move this up, so its also run for scripted installs (i.e. clear error message when there are no local Postgres installs, and don't require specifying a port when there is only a single Postgres install detected)

setup/steps/establish_superuser_connection.go
```diff
@@ -0,0 +1,152 @@
...
+	SocketDir string
+	LocalAddr string
+	Port      int
+}
+
+var pgsqlDomainSocketPortRe = regexp.MustCompile("\\d+$")
+
+func getSocketDirMatches(dir string) ([]LocalPostgres, error) {
+	var result []LocalPostgres
+	// technically this should be a filepath.Join, but Unix-domain sockets do not work
```
lfittl, line 118: Technically, Unix-domain sockets work on Windows these days (due to fixes in Postgres), so maybe just say `but this code path doesn't run on Windows` or something like that :)

setup/steps/establish_superuser_connection.go
```diff
@@ -0,0 +1,152 @@
...
+			return nil, err
+		}
+		result = append(result, LocalPostgres{SocketDir: dir, Port: port})
+	}
+	return result, nil
+}
+
+func discoverLocalPostgres() ([]LocalPostgres, error) {
+	// TODO: find tcp sockets if no unix sockets?
+	// TODO: confirm these are live by checking pids?
```
lfittl, line 141: We can remove these TODOs as discussed - maybe clarify the function name by calling it `discoverLocalPostgresFromUnixDomainSockets` (or similar)

setup/steps/restart_postgres.go
```diff
@@ -0,0 +1,67 @@
...
+		var pendingSettings []string
+		for _, row := range rows {
+			pendingSettings = append(pendingSettings, row.GetString(0))
+		}
+
+		pendingList := util.JoinWithAnd(pendingSettings)
+		var restartNow bool
+		if state.Inputs.Scripted {
+			if !state.Inputs.ConfirmPostgresRestart.Valid || !state.Inputs.ConfirmPostgresRestart.Bool {
+				return errors.New("confirm_postgres_restart flag not set but Postgres restart required")
```
lfittl, line 36: `but Postgres restart required for settings %s to take effect` + list out the settings (to help with debugging)

setup/steps/run_pg_sleep.go
```diff
@@ -0,0 +1,57 @@
...
+	Check: func(state *s.SetupState) (bool, error) {
+		needsSleep := (state.Inputs.SkipLogInsights.Valid && !state.Inputs.SkipLogInsights.Bool) ||
+			(state.Inputs.SkipAutomatedExplain.Valid && !state.Inputs.SkipAutomatedExplain.Bool)
+		return !needsSleep || state.DidPgSleep ||
+			(state.Inputs.SkipPgSleep.Valid && state.Inputs.SkipPgSleep.Bool), nil
+	},
+	Run: func(state *s.SetupState) error {
+		var doPgSleep bool
+		if state.Inputs.Scripted {
+			if state.Inputs.SkipPgSleep.Valid {
```
lfittl, line 22: Might be better if we call this `RunPgSleep` (or similar) and flip the logic. That way, we run the pg_sleep test only if that is set (and skip otherwise, i.e. if false or not set, do the skip)

setup/steps/run_pg_sleep.go
```diff
@@ -0,0 +1,57 @@
...
+			}, &doPgSleep)
+			if err != nil {
+				return err
+			}
+			state.Inputs.SkipPgSleep = null.BoolFrom(!doPgSleep)
+		}
+
+		if !doPgSleep {
+			return nil
+		}
```
lfittl, line 46: We should check whether `log_min_duration_statement` / `auto_explain.log_min_duration` is set to -1 (and raise a notice in that case, instead of running the query below)

setup/steps/select_databases.go
```diff
@@ -0,0 +1,137 @@
...
+					for _, opt := range dbOpts {
+						if trimmed == opt {
+							dbNames = append(dbNames, trimmed)
+							break
+						}
+					}
+				}
+
+				if len(dbNames) != i+1 {
+					return fmt.Errorf("database %s not found", trimmed)
```
lfittl, line 65: `database %s configured for db_name, but not found in Postgres`

setup/steps/select_databases.go
```diff
@@ -0,0 +1,137 @@
...
+			}, &primaryDb)
+			if err != nil {
+				return err
+			}
+
+			dbNames = append(dbNames, primaryDb)
+			if len(dbOpts) == 1 {
+				var monitorAll bool
+				err = survey.AskOne(&survey.Confirm{
+					Message: "Monitor other databases created in the future (will be saved to collector config)?",
```
lfittl, line 83: `Monitor all other databases created in the future` (add the `all`)

setup/steps/set_up_monitoring_user.go
```diff
@@ -0,0 +1,72 @@
...
+var SetUpMonitoringUser = &s.Step{
+	Description: "Set up monitoring user",
+	Check: func(state *s.SetupState) (bool, error) {
+		pgaUserKey, err := state.CurrentSection.GetKey("db_username")
+		if err != nil {
+			return false, err
+		}
+		pgaUser := pgaUserKey.String()
+
+		// TODO: deal with postgres <10
```
lfittl, line 23: We can probably drop this TODO since we don't support anything older than 10

setup/steps/set_up_monitoring_user.go
```diff
@@ -0,0 +1,72 @@
...
+		row, err := state.QueryRunner.QueryRow(
+			fmt.Sprintf(
+				"SELECT usesuper OR pg_has_role(usename, 'pg_monitor', 'member') FROM pg_user WHERE usename = %s",
```
lfittl, line 26: As discussed, we might want to have `pg_has_role(usename, 'pg_monitor', 'usage')` here instead

setup/steps/set_up_monitoring_user.go
```diff
@@ -0,0 +1,72 @@
...
+		return row.GetBool(0), nil
+	},
+	Run: func(state *s.SetupState) error {
+		pgaUserKey, err := state.CurrentSection.GetKey("db_username")
+		if err != nil {
+			return err
+		}
+		pgaUser := pgaUserKey.String()
+
+		// TODO: deal with postgres <10
```
lfittl, line 45: We can drop this TODO

setup/util/util.go
```diff
@@ -0,0 +1,154 @@
...
+            name || ' = ', ''
+          ),
+      -1),
+    -1)
+    ELSE current_setting(name)
+  END AS pending_value
+FROM
+  pg_settings
+WHERE
+  name = 'shared_preload_libraries'`)
```
lfittl, line 127: TODO: @lfittl to review this once more

setup/steps/check_pgss_available.go
uhoh-itsmaciek: Ended up setting search path before every query in `query/runner.go` (see ef8c5d2cf2363459351cabf9434d450cb7110582 once it's pushed)..

setup/util/util.go
lfittl: Reviewing this, I can follow the logic, however I think there is a bug here:

It seems that we skip over empty lines in the inner query:

```
SELECT row_number() OVER () AS line_no, line FROM
                regexp_split_to_table(
                  pg_read_file(
                    COALESCE(sourcefile, 'postgresql.auto.conf')
                  ), '\s*$\s*', 'm'
                ) AS lines(line)
```

And this then throws off the `line_no = sourceline` comparison in the outer SELECT. I think we need to preserve empty lines there, so the line number comparison works.

setup/steps/establish_superuser_connection.go
uhoh-itsmaciek: Come to think of it, there's no good reason not to use Join here so just did that.

setup/steps/li_confirm_set_up_auto_explain.go
```diff
@@ -0,0 +1,76 @@
...
+	Kind:        state.LogInsightsStep,
+	Description: "Confirm whether to set up the optional Automated EXPLAIN feature",
+	Check: func(state *s.SetupState) (bool, error) {
+		// skip the question if we've already answered one way or the other
+		if state.Inputs.ConfirmSetUpAutomatedExplain.Valid {
+			return true, nil
+		}
+
+		// also skip the question if it's already set up
+		if state.CurrentSection.HasKey("enable_log_explain") {
```
lfittl, line 26: I think there is a bug here, because we return early if the key exists, making the lines that follow not work if the key doesn't exist:

```
* Confirm whether to set up the optional Automated EXPLAIN feature
  ✗ step check failed: error when getting key of section "server1": key "enable_log_explain" not exists
```

I assume the intent was something like this?

```golang
                if state.CurrentSection.HasKey("enable_log_explain") {
                        isLogExplainKey, err := state.CurrentSection.GetKey("enable_log_explain")
                        if err != nil {
                                return false, err
                        }
                        isLogExplain, err := isLogExplainKey.Bool()
                        if err != nil {
                                return false, err
                        }
                        if isLogExplain {
                                return true, nil
                        }
                }

                // assume auto_explain if we got this far
```

setup/util/util.go
lfittl: It seems there is another problem here as well, related to how Postgres determines line numbers, or rather when it sets the `sourceline` value. It seems we can't trust what we're getting from `pg_settings`.

postgresql.auto.conf:

```
# Do not edit this file manually!
# It will be overwritten by the ALTER SYSTEM command.
log_line_prefix = '%m [%p] %q[user=%u,db=%d,app=%a] '
shared_preload_libraries = 'pg_stat_statements, auto_explain'
```

```
postgres=# SELECT                  
  name, sourcefile, sourceline FROM
  pg_settings
WHERE
  name = 'log_line_prefix';
      name       |                    sourcefile                    | sourceline 
-----------------+--------------------------------------------------+------------
 log_line_prefix | /var/lib/postgresql/13/main/postgresql.auto.conf |          3
(1 row)

postgres=# SELECT                  
  name, sourcefile, sourceline FROM
  pg_settings
WHERE
  name = 'shared_preload_libraries';
           name           |                    sourcefile                    | sourceline 
--------------------------+--------------------------------------------------+------------
 shared_preload_libraries | /var/lib/postgresql/13/main/postgresql.auto.conf |          3
(1 row)
```

The `shared_preload_libraries` value here is of course wrong, which then leads to us getting the wrong value, and this query not working. Interestingly, after a restart this was fixed. It seems this may be related to subsequent `ALTER SYSTEM` commands not updating the `sourceline` correctly? This was with the regular setup flow, which first sets `shared_preload_libraries = 'pg_stat_statements'` (which would make it line 3), then sets `log_line_prefix` (which presumably adds it first in line 4), and then again `shared_preload_libraries = 'auto_explain, pg_stat_statements'`, which causes the order to change, but apparently doesn't update the `sourceline` value correctly.

If we only match based on the line pattern, this works as expected:

```
SELECT
  name,
  CASE
    WHEN pending_restart THEN
      left(
        right(
          regexp_replace(
            (SELECT line FROM
              (SELECT row_number() OVER () AS line_no, line FROM
                regexp_split_to_table(
                  pg_read_file(
                    COALESCE(sourcefile, 'postgresql.auto.conf')
                  ), '\s*$\s*', 'm'
                ) AS lines(line)
              ) AS numbered_lines(line_no, line)
             WHERE
               line LIKE name || ' = %'                                                                
            ),
            name || ' = ', ''
          ),
      -1),
    -1)
    ELSE current_setting(name)
  END AS pending_value
FROM
  pg_settings
WHERE
  name = 'shared_preload_libraries';

           name           |          pending_value           
--------------------------+----------------------------------
 shared_preload_libraries | pg_stat_statements, auto_explain
(1 row)
```

setup/steps/confirm_test_command.go
```diff
@@ -0,0 +1,75 @@
...
+			var features string
+			if hasAutomatedExplain {
+				features = "Log Insights and Automated EXPLAIN"
+			} else {
+				features = "Log Insights"
+			}
+			err := survey.AskOne(&survey.Confirm{
+				Message: fmt.Sprintf("Run a test command (%s) to confirm %s configuration?", testCmdType, features),
+				Default: true,
+				Help:    "You should see results in pganalyze a few seconds after the query completes",
```
lfittl, line 50: This here is a bit tricky, because in the initial setup, you haven't actually reloaded the collector yet (nor sent a test snapshot), so this doesn't actually work as intended.

If the customer goes fast enough, and runs the test command within 1 minute of the setup finishing, they would still get a benefit from this (since we go back 1 minute in the log tail), but not sure if we can rely on that.

It might be better if we move this logic to the test command in the main collector code instead? (especially the pg_sleep for verifying the EXPLAIN setup would be very useful there -- and I think we already issue the NOTICE, so I'm not sure if we need to do anything in that regard)

lfittl (approved):
This looks good - I noticed some more minor things when doing a final check, but I think we can ship it after those are resolved. 💯

setup/steps/confirm_test_command.go
uhoh-itsmaciek: We do reload the collector in an earlier step. Is that not enough without the test snapshot? You're right that consolidating this is probably a good idea, though. Should we remove this step entirely (though merging the sleep functionality into the collector test command as you suggest)?

lfittl: Ah, I missed that - that would be enough to make it show up, since we run this step after the reload.

Since we also have you run the test command though, maybe still makes sense to drop it here and only doing in the test command? (but I'm also good to leave as is)

setup/util/util.go
uhoh-itsmaciek: Yeah, you're right--I think the issue is that `pg_settings` has the `sourcefile` and `sourceline` of the *current* setting. Essentially it's metadata about the snapshot of the config file from which the setting as applied was loaded (and when we read it, the contents may be different than the snapshot). We `pg_reload_conf` after every change so this is not an issue *except* in the very important case of `shared_preload_libraries` post change but before restart.

I think the solution is as you're suggesting: skip the line number and just match on the name. I was using that as a fallback because I thought the line number method was higher fidelity, but I did not account for the subtle semantics here.

uhoh-itsmaciek: Thinking about this some more, even `sourcefile` can be bogus here, right? Consider the following:

1. I have an existing `shared_preload_libraries = some_module` manually set in `postgresql.conf`
2. I run the `pg_stat_statements` step which correctly finds that based on `sourcefile`, adds `pg_stat_statements`, and saves the resulting `some_module, pg_stat_statements` to `postgresql.auto.conf` using `ALTER SYSTEM`
3. I run the `auto_explain` step, which I think at this point (pre-restart), will still incorrectly see `postgresql.conf` as the `sourcefile` (because it's the source of the current pre-restart value), grab the out-of-date value, and incorrectly save the new value of `some_module, auto_explain`, losing `pg_stat_statements`.

Maybe given this, the right thing to do here is if the key is set in postgresql.auto.conf, always use that value, and only fall back to `sourcefile` when it is not present there?

setup/steps/confirm_test_command.go
uhoh-itsmaciek: Ok, let's drop it then and move the sleep logic to the test command--I think that makes more sense. Should we drop the collector reload step then as well (and rely on the test command for that)?

integration_test/Makefile
```diff
@@ -211,3 +211,30 @@ pg13:
...
+
+	docker exec --user postgres pganalyze-collector-test pg_ctl --pgdata /var/lib/postgresql/10/main --wait stop --mode fast
+	docker rm -f pganalyze-collector-test
+	docker rmi pganalyze-collector-test
+	diff -Nau guided-setup.postgres.expected guided-setup.postgres.out && echo 'postgres configuration modified as expected'
+	grep -v db_password guided-setup.collector-config.out > guided-setup.collector-config.nopass.out
+	diff -Nau guided-setup.collector-config.nopass.expected guided-setup.collector-config.nopass.out \
+	  && grep --extended-regexp --quiet 'db_password\s+=\s+[a-f0-9]{32}' guided-setup.collector-config.out && echo 'collector config file modified as expected'
+	jq '{fns: .functionReferences, roles: .roleReferences[] |  select(.name == "pganalyze")}' < guided-setup.snapshot.json.out > guided-setup.snapshot-subset.json.out
+	diff -Nau guided-setup.snapshot-subset.json.expected guided-setup.snapshot-subset.json.out && echo 'success'
```
uhoh-itsmaciek, line 240: @lfittl I was thinking about what to verify here, and since we're testing the setup program, I did a `--dry-run` smoke test to ensure that the collector will run, but I mostly wanted to test that Postgres and the config file are actually set up correctly.

One issue I ran into is that the current `pg_ctl restart` mechanism somehow kills Postgres logging, so the log test fails: log output stops when the server shuts down, and does not resume again when the server restarts. I think this has to do with Go's handling of stdout/stderr in `exec.Command`, but I'm not sure how to fix this. I tried throwing away my hacky stdout/stderr mechanism and just install sudo as we'd discussed, but I hit the same problem. The test does still run, but it's not able to validate log setup.

uhoh-itsmaciek: Tried `nohup` as well just in case, but no dice. I also tried setting both Stderr and Stdout to `ioutil.Discard`, but also no dice.

setup/steps/confirm_test_command.go
lfittl: @uhoh-itsmaciek Good question, I'm actually not sure whether we want to keep the reload step in the setup itself.

The benefit of keeping the reload step is that in case someone doesn't run the test command, or forgets to run it with reload, it would still give them data in-app.

That actually makes me think again, maybe we should change the in-app experience to be an explicit "Check for new test snapshots" button (or similar), and actually run the `--test --reload` command for you at the end of the setup?

(our main motivation for not doing so previously was that we were worried the in-app experience would have timed out already - maybe we can avoid that problem & make it one step overall)

integration_test/Makefile
lfittl: We may be able to use systemd after all, if it helps us avoid the `pg_ctl` issue here.

I explored that in https://github.com/pganalyze/collector/pull/116/commits/6080a1bc856cf159949ad6cac8f93d77c54b82c0 based on some tricks I found. That seems to work for both Ubuntu 18.04 image (as tested there) and Ubuntu 20.04.

Maybe that is the right approach after all? (note I didn't test the log aspect yet, but does run Postgres using systemd successfully)

uhoh-itsmaciek:
@lfittl I'll do a little more manual testing, but I think this is ready now if you have any feedback on the last changes.

uhoh-itsmaciek:
Did some more manual tests: fixed a minor issue, and also ran into this:

```
* Invoke the collector self-test to verify the installation
? The collector is now ready to begin monitoring. Run test command and reload collector configuration if successful? Yes
  
2020/12/04 21:52:31 I Running collector test with pganalyze-collector 0.34.0
2020/12/04 21:52:31 I [server1] Testing statistics collection...
2020/12/04 21:52:32 E [server1] Could not process server: Error when getting grant: ERROR: Unauthorized, invalid API key
2020/12/04 21:52:32 I [server1] Testing activity snapshots...
2020/12/04 21:52:40 E [server1] Could not collect activity for server: could not get default grant for activity snapshot: Error when getting grant: ERROR: Unauthorized, invalid API key
2020/12/04 21:52:40 I [server1] Testing log collection (local)...
2020/12/04 21:52:46 I [server1]   Local log test successful
2020/12/04 21:52:46 E Error: Reload requested, but ignoring since configuration errors are present
  
    re-checking...
  ✓ step completed


Collector setup complete!
```

The self-test plays a little fast and loose with exit codes, so we can have a situation where the self-test fails but we congratulate you. Should we ensure we exit non-zero in these situations to make this less confusing?

lfittl:
> Did some more manual tests: fixed a minor issue, and also ran into this:
> 
> ```
> * Invoke the collector self-test to verify the installation
> ? The collector is now ready to begin monitoring. Run test command and reload collector configuration if successful? Yes
>   
> 2020/12/04 21:52:31 I Running collector test with pganalyze-collector 0.34.0
> 2020/12/04 21:52:31 I [server1] Testing statistics collection...
> 2020/12/04 21:52:32 E [server1] Could not process server: Error when getting grant: ERROR: Unauthorized, invalid API key
> 2020/12/04 21:52:32 I [server1] Testing activity snapshots...
> 2020/12/04 21:52:40 E [server1] Could not collect activity for server: could not get default grant for activity snapshot: Error when getting grant: ERROR: Unauthorized, invalid API key
> 2020/12/04 21:52:40 I [server1] Testing log collection (local)...
> 2020/12/04 21:52:46 I [server1]   Local log test successful
> 2020/12/04 21:52:46 E Error: Reload requested, but ignoring since configuration errors are present
>   
>     re-checking...
>   ✓ step completed
> 
> 
> Collector setup complete!
> ```
> 
> The self-test plays a little fast and loose with exit codes, so we can have a situation where the self-test fails but we congratulate you. Should we ensure we exit non-zero in these situations to make this less confusing?

Yeah, if its straightforward to do, it would be great if we can fix the test behavior and return an error code here (and thus an error (or warning?) message instead of "Collector setup complete!".

lfittl (approved):
Looks good, including the recent changes.

PR #114, opened 2020-11-26
Update docs on CREATE EXTENSION to match website

joehorsnell:
I noticed that the [instructions on pganalyze.com][1] recommend the explicit public schema when
creating the `pg_stat_statements` extension, but this README does not. Makes sense for them to be
consistent?

Also, probably makes sense to do the same for `pg_buffercache`?

[1]: https://pganalyze.com/docs/install/amazon_rds/02_create_monitoring_user

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ to run as a superuser, you can setup a separate monitoring user like this:
 ```
 CREATE SCHEMA pganalyze;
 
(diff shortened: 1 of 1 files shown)


lfittl (approved):
@joehorsnell Good catch - thanks for the PR!

PR #115, opened 2020-12-02
Add pganalyze system user to adm group in Debian/Ubuntu packages

uhoh-itsmaciek:
This gives it permission to read Postgres log files in a default
install, simplifying Log Insights setup.

The adm group according to the Debian wiki [1]:

  Group adm is used for system monitoring tasks. Members of this
  group can read many log files in /var/log, and can use xconsole

[1]: https://wiki.debian.org/SystemGroups


lfittl (approved):
✅

PR #117, opened 2020-12-04
Add a skip_if_replica / SKIP_IF_REPLICA option

uhoh-itsmaciek:
This allows users to configure the collector in a no-op mode on
replicas (we only query if the monitored database is a replica), and
automatically switch to active monitoring when the database is no
longer a replica.

### DIFF
--- a/input/full.go
+++ b/input/full.go
@@ -16,21 +16,7 @@ import (
 // CollectFull - Collects a "full" snapshot of all data we need on a regular interval
 func CollectFull(server *state.Server, connection *sql.DB, globalCollectionOpts state.CollectionOpts, logger *util.Logger) (ps state.PersistedState, ts state.TransientState, err error) {
 	systemType := server.Config.SystemType
-
 	ps.CollectedAt = time.Now()
-
-	if server.Config.SkipIfReplica {
-		var isReplica bool
-		isReplica, err = postgres.GetIsReplica(logger, connection)
-		if err != nil {
-			logger.PrintError("Error checking replication status")
-			return
-		}
-		if isReplica {
-			err = state.ErrReplicaCollectionDisabled
-			return
-		}
-	}
 
 	ts.Version, err = postgres.GetPostgresVersion(logger, connection)
 	if err != nil {
--- a/input/full.go
+++ b/input/full.go
@@ -19,6 +19,19 @@ func CollectFull(server *state.Server, connection *sql.DB, globalCollectionOpts 
 
 	ps.CollectedAt = time.Now()
 
+	if server.Config.SkipIfReplica {
+		var isReplica bool
+		isReplica, err = postgres.GetIsReplica(logger, connection)
+		if err != nil {
+			logger.PrintError("Error checking replication status")
+			return
+		}
+		if isReplica {
+			err = state.ErrReplicaCollectionDisabled
+			return
+		}
+	}
+
 	ts.Version, err = postgres.GetPostgresVersion(logger, connection)
 	if err != nil {
 		logger.PrintError("Error collecting Postgres Version")
@@ -67,7 +80,7 @@ func CollectFull(server *state.Server, connection *sql.DB, globalCollectionOpts 
 	}
 
 	if globalCollectionOpts.CollectPostgresSettings {
-		ts.Settings, err = postgres.GetSettings(connection, ts.Version)
+		ts.Settings, err = postgres.GetSettings(connection)
 		if err != nil {
 			logger.PrintError("Error collecting config settings")
 			return
--- a/state/state.go
+++ b/state/state.go
@@ -1,7 +1,6 @@
 package state
 
 import (
-	"errors"
 	"sync"
 	"time"
 
@@ -216,8 +215,6 @@ type GrantS3 struct {
 	S3URL    string            `json:"s3_url"`
 	S3Fields map[string]string `json:"s3_fields"`
 }
-
-var ErrReplicaCollectionDisabled error = errors.New("monitored server is replica and replication collection disabled via config")
 
 type CollectionStatus struct {
 	CollectionDisabled        bool
--- a/state/state.go
+++ b/state/state.go
@@ -1,6 +1,7 @@
 package state
 
 import (
+	"errors"
 	"sync"
 	"time"
 
@@ -216,7 +217,11 @@ type GrantS3 struct {
 	S3Fields map[string]string `json:"s3_fields"`
 }
 
+var ErrReplicaCollectionDisabled error = errors.New("monitored server is replica and replication collection disabled via config")
+
 type CollectionStatus struct {
+	CollectionDisabled        bool
+	CollectionDisabledReason  string
 	LogSnapshotDisabled       bool
 	LogSnapshotDisabledReason string
 }
--- a/config/config.go
+++ b/config/config.go
@@ -143,6 +143,11 @@ type ServerConfig struct {
 	// the collector multiple times against the same database server
 	MaxCollectorConnections int `ini:"max_collector_connections"`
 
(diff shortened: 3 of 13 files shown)


uhoh-itsmaciek:
Tested locally by mocking out `pg_is_in_recovery`

uhoh-itsmaciek:
Output looks like:

```
2020/12/04 14:28:35 V [server1] Identified as api_system_type: self_hosted, api_system_scope: 5432/mydb1*, api_system_id: starling
2020/12/04 14:28:35 V [server1] Successfully recovered state from on-disk file
2020/12/04 14:28:35 I [server1] All monitoring disabled for this server: monitored server is replica and replication collection disabled via config
2020/12/04 14:28:35 V [server1] Searching for log file(s) in /var/log/postgresql/postgresql-11-main.log
2020/12/04 14:28:35 V Scheduled next run for full snapshot of all servers in 1m24.601508004s
2020/12/04 14:28:35 V [server1] Tailing log file /var/log/postgresql/postgresql-11-main.log
2020/12/04 14:28:35 V Scheduled next run for high frequency query statistics of all servers in 24.601264199s
2020/12/04 14:28:35 V Scheduled next run for activity snapshot of all servers in 4.601233248s
2020/12/04 14:28:40 V [server1] All monitoring suspended while server is replica
2020/12/04 14:28:40 V Scheduled next run for activity snapshot of all servers in 9.922314928s
2020/12/04 14:28:50 V [server1] All monitoring suspended while server is replica
2020/12/04 14:28:50 V Scheduled next run for activity snapshot of all servers in 9.941179777s
2020/12/04 14:29:00 V [server1] All monitoring suspended while server is replica
2020/12/04 14:29:00 V Scheduled next run for high frequency query statistics of all servers in 59.975794971s
2020/12/04 14:29:00 V [server1] All monitoring suspended while server is replica
2020/12/04 14:29:00 V Scheduled next run for activity snapshot of all servers in 9.930939515s
2020/12/04 14:29:10 V [server1] All monitoring suspended while server is replica
2020/12/04 14:29:10 V Scheduled next run for activity snapshot of all servers in 9.935312155s
2020/12/04 14:29:20 V [server1] All monitoring suspended while server is replica
2020/12/04 14:29:20 V Scheduled next run for activity snapshot of all servers in 9.935918593s
2020/12/04 14:29:30 V [server1] All monitoring suspended while server is replica
2020/12/04 14:29:30 V Scheduled next run for activity snapshot of all servers in 9.938075303s
2020/12/04 14:29:40 V [server1] All monitoring suspended while server is replica
2020/12/04 14:29:40 V Scheduled next run for activity snapshot of all servers in 9.936422518s
2020/12/04 14:29:50 V [server1] All monitoring suspended while server is replica
2020/12/04 14:29:50 V Scheduled next run for activity snapshot of all servers in 9.941714174s
2020/12/04 14:30:00 V Skipping run for high frequency query statistics of all servers since it overlaps with primary group time
2020/12/04 14:30:00 V Scheduled next run for high frequency query statistics of all servers in 59.999801078s
2020/12/04 14:30:00 V [server1] All monitoring suspended while server is replica
2020/12/04 14:30:00 V Scheduled next run for activity snapshot of all servers in 9.953950292s
2020/12/04 14:30:00 V [server1] All monitoring suspended while server is replica
2020/12/04 14:30:00 V Scheduled next run for full snapshot of all servers in 9m59.908124716s
2020/12/04 14:30:10 V [server1] All monitoring suspended while server is replica
2020/12/04 14:30:10 V Scheduled next run for activity snapshot of all servers in 9.930410724s
2020/12/04 14:30:20 V [server1] Detected PostgreSQL Version 110010 (PostgreSQL 11.10 (Ubuntu 11.10-1.pgdg20.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0, 64-bit)
2020/12/04 14:30:20 V [server1] Successfully prepared S3 request - size of request body: 0.0027 MB
2020/12/04 14:30:20 I [server1] Submitted compact activity snapshot successfully
2020/12/04 14:30:20 V Scheduled next run for activity snapshot of all servers in 9.473659146s
2020/12/04 14:30:30 V [server1] Detected PostgreSQL Version 110010 (PostgreSQL 11.10 (Ubuntu 11.10-1.pgdg20.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0, 64-bit)
2020/12/04 14:30:30 V [server1] Successfully prepared S3 request - size of request body: 0.0027 MB
2020/12/04 14:30:30 I [server1] Submitted compact activity snapshot successfully
```

uhoh-itsmaciek:
Looking at this again, I think this message:
>2020/12/04 14:28:35 I [server1] All monitoring disabled for this server: monitored server is replica and replication collection disabled via config

could probably be improved. Thoughts? I was trying to stick with the more general collection-disabling mechanism we sketched out in #95 and #102, so the first part of this message only understands that monitoring is disabled, and the second part is the "reason" from the more specific error. Maybe I'll change to "All monitoring suspended for this server" since we expect this condition to change dynamically in the course of a collector config anyway.

input/full.go
```diff
@@ -19,6 +19,19 @@ func CollectFull(server *state.Server, connection *sql.DB, globalCollectionOpts
 
 	ps.CollectedAt = time.Now()
 
+	if server.Config.SkipIfReplica {
+		var isReplica bool
+		isReplica, err = postgres.GetIsReplica(logger, connection)
+		if err != nil {
+			logger.PrintError("Error checking replication status")
```
lfittl, line 26: I wonder if we could hit permission issues here? (not sure if `pg_is_in_recovery` is accessible to all users)

Also, probably should double check special cases, like here: https://github.com/pganalyze/collector/blob/master/input/postgres/replication.go#L86

Thus, I'm wondering if we shouldn't accept failure here, and continue with collection if we can't check replication status.

state/state.go
```diff
@@ -216,7 +217,11 @@ type GrantS3 struct {
 	S3Fields map[string]string `json:"s3_fields"`
 }
 
+var ErrReplicaCollectionDisabled error = errors.New("monitored server is replica and replication collection disabled via config")
```
lfittl, line 220: Nice - good idea to keep errors in a central place. I wonder if we should even introduce a new `state/errors.go` file for this?

lfittl:
> Looking at this again, I think this message:
> 
> > 2020/12/04 14:28:35 I [server1] All monitoring disabled for this server: monitored server is replica and replication collection disabled via config
> 
> could probably be improved. Thoughts? I was trying to stick with the more general collection-disabling mechanism we sketched out in #95 and #102, so the first part of this message only understands that monitoring is disabled, and the second part is the "reason" from the more specific error. Maybe I'll change to "All monitoring suspended for this server" since we expect this condition to change dynamically in the course of a collector config anyway.

I think the message is okay, but "All monitoring suspended for this server" is a bit more clear, so +1 to adjust it to that.

lfittl:
@uhoh-itsmaciek So one interesting problem I just realized:

With the logic here as is, we still end up getting a grant from the server, which is the part that would create a server record in the pganalyze app. We don't submit a snapshot, so the server wouldn't be considered "active", and thus we could ignore it for billing based on that.

But it does bring up the point: Do we want the ignored replicas to show up in pganalyze? If they do show up, do we want to present some level of information about them? (e.g. that they're suspended for collection)

If we have them show up in pganalyze (i.e. by getting a grant) I could see it being useful to have some way of distinguishing these from servers that failed to send any data after the test snapshot (e.g. because someone forgot to reload the collector in the background). Maybe a very simple full snapshot every 10 minutes is acceptable from an overhead perspective, to clearly identify these replicas.

lfittl (approved):
:+1:

PR #123, opened 2020-12-11
pg_stat_statements: Ignore "insufficient privilege" queries w/o queryid

lfittl:
There is no way to correctly diff these kinds of queries, because Postgres
will report one entry for each query, but won't let us see the text or
the queryid.

For Postgres 10 and newer, the pg_monitor role should be used, which
avoids this problem altogether. On older Postgres releases, we only
support monitoring a single role (that which owns the monitoring helper
function).


uhoh-itsmaciek (approved):
:+1:

PR #125, opened 2020-12-17
Report additional collector config settings

uhoh-itsmaciek:
Report additional collector config settings

This adds some more metadata about collector configuration to full
snapshots:

 - whether aws_assume_role is set
 - whether aws_account_id is set
 - the skip_if_replica setting
 - the filter_query_text setting


lfittl (approved):
:+1:

PR #126, opened 2020-12-17
Update AWS Go SDK to 1.36.10

uhoh-itsmaciek:
This reduces a timeout contacting the AWS metadata service, which is
impractically large on older versions, and can be a problem when
hitting the http-put-response-hop-limit (which is easy to do if, e.g.,
running a Docker container on a VM).

See https://github.com/aws/aws-sdk-go/issues/2972 and
https://github.com/aws/aws-sdk-go/commit/0c96ad536ae45eb23a2b5ed2b4ef9a840eec6f6e


lfittl (approved):
:+1:

PR #127, opened 2020-12-17
add necessary env var

caleb15:
I was on a call with pganalyze engineers and they recommended this var. I tested it and it worked once I set this along with [http_put_response_hop_limit](https://github.com/pganalyze/collector/issues/128) 2 :rocket:


caleb15:
@lfittl FYI and thanks for the help! :)

lfittl (approved):
Thanks @caleb15 for the PR!

I'll merge this for now, but FYI, we're likely to adjust the docs here to point to the collector settings docs page in the future: https://pganalyze.com/docs/collector/settings

PR #130, opened 2020-12-19
Fix error handling if last guided setup step fails

uhoh-itsmaciek:
We intend to ignore errors reporting setup progress, but we do need to
short-circuit when there *is* an error.


uhoh-itsmaciek:
Found via https://github.com/pganalyze/collector/runs/1579430644?check_suite_focus=true#step:5:7725

lfittl (approved):

PR #131, opened 2020-12-21
Ensure we only set collected_local_catalog_data when appropriate

uhoh-itsmaciek:
Right now, the flag indicates whether schema collection was successful
for a particular database, but we set it if we are at all able to
connect to a database, even if the actual data collection fails
afterward (we print warnings but that's not sufficient here).

Instead, only set the flag if collection of relation and function
metadata, and of relation and index stats, all succeed.

### DIFF
--- a/input/postgres/schema.go
+++ b/input/postgres/schema.go
@@ -2,6 +2,7 @@ package postgres
 
 import (
 	"database/sql"
+	"fmt"
 
 	"github.com/pganalyze/collector/state"
 	"github.com/pganalyze/collector/util"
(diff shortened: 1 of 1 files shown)


input/postgres/schema.go
```diff
@@ -27,28 +28,13 @@ func CollectAllSchemas(server *state.Server, collectionOpts state.CollectionOpts
...
-		ps.SchemaStats[databaseOid] = &state.SchemaStats{
-			RelationStats: make(state.PostgresRelationStatsMap),
-			IndexStats:    make(state.PostgresIndexStatsMap),
-		}
-
-		ps = collectSchemaData(collectionOpts, logger, schemaConnection, ps, databaseOid, dbName, ts.Version, server.Config.IgnoreSchemaRegexp)
+		ps = psNext
 		ts.DatabaseOidsWithLocalCatalog = append(ts.DatabaseOidsWithLocalCatalog, databaseOid)
-
-		schemaConnection.Close()
```
uhoh-itsmaciek, line 51: Extracted most of the loop to a function to make it easier to ensure the connection is closed in all paths by using `defer`.

lfittl (approved):
:+1:

Issue #124, opened 2020-12-15
Can't use the collector when there is a hash (#) in the password


shamil:
I guess it treat everything after hash as comment.

uhoh-itsmaciek:
Hi @shamil--thanks for the report. Have you tried escaping the hash in the password value with a backslash, or quoting the entire value?

caleb15:
I ran into this too - see https://github.com/pganalyze/collector/issues/132

uhoh-itsmaciek:
@caleb15 oh wow, it definitely shouldn't segfault. I only suggested quoting or escaping because a hash is a `.ini` comment character, so I assumed the hash and everything after were being ignored, leading to a different password. Are you setting the password via `db_password` or `db_url`?

uhoh-itsmaciek:
So I tried this myself with db_url and I was able to reproduce the crash. The hash character is treated as a comment by https://github.com/go-ini/ini, which we use for parsing the config file, and the resulting invalid URL fails to parse, and we're not doing proper error-checking regarding valid URLs. Unfortunately, it looks like the quoting and escaping suggestions I made earlier do *not* solve this--looks like go-ini does not handle this correctly: https://github.com/go-ini/ini/issues/277 https://github.com/go-ini/ini/issues/187 .

uhoh-itsmaciek:
We have a fix in progress for the error handling in #133 to avoid the crash and allow `#` in values, but the root cause here is that the character is reserved in URLs (since it delineates the start of the fragment), so it must be URL-encoded as `%23`. Could you please try that @caleb15 @shamil? That should work even in current collector versions.

uhoh-itsmaciek:
Closing since #133 should fix this; please reopen if there are still issues.

caleb15:
I'm using db_url. I fixed it by removing the hash.

PR #133, opened 2020-12-23
Validate db_url/DB_URL, if set, when initially reading config

uhoh-itsmaciek:
Right now, we do not validate it, so if set to an invalid URL, it can
crash some of our helper functions with a segfault due to a nil
pointer reference.

Validate at startup, but allow to continue to avoid retrofitting error
handling across helpers that right now cannot return errors. The helpers
return dummy/default values which will fail authentication.

See #124 (this is not a fix for that, since go-ini does not parse
comments correctly, but at least it will give users a proper error
message instead of a segfault.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -168,7 +168,10 @@ func (config ServerConfig) GetPqOpenString(dbNameOverride string) string {
 	var dbPort int
 
 	if config.DbURL != "" {
-		u, _ := url.Parse(config.DbURL)
+		u, err := url.Parse(config.DbURL)
+		if err != nil {
+			return ""
+		}
 
 		if u.User != nil {
 			dbUsername = u.User.Username()
@@ -297,7 +300,10 @@ func (config ServerConfig) GetPqOpenString(dbNameOverride string) string {
 // GetDbHost - Gets the database hostname from the given configuration
 func (config ServerConfig) GetDbHost() string {
 	if config.DbURL != "" {
-		u, _ := url.Parse(config.DbURL)
+		u, err := url.Parse(config.DbURL)
+		if err != nil {
+			return ""
+		}
 		parts := strings.Split(u.Host, ":")
 		return parts[0]
 	}
@@ -322,7 +328,10 @@ func (config ServerConfig) GetDbURLRedacted() string {
 // GetDbPort - Gets the database port from the given configuration
 func (config ServerConfig) GetDbPort() int {
 	if config.DbURL != "" {
-		u, _ := url.Parse(config.DbURL)
+		u, err := url.Parse(config.DbURL)
+		if err != nil {
+			return 5432
+		}
 		parts := strings.Split(u.Host, ":")
 
 		if len(parts) == 2 {
@@ -339,7 +348,10 @@ func (config ServerConfig) GetDbPort() int {
 // GetDbUsername - Gets the database hostname from the given configuration
 func (config ServerConfig) GetDbUsername() string {
 	if config.DbURL != "" {
-		u, _ := url.Parse(config.DbURL)
+		u, err := url.Parse(config.DbURL)
+		if err != nil {
+			return ""
+		}
 		if u != nil && u.User != nil {
 			return u.User.Username()
 		}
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -322,7 +328,10 @@ func (config ServerConfig) GetDbURLRedacted() string {
 // GetDbPort - Gets the database port from the given configuration
 func (config ServerConfig) GetDbPort() int {
 	if config.DbURL != "" {
-		u, _ := url.Parse(config.DbURL)
+		u, err := url.Parse(config.DbURL)
+		if err != nil {
+			return 5432
```
lfittl, line 333: I'm thinking we should also return the zero value here (i.e. `0`), since a URL was passed that can't be handled.

lfittl (approved):
Good catch, we definitely shouldn't be ignoring these parsing problems.

I wonder if we should output a warning in this case, so that it's clear whats happening? (downside is that requires passing the logger around, but seems like that might be worth the effort - or we return an error and have the caller handle it)

config/config.go
uhoh-itsmaciek: I was afraid that might be confusing, but I think you're right, that makes more sense.

uhoh-itsmaciek:
@lfittl well, I added https://github.com/pganalyze/collector/pull/133/files#diff-b6ac6ad02a4e2347983afa3479b0bbe6673db968102194c2d52d396a6d1e8b12R424 in `Read`--that should warn when the config is read in, so any usage errors should be obvious. I don't think logging directly in the helpers has much advantage, does it?

config/read.go
```diff
@@ -420,6 +420,14 @@ func Read(logger *util.Logger, filename string) (Config, error) {
 					conf.Servers = append(conf.Servers, *config)
 				}
 			}
+
+			if config.DbURL != "" {
+				_, err := url.Parse(config.DbURL)
+				if err != nil {
+					prefixedLogger := logger.WithPrefix(config.SectionName)
+					prefixedLogger.PrintError("Could not parse db_url")
```
lfittl, line 428: Maybe clarify as "Could not parse db_url, check URL format and that special characters are correctly escaped"

lfittl:
> @lfittl well, I added https://github.com/pganalyze/collector/pull/133/files#diff-b6ac6ad02a4e2347983afa3479b0bbe6673db968102194c2d52d396a6d1e8b12R424 in `Read`--that should warn when the config is read in, so any usage errors should be obvious. I don't think logging directly in the helpers has much advantage, does it?

Ah, yeah, that works :)

config/read.go
uhoh-itsmaciek: Well, the thing is, quoting and escaping values in go-ini are both pretty broken (see https://github.com/pganalyze/collector/issues/124#issuecomment-749835070 ), so I don't think that's going to be very helpful.

However, I went looking to see if I could fix this upstream, and while that's probably not worth it (there are a lot of different flags and behaviors that might break), I did find an option that works around the issue for the most part: with `SpaceBeforeInlineComment` set to true, a `#` only begins a single-line comment if it's preceded by a space. While theoretically this could break some configurations, I think it's fairly unlikely anyone was relying on this behavior (starting a line comment *without* offsetting it from the value with a space) in practice.

In the course of this, I realized that the real issue here is that the problem character needs to be percent-encoded, so most of this is moot. But I think we can still improve the message here along the lines of your suggestion.

I also was wondering if we should tell you what the error is, but unfortunately, Go can sometimes include part of the input string in the error message:
```
maciek@starling:~/duboce-labs/collector$ grep db_url t.conf
db_url             = postgres://foo:bar#hello@localhost/dbname
maciek@starling:~/duboce-labs/collector$ ./pganalyze-collector --config=t.conf --dry-run
2020/12/28 11:48:27 E [server1] Could not parse db_url: parse "postgres://foo:bar": invalid port ":bar" after host; note that any special characters must be percent-encoded
2020/12/28 11:48:27 E Config Error: Configuration file is empty, please edit t.conf and reload the collector
```

Since this output could easily end up in your logs, that's probably not a good idea. (Also, the message here suggests the URL is still being cut off, but that's not the case: perhaps Go is only printing the substring because it considers the rest of the URL a part of the fragment, which is not relevant to this particular parsing problem.)

Since the PR is approved, I'll merge with an improved error message (without the `url.Parse` error), but let me know if you have any further thoughts.

lfittl: Makes sense re: why we can't word this like suggested. Also agreed that we should be careful with the error message containing any sensitive data, so passing through the original error message is probably not a good idea. I think this is good to go 👍

PR #134, opened 2020-12-29
Change <unidentified queryid> marker to clarify cause

uhoh-itsmaciek:
Right now, if pg_stat_statements.max is too low for a given query
workload, we may end up capturing stats for queries whose text is no
longer available to us. We flag this situation by using the string
`<unidentified queryid>` as the query text, but this is fairly opaque.
We have a change in progress to improve the display of this in-app,
but changing the actual marker string used here can help clarify
things as well.

Instead of `<unidentified queryid>`, this patch changes the text for
queries like this to `<query text unavailable>`, which describes the
situation more accurately.

### DIFF
--- a/output/transform/postgres_query.go
+++ b/output/transform/postgres_query.go
@@ -39,8 +39,8 @@ func upsertQueryReferenceAndInformation(s *snapshot.FullSnapshot, statementTexts
 
 	// Information
 	normalizedQuery := ""
-	if value.statement.Unidentified {
-		normalizedQuery = "<unidentified queryid>"
+	if value.statement.QueryTextUnavailable {
+		normalizedQuery = "<query text unavailable>"
 	} else if value.statement.InsufficientPrivilege {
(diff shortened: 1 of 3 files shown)


uhoh-itsmaciek:
Should we also change [the string we use for fingerprinting](https://github.com/pganalyze/collector/blob/master/output/transform/postgres_statements.go#L18)? Should we change the struct field name [here](https://github.com/pganalyze/collector/blob/master/state/postgres_statement.go#L13)?

lfittl:
> Should we also change [the string we use for fingerprinting](https://github.com/pganalyze/collector/blob/master/output/transform/postgres_statements.go#L18)? Should we change the struct field name [here](https://github.com/pganalyze/collector/blob/master/state/postgres_statement.go#L13)?

Makes sense to be consistent and use `<query text unavailable>` for the fingerprinting, and `QueryTextUnavailable` for the field name.

lfittl (approved):

PR #138, opened 2021-01-07
Ignore duplicate dbnames when collecting

uhoh-itsmaciek:
If a database is specified multiple times via dbname, we collect its
schema information multiple times, but fail to correctly assemble a
snapshot due to identified collisions. The inconsistent snapshots are
sent off correctly, but can cause problems when processing.

Instead, only collect each database once to avoid the issue.

We could potentially warn when this happens, but the intent is clear
so being overparticular here seems of little benefit.


lfittl (approved):
:+1:

PR #145, opened 2021-01-31
Add support for remote syslog

lfittl:
```
Add support for receiving logs from remote servers over syslog

You can now specify the new "db_log_syslog_server" config setting, or
"LOG_SYSLOG_SERVER" environment variable in order to setup the collector
as a syslog server that can receive logs from a remote server via syslog
to the server that runs the collector.

Note that the format of this setting is "listen_address:port", and its
recommended to use a high port number to avoid running the collector as root.

For example, you can specify "0.0.0.0:32514" and then send syslog messages
to the collector's server address at port 32514.

Note that you need to use protocol RFC5424, with an unencrypted TCP
connection. Due to syslog not being an authenticated protocol it is
recommended to only use this integration over private networks.
```

```
Add support for "pid=%p,user=%u,db=%d,app=%a,client=%h " log_line_prefix

This prefix misses a timestamp, but is useful when sending data over syslog.
```

```
Remote syslog: Add support for inline log line number and chunk
```

```
Add support for "user=%u,db=%d,app=%a,client=%h " log_line_prefix
```

### DIFF
--- a/input/system/selfhosted/log_receiver.go
+++ b/input/system/selfhosted/log_receiver.go
@@ -26,6 +26,16 @@ import (
 	uuid "github.com/satori/go.uuid"
 )
 
+type SelfHostedLogStreamItem struct {
+	Line string
+
+	// Optional, only used for syslog messages
+	OccurredAt time.Time
+	BackendPid int32
+	LogLineNumber int32
+	LogLineNumberChunk int32
+}
+
 const settingValueSQL string = `
 SELECT setting
 	FROM pg_settings
@@ -145,11 +155,17 @@ func SetupLogTails(ctx context.Context, servers []*state.Server, globalCollectio
 			if err != nil {
 				prefixedLogger.PrintError("ERROR - %s", err)
 			}
+		} else if server.Config.LogSyslogServer != "" {
+			logStream := logReceiver(ctx, server, globalCollectionOpts, prefixedLogger, nil)
+			err := setupSyslogHandler(ctx, server.Config.LogSyslogServer, logStream, prefixedLogger)
+			if err != nil {
+				prefixedLogger.PrintError("ERROR - %s", err)
+			}
 		}
 	}
 }
 
-func tailFile(ctx context.Context, path string, out chan<- string, prefixedLogger *util.Logger) error {
+func tailFile(ctx context.Context, path string, out chan<- SelfHostedLogStreamItem, prefixedLogger *util.Logger) error {
 	prefixedLogger.PrintVerbose("Tailing log file %s", path)
 
 	t, err := follower.New(path, follower.Config{
@@ -167,7 +183,7 @@ func tailFile(ctx context.Context, path string, out chan<- string, prefixedLogge
 		for {
 			select {
 			case line := <-t.Lines():
-				out <- line.String()
+				out <- SelfHostedLogStreamItem{Line: line.String()}
 			case <-ctx.Done():
 				prefixedLogger.PrintVerbose("Stopping log tail for %s (stop requested)", path)
 				break TailLoop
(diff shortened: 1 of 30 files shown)


input/system/selfhosted/log_receiver.go
```diff
@@ -386,10 +402,23 @@ func logReceiver(ctx context.Context, server *state.Server, globalCollectionOpts
 				// Note that we need to restore the original trailing newlines since
 				// ProcessLogStream below expects them and they are not present in the tail
 				// log stream.
```
uhoh-itsmaciek, line 404: Is this still true for syslog?

input/system/selfhosted/syslog_log_receiver.go
```diff
@@ -0,0 +1,64 @@
...
+
+				item.OccurredAt, _ = logParts["timestamp"].(time.Time)
+
+				pidStr, _ := logParts["proc_id"].(string)
+				if s, err := strconv.ParseInt(pidStr, 10, 32); err == nil {
+					item.BackendPid = int32(s)
+				}
+
+				logLineNumberStr, _ := logParts["structured_data"].(string)
+				logLineNumberParts := regexp.MustCompile(`^\[(\d+)-(\d+)\]$`).FindStringSubmatch(logLineNumberStr)
```
uhoh-itsmaciek, line 42: Should we lift this regexp initialization? In terms of performance it probably doesn't matter (and/or Go is smart enough to optimize that itself), but it would be a little more readable.

uhoh-itsmaciek (approved):
Nice. A couple of minor questions, but looks great.

input/system/selfhosted/log_receiver.go
lfittl: Yes, the way Postgres sends it it does not contain newlines (as Postgres splits at the newline boundary, or 900 characters, whichever is earlier):

https://github.com/postgres/postgres/blob/REL_13_STABLE/src/backend/utils/error/elog.c#L1948

input/system/selfhosted/syslog_log_receiver.go
lfittl: Good idea - I'll pull this out for better readability.

logs/parse.go
```diff
@@ -72,6 +73,7 @@ var LogPrefixCustom8Regexp = regexp.MustCompile(`(?s)^\[` + PidRegexp + `\]: \[`
 var LogPrefixCustom9Regexp = regexp.MustCompile(`(?s)^` + TimeRegexp + ` ` + HostAndPortRegexp + ` ` + UserRegexp + ` ` + AppRegexp + ` \[` + SessionIdRegexp + `\] \[` + PidRegexp + `\] ` + LevelAndContentRegexp)
 var LogPrefixCustom10Regexp = regexp.MustCompile(`(?s)^` + TimeRegexp + ` \[` + PidRegexp + `\]: \[` + LogLineCounterRegexp + `-1\] db=` + DbRegexp + `,user=` + UserRegexp + ` ` + LevelAndContentRegexp)
 var LogPrefixCustom11Regexp = regexp.MustCompile(`(?s)^pid=` + PidRegexp + `,user=` + UserRegexp + `,db=` + DbRegexp + `,app=` + AppInsideBracketsRegexp + `,client=` + HostRegexp + ` ` + LevelAndContentRegexp)
+var LogPrefixCustom12Regexp = regexp.MustCompile(`(?s)^user=` + UserRegexp + `,db=` + DbRegexp + `,app=` + AppInsideBracketsRegexp + `,client=` + HostRegexp + ` ` + LevelAndContentRegexp)
```
uhoh-itsmaciek, line 76: I noticed I missed this in `Custom11` before, but why `AppInsideBracketsRegexp` instead of `AppRegexp`? I see that `Custom6` is also using `AppInsideBracketsRegexp` ostensibly outside of brackets, but `Custom9` is using just `AppRegexp`. Are these misnamed or am I missing something?

uhoh-itsmaciek: Oh I see, based on the test, it looks like the value itself has brackets around it? But are the brackets not just part of the application name in that case? And what if they are not set by the application. `%a` itself does *not* add brackets, right?

uhoh-itsmaciek (approved):

logs/parse.go
lfittl: Nice, that is a good catch, I certainly got this wrong :)

Adjusted to use the correct regexp for the new prefixes, and also adjusted a few others that were incorrect in this regard.

PR #147, opened 2021-02-16
Tweak collector sample config file

uhoh-itsmaciek:
* put db_host first since this is arguably most important
 * drop port because this is almost always 5432 and an example is
   still available in the second, fuller, commented-out config


lfittl (approved):
:+1:

PR #149, opened 2021-02-17
Update log location discovery

uhoh-itsmaciek:
The helper fails on a default AL2 install since the binary is called
postmaster not postgres, and there's no apparent reason not to use the
GUCS directly. This lets us drop the helper subcommand entirely.

Also, do not show a db_log_location if the log directory is inside the
data directory: this is almost certain to cause permissions problems,
and instead we can direct users to address this in our setup guide.

### DIFF
--- a/input/system/selfhosted/log_receiver.go
+++ b/input/system/selfhosted/log_receiver.go
@@ -30,9 +30,9 @@ type SelfHostedLogStreamItem struct {
 	Line string
 
 	// Optional, only used for syslog messages
-	OccurredAt time.Time
-	BackendPid int32
-	LogLineNumber int32
+	OccurredAt         time.Time
+	BackendPid         int32
+	LogLineNumber      int32
 	LogLineNumberChunk int32
 }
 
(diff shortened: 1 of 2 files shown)


input/system/selfhosted/log_receiver.go
```diff
@@ -103,20 +103,17 @@ func DiscoverLogLocation(servers []*state.Server, globalCollectionOpts state.Col
...
-				continue
-			}
-			if !strings.HasPrefix(logDirectory, "/") {
-				logDirectory = status.DataDirectory + "/" + logDirectory
+
+			if strings.HasPrefix(logDirectory, "/") {
+				prefixedLogger.PrintInfo("Found log location, add this to your pganalyze-collector.conf in the [%s] section:\ndb_log_location = %s", server.Config.SectionName, logDirectory)
+			} else {
+				prefixedLogger.PrintInfo("WARNING: found log location in data directory - please check our setup guide for instructions.\n")
 			}
```
uhoh-itsmaciek, line 116: @lfittl do you think hiding the result in this case is reasonable? Or should we still show it, but indicate you may need to make changes? My understanding is that this is only likely to work with pretty questionable permissions setups to begin with, but maybe it's worth keeping anyway?

input/system/selfhosted/log_receiver.go
```diff
@@ -76,7 +76,7 @@ func DiscoverLogLocation(servers []*state.Server, globalCollectionOpts state.Col
 		}
 
 		if logDestination == "syslog" {
-			prefixedLogger.PrintInfo("Log location detected as syslog - please check our setup guide for rsyslogd or syslog-ng instructions")
+			prefixedLogger.PrintInfo("WARNING: Logging via syslog - please check our setup guide for rsyslogd or syslog-ng instructions")
```
uhoh-itsmaciek, line 79: I think WARNINGs here make it clear you need to change something, but don't imply the setup is broken.

input/system/selfhosted/log_receiver.go
lfittl: Maybe we can include it in the warning message, like this:

```
WARNING: Found relative log location "log" inside data directory - please check our setup guide for instructions\n
```

(I suspect this will be the default, i.e. `log`, 95% of the time)

lfittl (approved):

PR #150, opened 2021-02-17
Improve --test-logs behavior

uhoh-itsmaciek:
Add the privilege drop test when running as root.

@lfittl is this the right approach here? The test path is a little hard to follow.

### DIFF
--- a/main.go
+++ b/main.go
@@ -116,7 +116,7 @@ func run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
 			}
 			return
 		} else if globalCollectionOpts.TestRunLogs {
-			runner.TestLogsForAllServers(servers, globalCollectionOpts, logger)
+			reloadOkay = doLogTest(servers, globalCollectionOpts, logger)
 			return
 		} else {
 			var allFullSuccessful bool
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -116,7 +116,7 @@ func run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
 			}
 			return
 		} else if globalCollectionOpts.TestRunLogs {
-			runner.TestLogsForAllServers(servers, globalCollectionOpts, logger)
+			reloadOkay = doLogTest(servers, globalCollectionOpts, logger)
 			return
 		} else {
 			var allFullSuccessful bool
```
uhoh-itsmaciek, line 122: GitHub won't let me comment on the lines below, but should we consider the `doLogTest` return value for the regular test? Right now we have `reloadOkay = allFullSuccessful && allActivitySuccessful` on line 134--should we add the log test result to the conjunction?

main.go
```diff
@@ -559,56 +559,61 @@ func Reload(logger *util.Logger) {
...
-		if err != nil {
-			logger.PrintError("Could not run collector log test as \"pganalyze\" user: %s", err)
-			return
-		}
+	pgaUser, err := user.Lookup("pganalyze")
+	if err != nil {
+		logger.PrintVerbose("Could not locate pganalyze user, skipping privilege drop test: %s", err)
+		return false
+	} else if curUser.Uid == pgaUser.Uid {
+		logger.PrintVerbose("Current user is already pganalyze user, skipping privilege drop test")
```
uhoh-itsmaciek, line 589: Should this be `PrintError`? Or should we return `true` in these cases? Right now the output can be confusing:

```
$ sudo ./pganalyze-collector --test-logs --reload --config=./pganalyze-collector-local.conf
2021/02/17 11:50:27 I Running collector test with pganalyze-collector 0.36.0
2021/02/17 11:50:27 I [server1] Testing log collection (local)...
2021/02/17 11:50:33 I [server1]   Local log test successful
2021/02/17 11:50:33 E Error: Reload requested, but ignoring since configuration errors are present
$ sudo ./pganalyze-collector --test-logs --reload --config=./pganalyze-collector-local.conf --verbose
2021/02/17 11:50:51 V [server1] Identified as api_system_type: self_hosted, api_system_scope: 5432/mydb1*, api_system_id: starling
2021/02/17 11:50:51 V Did not open state file: open /var/lib/pganalyze-collector/state: no such file or directory
2021/02/17 11:50:51 I Running collector test with pganalyze-collector 0.36.0
2021/02/17 11:50:51 I [server1] Testing log collection (local)...
2021/02/17 11:50:51 V [server1] Searching for log file(s) in /var/log/postgresql/postgresql-11-main.log
2021/02/17 11:50:51 V [server1] Tailing log file /var/log/postgresql/postgresql-11-main.log
2021/02/17 11:50:57 I [server1]   Local log test successful
2021/02/17 11:50:57 V [server1] Log file fsnotify watcher received stop signal
2021/02/17 11:50:57 V [server1] Stopping log tail for /var/log/postgresql/postgresql-11-main.log (stop requested)
2021/02/17 11:50:57 V Could not locate pganalyze user, skipping privilege drop test: user: unknown user pganalyze
2021/02/17 11:50:57 E Error: Reload requested, but ignoring since configuration errors are present
```
(The first test fails because I don't have a pganalyze user on this machine, but the test output does not make that clear.)

main.go
lfittl: We intentionally don't fail for the regular test command, since you may not have Log Insights enabled on your plan (which will make the log test fail when its getting the log grant). In those situations we still want `--test` to be successful (i.e. issue a reload), but `--test-logs` should fail (and not reload).

Thus, the code as written here is correct I think. We could add a comment explaining this for future reference in the `--test` command call of `doLogTest` (where we ignore the return value).

main.go
lfittl: You commented on the line that has `logger.PrintVerbose("Current user is already pganalyze user, skipping privilege drop test")` - I assume you actually meant the line that has `logger.PrintVerbose("Could not locate pganalyze user, skipping privilege drop test: %s", err)`?

I think we could consider making that a warning, and doing `return true` in that case. Setups that don't have a pganalyze user are rare (they'd be self compiled), but it might be nice to support running the collector as root in such situations (and having a successful `--test-logs` command), if someone really wants to.

I think the if branch this comment is on `( if curUser.Uid == pgaUser.Uid {`) can be left as-is.

lfittl (approved):
:+1:

PR #152, opened 2021-02-18
Check whether pg_stat_statements exists in a different schema

uhoh-itsmaciek:
Right now, we require the pg_stat_statements extension to be installed
in the public schema, and try to create it if it does not exist.
However, we only look for it in the public schema, and run the create
with IF NOT EXISTS, which will be a no-op if it does exist in a
different schema. At this point, assuming we created it successfully,
we'll try to query pg_stat_statements again, and get a confusing
"pg_stat_statements does not exist" message.

Instead, check to see if the extension exists in another schema first,
and give a more appropriate error message.

### DIFF
--- a/input/postgres/statements.go
+++ b/input/postgres/statements.go
@@ -124,20 +124,29 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


lfittl (approved):
:+1:

PR #153, opened 2021-02-19
Update Go to 1.16

lfittl:
This introduces a change to Go's certificate handling, which may break
certain older versions of Amazon RDS certificates, as they do not
include a SAN. When this is the case you will see an error message like
"x509: certificate relies on legacy Common Name field".

As a temporary workaround you can run the collector with the
GODEBUG=x509ignoreCN=0 environment setting, which ignores these incorrect
fields in these certificates. For a permanent fix, you need to update
your RDS certificates to include the correct SAN field:

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL-certificate-rotation.html


uhoh-itsmaciek:
Test failure seems to be 9.2 specific, so trying https://github.com/pganalyze/collector/pull/154 -- not sure if that's the right approach, but we can consider it.

uhoh-itsmaciek:
Done with the additional changes in #154, so closing this one.

PR #157, opened 2021-02-25
Don't use pganalyze query marker for "--test-explain" command

lfittl:
The marker means the resulting query gets hidden from the EXPLAIN plan
list, which is what we don't want for this test query - its intentional
that we can see the EXPLAIN plan we're generating for the test.


uhoh-itsmaciek (approved):
Nice catch--I think this could explain some of the explain setup problems I saw with guided setup.

PR #158, opened 2021-03-03
Add curl-able install script to streamline installation across systems

uhoh-itsmaciek:
Add a one-step install script for the collector. This lets us simplify package installation instructions to something like

```
curl https://packages.pganalyze.com/collector-install.sh | bash
```

This uses sudo (prompting if necessary) if not already running as root and will install the correct package based on system type. If an unrecognized version of a supported distribution is detected, we prompt whether the user would like to install the latest available package for that distribution instead.

The script supports several environment variables:

 - `PGA_INSTALL_NONINTERACTIVE`: if non-empty, do not prompt for user input and assume `y` to all prompts
 - `PGA_API_BASEURL`: if non-empty, api_base_url to add to collector config as part of install
 - `PGA_API_KEY`: if non-empty, pganalyze API key to write to collector config file as part of install
 - `PGA_GUIDED_SETUP`: if non-empty, run guided setup immediately after install (if `PGA_INSTALL_NONINTERACTIVE` is also set, guided setup is non-interactive and uses recommended settings)
   - `DB_NAME`: if the above `PGA_GUIDED_SETUP` is set, `db_name` to write to collector config file as part of install (ignored otherwise); if this is not set, `db_name` defaults to 'postgres'

These can be passed to the script like this:

```
curl https://packages.pganalyze.com/collector-install.sh | env VAR1=foo VAR2=bar bash
```

TODO:
 - [x] add shellcheck linting to CI
 - [x] actually run through script in CI?
 - [x] ~prompt for confirmation if passwordless sudo is enabled~
   - on second thought, this is not necessary: we run `sudo -k` to clear previous authorization, but an additional prompt is not necessary
 - [x] link to install instructions and/or run guided setup after installation
 - [x] determine deployment process
   - we'll put this in packages.pganalyze.com
 - [x] test Ubuntu install works
 - [x] fix Debian install [1]
 - [x] fix AL2 install [2]
 - [x] fix Fedora install [3]
 - [x] fix RHEL install [4]
 - [x] manually retest install on supported distributions
   - [x] AL2
   - [x] RHEL8
   - [x] Fedora 33 (falls back to Fedora 30 package)
   - [x] Ubuntu Focal (20.04)
   - [x] Debian Buster (10) (requires gnupg install to be confirmed)
 - [x] ~copy and validate package fingerprints?~
   - on second thought, this is tricky to do automatically, and it's probably better to have fingerprint verification happen in a system more operationally isolated from the install packages (i.e., have the user do it based on fingerprints listed in the documentation)

[1]: currently this fails on `apt-key add`:
```
$ curl https://raw.githubusercontent.com/pganalyze/collector/f5cede1af633f8b35a843b2644564ecd5d24a704/contrib/install.sh | bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  3318  100  3318    0     0   140k      0 --:--:-- --:--:-- --:--:--  140k
This script requires superuser access to install packages
You will be prompted for your password by sudo
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1625  100  1625    0     0  16752      0 --:--:-- --:--:-- --:--:-- 16752
E: gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation
```

[2]: current failure (the confirmation prompt is ignored so script fails):
```
Retrieving key from https://packages.pganalyze.com/pganalyze_signing_key.asc
Importing GPG key 0xA2B5F2F9:
 Userid     : "pganalyze <team@pganalyze.com>"
 Fingerprint: c09b 2cab 0db3 78f6 e7fd 93f1 0e6d ec71 a2b5 f2f9
 From       : https://packages.pganalyze.com/pganalyze_signing_key.asc
pganalyze_collector/signature                                                                                                                                                       | 2.9 kB  00:00:00 !!! 
https://packages.pganalyze.com/el/7/repodata/repomd.xml: [Errno -1] Gpg Keys not imported, cannot verify repomd.xml for repo pganalyze_collector
Trying other mirror.


 One of the configured repositories failed (pganalyze_collector),
 and yum doesn't have enough cached data to continue. At this point the only
 safe thing yum can do is fail. There are a few ways to work "fix" this:

     1. Contact the upstream for the repository and get them to fix the problem.

     2. Reconfigure the baseurl/etc. for the repository, to point to a working
        upstream. This is most often useful if you are using a newer
        distribution release than is supported by the repository (and the
        packages for the previous distribution release still work).

     3. Run the command with the repository temporarily disabled
            yum --disablerepo=pganalyze_collector ...

     4. Disable the repository permanently, so yum won't use it by default. Yum
        will then just ignore the repository until you permanently enable it
        again or use --enablerepo for temporary usage:

            yum-config-manager --disable pganalyze_collector
        or
            subscription-manager repos --disable=pganalyze_collector

     5. Configure the failing repository to be skipped, if it is unavailable.
        Note that yum will try to contact the repo. when it runs most commands,
        so will have to try and fail each time (and thus. yum will be be much
        slower). If it is a very temporary problem though, this is often a nice
        compromise:

            yum-config-manager --save --setopt=pganalyze_collector.skip_if_unavailable=true

failure: repodata/repomd.xml from pganalyze_collector: [Errno 256] No more mirrors to try.
https://packages.pganalyze.com/el/7/repodata/repomd.xml: [Errno -1] Gpg Keys not imported, cannot verify repomd.xml for repo pganalyze_collector
```

[3]: current failure (just need to configure package?)
```
$ curl https://raw.githubusercontent.com/pganalyze/collector/f5cede1af633f8b35a843b2644564ecd5d24a704/contrib/install.sh | bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  3318  100  3318    0     0  36461      0 --:--:-- --:--:-- --:--:-- 36866
Collector install failed: unrecognized Fedora version: 33

Please contact support@pganalyze.com for help and include information about your platform
```

[4]: current failure (similar to AL2 failure)
```
Importing GPG key 0xA2B5F2F9:
 Userid     : "pganalyze <team@pganalyze.com>"
 Fingerprint: C09B 2CAB 0DB3 78F6 E7FD 93F1 0E6D EC71 A2B5 F2F9
 From       : https://packages.pganalyze.com/pganalyze_signing_key.asc
Is this ok [y/N]: pganalyze_collector                                                                    [   ===                                                                           ] ---  B/s |   0pganalyze_collector                                                                                                                                                        3.3 kB/s | 836  B     00:00    
Error: Failed to download metadata for repo 'pganalyze_collector': repomd.xml GPG signature verification error: Bad GPG signature
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,3 +34,4 @@ jobs:
         make build OUTFILE=pganalyze-collector-linux-amd64
         make test
         make integration_test
+        shellcheck contrib/install.sh
--- a/contrib/install.sh
+++ b/contrib/install.sh
@@ -177,6 +177,11 @@ else
   fail "unrecognized package kind: $pkg"
 fi
 
+if [ -n "$PGA_API_BASE_URL" ];
+then
+  $maybe_sudo sed -i "/^\[pganalyze\]$/a api_base_url = ${PGA_API_BASE_URL}" /etc/pganalyze-collector.conf
+fi
+
 if [ -n "$PGA_API_KEY" ];
 then
   $maybe_sudo sed -i "s/^#api_key = your_api_key$/api_key = ${PGA_API_KEY}/" /etc/pganalyze-collector.conf
--- a/contrib/install.sh
+++ b/contrib/install.sh
@@ -88,8 +88,9 @@ if [ "$(id -u)" != "0" ]; then
     fail "not running as root and could not find sudo command"
   fi
 
-  echo "You will be prompted for your password by sudo"
-  # clear any previous sudo permission to avoid inadvertent 
+  echo "You may be prompted for your password by sudo"
+
+  # clear any previous sudo permission to avoid inadvertent confirmation
   $maybe_sudo -k
 fi
 
--- a/contrib/install.sh
+++ b/contrib/install.sh
@@ -1,0 +1,123 @@
+#!/bin/bash
+
+set -e
+
+fail () {
+  >&2 echo "Collector install failed: $1"
+  >&2 echo
+  >&2 echo "Please contact support@pganalyze.com for help and include information about your platform"
+  exit 1
+}
+
+distribution=''
+version=''
+pkg=''
+
+if ! test -r /etc/os-release;
+then
+  fail "cannot read /etc/os-release to determine distribution"
+fi
+
+arch=$(uname -m)
+if [ "$arch" != 'x86_64' ];
+then
+  fail "unsupported architecture: $arch"
+fi
+
+if grep -q '^ID="amzn"$' /etc/os-release && grep -q '^VERSION_ID="2"$' /etc/os-release;
+then
+  # Amazon Linux 2, based on RHEL7
+  pkg=yum
+  distribution=el
+  version=7
+elif grep -q '^ID="rhel"$' /etc/os-release;
+then
+  # RHEL
+  pkg=yum
+  distribution=el
+  version=$(grep VERSION_ID /etc/os-release | cut -d= -f2 | tr -d '"' | cut -d. -f1)
+  if [ "$version" != 7 ] && [ "$version" != 8 ];
+  then
+    fail "unrecognized RHEL version: ${version}"
+  fi
+elif grep -q '^ID=fedora$' /etc/os-release;
+then
+  # Fedora
+  pkg=yum
+  distribution=fedora
+  version=$(grep VERSION_ID /etc/os-release | cut -d= -f2)
+
+  if [ "$version" != 30 ] && [ "$version" != 29 ];
+  then
+    fail "unrecognized Fedora version: ${version}"
+  fi
+elif grep -q '^ID=ubuntu$' /etc/os-release;
+then
+  # Ubuntu
+  pkg=deb
+  distribution=ubuntu
+  version=$(grep VERSION_CODENAME /etc/os-release | cut -d= -f2)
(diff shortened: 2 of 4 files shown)


contrib/install.sh
```diff
@@ -0,0 +1,199 @@
...
+  >&2 echo "Install failed: $1"
+  >&2 echo
+  >&2 echo "Please contact support@pganalyze.com for help and include information about your platform"
+  exit 1
+}
+
+user_input=''
+yum_opts=''
+apt_opts=''
+pgags_opts=''
```
lfittl, line 16: Do we need to add explicit support for `PGA_API_BASEURL` somewhere here? (it looks like `PGA_API_KEY` makes it through, but I couldn't set the base URL programatically, e.g. to test on staging)

lfittl (approved):
💯

contrib/install.sh
uhoh-itsmaciek: Sure, I can add that.

.github/workflows/ci.yml
```diff
@@ -34,3 +34,4 @@ jobs:
         make build OUTFILE=pganalyze-collector-linux-amd64
         make test
         make integration_test
+        shellcheck contrib/install.sh
```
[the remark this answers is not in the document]
uhoh-itsmaciek: Yeah, it's neat that shellcheck is available out-of-the-box in GitHub Actions.

PR #162, opened 2021-03-16
Docker builds: Increase stack size to 2MB to prevent rare crashes

lfittl:
Alpine has a very small stack size by default (80kb) which is less than
the default that Postgres expects (100kb). Since there is no good reason
to reduce it to such a small amount, increase to usually common Linux
default of 2MB stack size.

This would have surfaced as a hard crash of the Docker container with
error code 137 or 139, easily confused with out of memory errors, but
clearly distinct from it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ WORKDIR $CODE_DIR
 RUN apk add --no-cache --virtual .build-deps make curl libc-dev gcc go git tar \
   && apk add --no-cache ca-certificates \
   && curl -o /usr/local/bin/gosu -sSL "https://github.com/tianon/gosu/releases/download/1.6/gosu-amd64" \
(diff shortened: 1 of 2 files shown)


uhoh-itsmaciek (approved):
Nice. I assume the stack size is small to minimize overhead, and since 2MB is a small fraction of typical collector memory usage anyway, and we normally only have a single process, this does not really impact things much?

lfittl:
> Nice. I assume the stack size is small to minimize overhead, and since 2MB is a small fraction of typical collector memory usage anyway, and we normally only have a single process, this does not really impact things much?

Thats right - since we have a single process this is not really noticeable, and provides some extra headroom for other C level functions that recurse. Alpine's default here (of 80kb) seems rather aggressive, but might make sense for some embedded systems where every megabyte counts.

PR #165, opened 2021-03-30
Update to pg_query 2.0 and Postgres 13 parser

lfittl:
This is a major upgrade in terms of supported syntax (Postgres 10 to 13),
as well as a major change in the fingerprints, which are now shorter and
not compatible with the old format.

When you upgrade to this version of the collector you will see a break
in statistics, that is, you will see new query entries in pganalyze after
adopting this version of the collector.


uhoh-itsmaciek (approved):
Awesome!

PR #166, opened 2021-03-31
RDS logs: Support long log events, handle errors better

lfittl:
This changes the RDS log download logic to only analyze log lines after
they have been downloaded and stitched. To prevent running out of
memory when the collector first starts (where the marker is not set yet)
we now explicitly restrict to the last 10 megabytes in the file.

In passing improve the error handling to return errors to the caller
instead of calling the logger directly.

With this change very long EXPLAIN plans can be successfully parsed,
such as for this query, which causes a plan to exceed 10k lines with
full auto_explain configuration:

```
SELECT 1 FROM pg_sleep(5) UNION SELECT 1 --- repeat "UNION SELECT 1" 414 more times
```

### DIFF
--- a/input/logs.go
+++ b/input/logs.go
@@ -14,7 +14,10 @@ func DownloadLogs(server *state.Server, prevLogState state.PersistedLogState, co
 	var querySamples []state.PostgresQuerySample
 
 	tls.CollectedAt = time.Now()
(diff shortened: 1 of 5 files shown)


uhoh-itsmaciek (approved):
Nice! This reads surprisingly straightforward.

Issue #87, opened 2020-09-15
Container running as root


JordanP:
Hi,
Is there a reason with the container is running as root ?

It looks like the Dockeerfile (https://github.com/pganalyze/collector/blob/master/Dockerfile#L4) is adding a user but no "USER" directive, that user is not actually used.

JordanP:
Hum, actually it's using gosu. Why not just use "USER" in the Dockerfile. Problem is, anybody that can "exec" into the container is actually "root".

lfittl:
@JordanP I think this is mostly for historic reasons, since Docker didn't always support running the container as non-root. We'll review whether we can move away from the `gosu`-based approach, and instead use the `USER` command.

We would of course also welcome a patch for this, if you'd like.

lfittl:
@JordanP FYI, we have a test image available at #174 that uses `USER` instead of gosu for the privilege drop. It would be great if you can give that a look and see whether it works for your environment.

lfittl:
Closing based on #174 being merged.

PR #174, opened 2021-04-06
Docker: Use Docker's USER command to set user, to support running as non-root

lfittl:
This enables the collector container to run in environments that require the
whole container to run as a non-root user, which previously was not the case.

For compatibility reasons the container can still be run as root explicitly,
in which case the setpriv command is used to drop privileges. setpriv replaces
gosu since its available for installation in most distributions directly, and
fulfills the same purpose here.

---

**Docker tag:** ```docker pull quay.io/pganalyze/collector:hotfix-docker-user```


uhoh-itsmaciek (approved):
Nice!

uhoh-itsmaciek:
Fixes #87, right?

PR #176, opened 2021-04-13
Statement stats retrieval: Get all rows first, before fingerprinting queries

lfittl:
This avoids showing a bogus ClientWrite event on the Postgres server side whilst
the collector is running the fingerprint method. There is a trade-off here,
because we now need to retrieve all statement texts (for the full snapshot) before
doing the fingerprint, leading to a slight increase in memory usage. Nonetheless,
this improves debuggability, and avoids bogus statement timeout issues.

---

Docker Tag: `quay.io/pganalyze/collector:hotfix-statement-timeout`


uhoh-itsmaciek (approved):

uhoh-itsmaciek:
Interesting test failure:

```
diff -Naur pg93.expected pg93.out && echo 'success'
--- pg93.expected	2021-04-13 17:55:47.516395075 +0000
+++ pg93.out	2021-04-13 17:59:59.830438282 +0000
@@ -1,5 +1,5 @@
-"BEGIN;"
-"END;"
+"begin"
+"commit"
 "INSERT INTO pgbench_history (tid, bid, aid, delta, mtime) VALUES (?, ?, ?, ?, CURRENT_TIMESTAMP);"
 "SELECT abalance FROM pgbench_accounts WHERE aid = ?;"
 "select count(*) from pgbench_branches"
make[1]: *** [Makefile:31: pg93] Error 1
```

Is this a pg_bench change?

lfittl:
> Interesting test failure:
> 
> ```
> diff -Naur pg93.expected pg93.out && echo 'success'
> --- pg93.expected	2021-04-13 17:55:47.516395075 +0000
> +++ pg93.out	2021-04-13 17:59:59.830438282 +0000
> @@ -1,5 +1,5 @@
> -"BEGIN;"
> -"END;"
> +"begin"
> +"commit"
>  "INSERT INTO pgbench_history (tid, bid, aid, delta, mtime) VALUES (?, ?, ?, ?, CURRENT_TIMESTAMP);"
>  "SELECT abalance FROM pgbench_accounts WHERE aid = ?;"
>  "select count(*) from pgbench_branches"
> make[1]: *** [Makefile:31: pg93] Error 1
> make[1]: Leaving directory '/home/runner/work/collector/collector/integration_test'
> ```
> 
> Is this a pg_bench change?

Yeah, this is related to a sorting problem, not an actual issue - sometimes this resolves on retry - need https://github.com/pganalyze/collector/pull/178 to fix tests in general, then I'll fix this one.

uhoh-itsmaciek:
Looks like pg_bench is using both "BEGIN;" and "begin". Maybe there's a race condition as to which ends up in pgss first, but they both fingerprint the same, so whichever one it is sets the query text? I'll try restarting one more time just in case it's a fluke, but maybe we'll need to normalize for the expected result comparison...

PR #184, opened 2021-06-04
Move master branch to main

uhoh-itsmaciek:

PR #185, opened 2021-06-04
Add support for CentOS to install script

uhoh-itsmaciek:
Assume we can treat it exactly like RHEL and install those packages.

Tried copying the script to a CentOS 7 instance, where the problem was
originally reported, and the collector installs fine with these
changes.

### DIFF
--- a/contrib/install.sh
+++ b/contrib/install.sh
@@ -57,19 +57,19 @@ then
   pkg=yum
   distribution=el
   version=7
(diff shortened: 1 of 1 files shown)


uhoh-itsmaciek:
We'll need to rebase over #184 to trigger CI once that lands.

lfittl (approved):
:+1:

PR #186, opened 2021-06-28
Update to pg_query_go v2.0.4

lfittl:


uhoh-itsmaciek (changes requested):
> -"select o.n, p.partstrat, pg_catalog.count(i.inhparent) from pg_catalog.pg_class as c join pg_catalog.pg_namespace as n on (n.oid = c.relnamespace) cross join lateral (select pg_catalog.array_position(pg_catalog.current_schemas($1), n.nspname)) as o(n) left join pg_catalog.pg_partitioned_table as p on (p.partrelid = c.oid) left join pg_catalog.pg_inherits as i on (c.oid = i.inhparent) where c.relname = $2 and o.n is not null group by $4, $5 order by $6 asc limit $3"
+"select o.n, p.partstrat, pg_catalog.count(i.inhparent) from pg_catalog.pg_class as c join pg_catalog.pg_namespace as n on (n.oid = c.relnamespace) cross join lateral (select pg_catalog.array_position(pg_catalog.current_schemas($1), n.nspname)) as o(n) left join pg_catalog.pg_partitioned_table as p on (p.partrelid = c.oid) left join pg_catalog.pg_inherits as i on (c.oid = i.inhparent) where c.relname = $2 and o.n is not null group by 1, 2 order by $4 asc limit $3"

The failure looks like what you're fixing here (`group by $4, $5` versus `group by 1, 2`), so we just need to update the expected files.

lfittl:
> The failure looks like what you're fixing here (`group by $4, $5` versus `group by 1, 2`), so we just need to update the expected files.

Yep, that was actually an excellent catch by the integration tests - I also realized that we should do the same treatment to the `ORDER BY` clause, so updated accordingly and bumped the library version (and fixed the tests).

uhoh-itsmaciek (approved):
:+1:

PR #187, opened 2021-06-30
Release 0.40.0

lfittl:


uhoh-itsmaciek (approved):
We could also mention the reduced memory usage for complex queries, but not strictly necessary.

lfittl:
> We could also mention the reduced memory usage for complex queries, but not strictly necessary.

You are right, maybe calling that out explicitly in the release notes is a good idea. Thanks!

PR #190, opened 2021-07-27
Retry API requests in case of temporary network issues

seanlinsley:
This PR adds https://github.com/hashicorp/go-retryablehttp to retry any network requests not directly handled by the AWS SDK (since it has its own retry mechanism).

If any of these non-SDK requests should be only tried once we'll need to switch them back to `server.Config.HTTPClient`.

Here's an example of a request being retried:

Screen Shot 2021-07-27 at 6 22 48 PM


CONTRIBUTING.md
```diff
@@ -0,0 +1,19 @@
+## Contributing Instructions
+
+### Setup
+
+```sh
+# install golang, then:
+go get -u github.com/goware/modvendor
+make vendor
```
lfittl, line 8: Maybe we can have this as a separate "When you update dependencies, run the following command to vendor all included files:" (since its technically not needed during setup, only when you're adding new modules)

CONTRIBUTING.md
```diff
@@ -0,0 +1,19 @@
...
+```
+
+### Compiling and running tests
+
+```sh
+make build
+make test
+# see the others in the Makefile
```
lfittl, line 16: Maybe we can directly reference the integration tests here, like this:

---

### Compiling and running tests

To compile the collector and helper binaries:

```
make build
```

After building the collector you can find the binary in the repository folder:

```
./pganalyze-collector --help
```

To run the unit tests:

```
make test
```

To run the integration tests:

```
make integration_test
```

config/read.go
```diff
@@ -279,9 +281,18 @@ func CreateHTTPClient(conf ServerConfig) *http.Client {
...
-	return &http.Client{
-		Timeout:   120 * time.Second,
-		Transport: transport,
+	if retry {
+		client := retryablehttp.NewClient()
+		client.HTTPClient.Timeout = 120 * time.Second
+		client.HTTPClient.Transport = transport
+		return client.StandardClient()
+		// Note: StandardClient() only acts as a passthrough, handing the request to
+		// retryablehttp.Client whose nested HTTP client ends up using our custom transport.
```
lfittl, line 290: :+1:

config/read.go
```diff
@@ -279,9 +281,18 @@ func CreateHTTPClient(conf ServerConfig) *http.Client {
 		transport.TLSClientConfig = &tls.Config{MinVersion: tls.VersionTLS12}
 	}
 
-	return &http.Client{
-		Timeout:   120 * time.Second,
-		Transport: transport,
+	if retry {
+		client := retryablehttp.NewClient()
+		client.HTTPClient.Timeout = 120 * time.Second
+		client.HTTPClient.Transport = transport
```
lfittl, line 287: It may be a good idea to explicitly set the number of retries here, so that we have it explicitly defined and protect against any unexpected changes by upstream. I think the upstream defaults (https://github.com/hashicorp/go-retryablehttp/blob/v0.7.0/client.go#L470) seem sensible, so something like this could work:

```
client.RetryWaitMin = 1 * time.Second
client.RetryWaitMax = 30 * time.Second
client.RetryMax = 4
```

main.go
```diff
@@ -61,7 +61,8 @@ func run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
 		prefixedLogger := logger.WithPrefix(server.SectionName)
 		prefixedLogger.PrintVerbose("Identified as api_system_type: %s, api_system_scope: %s, api_system_id: %s", server.SystemType, server.SystemScope, server.SystemID)
 
-		conf.Servers[idx].HTTPClient = config.CreateHTTPClient(server)
+		conf.Servers[idx].HTTPClient = config.CreateHTTPClient(server, false)
+		conf.Servers[idx].HTTPClientWithRetry = config.CreateHTTPClient(server, true)
```
lfittl, line 65: We could pass the `prefixedLogger` here so HTTP error logs have the right prefix (I suspect we may also need that for the existing HTTP client?)

config/read.go
lfittl: Oh, and I think we should also consider setting the Logger variable (and have it passed to this function call, per my comment below)

seanlinsley:
Okay @lfittl I pushed updates to this branch. Unfortunately, my attempt at unifying the loggers is introducing formatting issues:

Screen Shot 2021-07-28 at 2 27 04 PM

Maybe this would fix it? https://github.com/hashicorp/go-retryablehttp/pull/101

lfittl:
> Okay @lfittl I pushed updates to this branch. Unfortunately, my attempt at unifying the loggers is introducing formatting issues:
> 
> Screen Shot 2021-07-28 at 2 27 04 PM
> 
> Maybe this would fix it? [hashicorp/go-retryablehttp#101](https://github.com/hashicorp/go-retryablehttp/pull/101)

Hmm, interesting. Yeah, I think at the very least we'd need something that hides the verbose messages (unless one is running with `-v`).

I wonder if we should just set `Logger` to `nil`? That gives a clean output, and I don't think the retry behavior is that interesting to output to end users:

```
lfittl@starfish pganalyze-collector % ./pganalyze-collector --test   
2021/07/28 14:31:44 I Running collector test with pganalyze-collector 0.40.0
2021/07/28 14:31:44 I [server1] Testing statistics collection...
2021/07/28 14:31:59 E [server1] Could not process server: Get "http://localhost:5000/v2/snapshots/grant": GET http://localhost:5000/v2/snapshots/grant giving up after 5 attempt(s): Get "http://localhost:5000/v2/snapshots/grant": dial tcp 127.0.0.1:5000: connect: connection refused
```

Additionally, when looking at the log output, I noticed something else we should consider. Previously when the server collection failed, e.g. because the API server isn't reachable at all, the actual error message from the collector (ignoring any extra output we'd get from the HTTP client) was like this:

```
2021/07/28 14:10:10 I [server1] Testing statistics collection...
2021/07/28 14:10:10 E [server1] Could not process server: Get "http://localhost:5000/v2/snapshots/grant": \
dial tcp [::1]:5000: connect: connection refused
```

Now the message looks like this:

```
2021/07/28 14:31:59 E [server1] Could not process server: Get "http://localhost:5000/v2/snapshots/grant": GET http://localhost:5000/v2/snapshots/grant giving up after 5 attempt(s): Get "http://localhost:5000/v2/snapshots/grant": dial tcp 127.0.0.1:5000: connect: connection refused
```

I wonder if we can get rid of the first `Get "http://localhost:5000/v2/snapshots/grant"` at least? (I think that may be caused by wrapping the HTTP error somewhere)

seanlinsley:
Okay, `Logger` has been set to `nil` to remove the more verbose logs.

Retryablehttp [adds the URL twice](https://github.com/hashicorp/go-retryablehttp/blob/02c1586c8f14be23e7eeb522f1094afbabf45e93/client.go#L693-L694) to error messages, and we then wrap that error, causing it to show the URL three times. I settled on a regex hack to remove extra URLs from the message so we can still keep the `connection refused` detail.

Not perfect, but here's the current version of the errors when the server is down:
```
2021/07/29 10:06:15 E [server1] Could not collect activity for server: could not get default grant for activity snapshot: "http://localhost:5000/v2/snapshots/grant": dial tcp [::1]:5000: connect: connection refused
2021/07/29 10:06:20 E [server1] Could not get log grant: Get "http://localhost:5000/v2/snapshots/grant_logs": "http://localhost:5000/v2/snapshots/grant_logs": dial tcp [::1]:5000: connect: connection refused
```

And when the server responds with an error:
```
2021/07/29 10:18:17 E [server1] Could not collect activity for server: could not get default grant for activity snapshot: http://localhost:5000/v2/snapshots/grant giving up after 5 attempt(s)
2021/07/29 10:18:19 E [server1] Could not get log grant: Get "http://localhost:5000/v2/snapshots/grant_logs": http://localhost:5000/v2/snapshots/grant_logs giving up after 5 attempt(s)
```

If you think this isn't a terrible approach and would like them to be more consistent, I'll revisit the error parsing.

lfittl:
> Retryablehttp [adds the URL twice](https://github.com/hashicorp/go-retryablehttp/blob/02c1586c8f14be23e7eeb522f1094afbabf45e93/client.go#L693-L694) to error messages, and we then wrap that error, causing it to show the URL three times. I settled on a regex hack to remove extra URLs from the message so we can still keep the `connection refused` detail.
> 
> Not perfect, but here's the current version of the errors when the server is down:
> 
> ```
> 2021/07/29 10:06:15 E [server1] Could not collect activity for server: could not get default grant for activity snapshot: "http://localhost:5000/v2/snapshots/grant": dial tcp [::1]:5000: connect: connection refused
> 2021/07/29 10:06:20 E [server1] Could not get log grant: Get "http://localhost:5000/v2/snapshots/grant_logs": "http://localhost:5000/v2/snapshots/grant_logs": dial tcp [::1]:5000: connect: connection refused
> ```
> 
> And when the server responds with an error:
> 
> ```
> 2021/07/29 10:18:17 E [server1] Could not collect activity for server: could not get default grant for activity snapshot: http://localhost:5000/v2/snapshots/grant giving up after 5 attempt(s)
> 2021/07/29 10:18:19 E [server1] Could not get log grant: Get "http://localhost:5000/v2/snapshots/grant_logs": http://localhost:5000/v2/snapshots/grant_logs giving up after 5 attempt(s)
> ```
> 
> If you think this isn't a terrible approach and would like them to be more consistent, I'll revisit the error parsing.

I think something that parses the error message could work, but I'm not a fan of doing this in the central logger, since we're introducing knowledge about a very specific error into the generic codepath.

Maybe instead we could do this at the individual call sites, e.g. by having a `FriendlyHTTPErrorMessage` method (or something like that) that is called before we pass back the error from functions that are using the retrying HTTP client.

seanlinsley:
Okay, the log cleaner is now in a separate `util` function. Now all four scenarios only show the URL once:
```
2021/07/30 08:58:05 E [server1] Could not collect activity for server: could not get default grant for activity snapshot: "http://localhost:5000/v2/snapshots/grant": dial tcp [::1]:5000: connect: connection refused
2021/07/30 08:58:18 E [server1] Could not get log grant: "http://localhost:5000/v2/snapshots/grant_logs": dial tcp [::1]:5000: connect: connection refused
```

```
2021/07/30 09:01:48 E [server1] Could not collect activity for server: could not get default grant for activity snapshot: http://localhost:5000/v2/snapshots/grant giving up after 5 attempt(s)
2021/07/30 09:01:52 E [server1] Could not get log grant: http://localhost:5000/v2/snapshots/grant_logs giving up after 5 attempt(s)
```

lfittl (approved):
:+1:

PR #191, opened 2021-07-29
Fix citus relation stats sql query typo

rauanmayemir:
This typo was causing failure for pganalyze table stats when used on DBs with citus extension enabled.
New release would be appreciated with this fix in, we can't see any stats in pganalyze due to this issue.


lfittl (approved):

lfittl:
@rauanmayemir Thanks for the contribution!

We should be able to get a new release out in the next 1-2 days.

rauanmayemir:
Gently pinging to ask if there's going to be a release soon.

PR #192, opened 2021-08-06
Store column statistics from pg_stats

seanlinsley:
This PR submits  `pg_stats` column statistics as part of the full snapshot every 10 minutes.

Notes:
- `full_snapshot.pb.go` and `shared.pb.go` have a lot more changes than I expected (the index/reference numbers incrementing on many types). Is that an issue?
- CI is failing because https://github.com/pganalyze/collector-snapshot/pull/13 needs to be merged


uhoh-itsmaciek:
Re: the generated `.pb.go` changes: I think (some of?) that might be because of https://github.com/pganalyze/collector/pull/192/files#diff-ccc0d3ed9e9e0ccadcb9951f69f63c0c683f940743d197fc3a92fb663e1b7632L3118-R3177 -- the NullDouble bumps out the other types in the array, renumbering everything.

That said, I noticed you're using protoc v3.15.8 (versus v3.14.0 for the existing generated files). It's probably fine, but maybe we should do that upgrade in a separate patch, and then rebase this over that? I don't feel strongly about that, just being cautious since I have limited protobuf experience.

I'll review the actual collection changes now.

uhoh-itsmaciek:
Oh and re: failing CI--we've been pointing git submodules at a branch to get CI passing before the dependency PR is merged, but that does require re-pointing after the dependency lands, so it's fine to do this as well (as long as CI re-runs after the dep merge).

input/postgres/relation_stats.go
```diff
@@ -156,3 +169,52 @@ func GetIndexStats(db *sql.DB, postgresVersion state.PostgresVersion, ignoreRege
...
+		var s state.PostgresColumnStats
+		var key string
+
+		err := rows.Scan(
+			&s.SchemaName, &s.TableName, &s.ColumnName, &s.Inherited, &s.NullFrac, &s.AvgWidth, &s.NDistinct, &s.Correlation)
+		if err != nil {
+			return nil, err
+		}
+
+		key = s.SchemaName + s.TableName + s.ColumnName
```
uhoh-itsmaciek, line 215: This isn't safe if both table `foo`, column `barbaz` and table `foobar`, column `baz` exist, I think. That's probably not very likely, but maybe it's likely enough that we should handle it better. It's a shame Go makes multilevel maps a pain--can't think of a super elegant solution here.

seanlinsley:
>That said, I noticed you're using protoc v3.15.8 (versus v3.14.0 for the existing generated files). It's probably fine, but maybe we should do that upgrade in a separate patch, and then rebase this over that? I don't feel strongly about that, just being cautious since I have limited protobuf experience.

I'm not seeing any lockfile-like version control for the protobuf CLI, so I'm not sure how we should handle this as a longer term issue. Maybe just specifying a certain version in `CONTRIBUTING.md`? CC @lfittl

uhoh-itsmaciek (changes requested):
So, overall this looks good, but I think we should do something about keying the column stats map. Maybe building a proper three-level map is overkill, but maybe we should key by schema oid (do we have that?) + `<delimiter>` + relation oid + `<delimiter>` + column position? Using numeric key pieces makes it easy to ensure our delimiter doesn't collide and we don't have aliasing. Alternately, I think it's okay if we use an uncommon character like `\t` as delimiter: if you're using that in your table or column names, you deserve misery.

uhoh-itsmaciek:
>I'm not seeing any lockfile-like version control for the protobuf CLI, so I'm not sure how we should handle this as a longer term issue. Maybe just specifying a certain version in CONTRIBUTING.md?

Yeah, I think that works. We haven't been changing this often, so it hasn't been an issue, but you're right it's awkward right now (you'll generally only see the issue if you're inspecting the generated output).

input/postgres/relation_stats.go
seanlinsley: I'll update the key to use `\t` as a delimiter for now. Unless @lfittl thinks a nested map is needed?

uhoh-itsmaciek:
For what it's worth, I don't object to bumping protoc whenever--I just think doing so in a separate patch (and regenerating based on the same definitions) would make it easier to debug any potential problems.

uhoh-itsmaciek (approved):
Cool, I think the chances of actual accidental real-world aliasing with this approach are vanishingly unlikely, and an adversarial scenario would be pretty far-fetched.

state/postgres_relation_stats.go
```diff
@@ -43,8 +43,20 @@ type PostgresIndexStats struct {
...
+  Inherited   bool
+  NullFrac    float64
+  AvgWidth    int32
+  NDistinct   float64
+  Correlation null.Float
+}
+
 type PostgresRelationStatsMap map[Oid]PostgresRelationStats
 type PostgresIndexStatsMap map[Oid]PostgresIndexStats
+type PostgresColumnStatsMap map[string][]PostgresColumnStats
```
lfittl, line 59: Can we use a more strongly typed key here? (see e.g. `PostgresStatementKey` in `state/postgres_statement.go`)

output/transform/postgres_relations.go
```diff
@@ -63,15 +63,38 @@ func transformPostgresRelations(s snapshot.FullSnapshot, newState state.Persiste
...
 
+		schemaStats, schemaStatsExist := newState.SchemaStats[relation.DatabaseOid]
+
 		if relation.ViewDefinition != "" {
 			info.ViewDefinition = &snapshot.NullString{Valid: true, Value: relation.ViewDefinition}
 		}
 		for _, column := range relation.Columns {
+			var stats []*snapshot.RelationInformation_ColumnStatistic
+			if schemaStatsExist {
+				key := relation.SchemaName + "\t" + relation.RelationName + "\t" + column.Name
```
lfittl, line 74: Per the note in the state file, I think it would be better to use a strongly typed key (i.e. have a struct as the key instead of a string), similar to how we do it for statements.

output/transform/postgres_relations.go
```diff
@@ -63,15 +63,38 @@ func transformPostgresRelations(s snapshot.FullSnapshot, newState state.Persiste
...
+					for _, stat := range columnStats {
+						correlation := snapshot.NullDouble{Valid: false}
+						if stat.Correlation.Valid {
+							correlation = snapshot.NullDouble{Valid: true, Value: stat.Correlation.Float64}
+						}
+						stats = append(stats, &snapshot.RelationInformation_ColumnStatistic{
+							Inherited: stat.Inherited,
+							AvgWidth: stat.AvgWidth,
+							NDistinct: stat.NDistinct,
+							Correlation: &correlation,
```
lfittl, line 86: Are we missing `NullFrac` here?

input/postgres/relation_stats.go
```diff
@@ -156,3 +169,52 @@ func GetIndexStats(db *sql.DB, postgresVersion state.PostgresVersion, ignoreRege
...
+	var sourceTable string
+
+	helperExists := false
+	db.QueryRow(QueryMarkerSQL + columnStatsHelperSQL).Scan(&helperExists)
+
+	if helperExists {
+		logger.PrintVerbose("Found pganalyze.get_column_stats() stats helper")
+		sourceTable = "pganalyze.get_column_stats()"
+	} else {
+		if systemType != "heroku" && !connectedAsSuperUser(db, systemType) && !connectedAsMonitoringRole(db) && globalCollectionOpts.TestRun {
```
lfittl, line 183: I think we can need to drop the `connectedAsMonitoringRole` part of this condition, since `pg_monitor` does not grant access to `pg_stats` (i.e. the user does need to create the helper always, even on newer Postgres versions)

input/postgres/relation_stats.go
```diff
@@ -156,3 +169,52 @@ func GetIndexStats(db *sql.DB, postgresVersion state.PostgresVersion, ignoreRege
...
+			logger.PrintInfo("Warning: You are not connecting as superuser. Please setup" +
+				" the monitoring helper functions (https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user)" +
+				" or connect as superuser, to get table statistics for all roles.")
```
lfittl, line 186: Since this is a new addition in terms of requirements (the in-app setup flow right now does not tell you to create this function), maybe its helpful if we specifically call out the function thats missing, like this:

```
			logger.PrintInfo("Warning: Limited access to table column statistics detected. Please setup" +
				" the monitoring helper function pganalyze.get_column_stats (https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user)" +
				" or connect as superuser, to get column statistics for all tables.")
```

input/postgres/relation_stats.go
```diff
@@ -72,6 +73,18 @@ SELECT s.indexrelid,
 			 AND ($1 = '' OR (s.schemaname || '.' || s.relname) !~* $1)
 `
 
+const columnStatsSQL = `
+SELECT schemaname, tablename, attname, inherited, null_frac, avg_width, n_distinct, correlation
+  FROM %s
```
lfittl, line 78: I think we can ignore pg_catalog here, i.e.

```sql
SELECT schemaname, tablename, attname, inherited, null_frac, avg_width, n_distinct, correlation
  FROM %s
 WHERE schemaname <> 'pg_catalog'
```

input/postgres/relation_stats.go
lfittl: Commented on this separately - I think we should match the pattern we have for statements (use a struct as the key)

input/postgres/relation_stats.go
seanlinsley: Good catch. Looks like `information_schema` should be skipped too.

output/transform/postgres_relations.go
seanlinsley: 🙀 not sure how I missed that

input/postgres/relation_stats.go
```diff
@@ -156,3 +170,51 @@ func GetIndexStats(db *sql.DB, postgresVersion state.PostgresVersion, ignoreRege
...
+	var sourceTable string
+
+	helperExists := false
+	db.QueryRow(QueryMarkerSQL + columnStatsHelperSQL).Scan(&helperExists)
+
+	if helperExists {
+		logger.PrintVerbose("Found pganalyze.get_column_stats() stats helper")
+		sourceTable = "pganalyze.get_column_stats()"
+	} else {
+		if systemType != "heroku" && !connectedAsSuperUser(db, systemType) && globalCollectionOpts.TestRun {
```
uhoh-itsmaciek, line 184: Oof, so `pg_monitor` is not enough here, right? I just tested and that seems to be the case. We'll always need the helper to collect stats unless running as superuser...

This is unfortunate, since stats for columns we don't have permissions to collect are just silently excluded. This means that anyone upgrading the collector will also need to run this additional step--it's outside the scope of this patch, but we should think about that workflow or we just won't get anyone actually collecting any column stats (I expect putting something in the release notes is not enough).

input/postgres/relation_stats.go
```diff
@@ -156,3 +170,51 @@ func GetIndexStats(db *sql.DB, postgresVersion state.PostgresVersion, ignoreRege
...
+			logger.PrintInfo("Warning: Limited access to table column statistics detected. Please setup" +
```
uhoh-itsmaciek, line 185: We should probably use `PrintVerbose` here--the cases where we `PrintInfo` for missing helpers seem to be in the (unused) collector reports. If this is running regularly it would be pretty noisy. Plus, superuser is not available on some systems, so there's not much you can do about the warning.

uhoh-itsmaciek (approved):
We should change that PrintInfo in the error message, but otherwise this looks good. I'd forgotten you could key maps by struct value--that makes proper handling of aliasing relatively painless.

Separately, we should document the helper and think about how to make sure upgrading users know they need to change their config, but that's outside the scope of this patch.

input/postgres/relation_stats.go
uhoh-itsmaciek: Oh, sorry--I missed that this only happens for a test run. It should stay as is.

input/postgres/relation_stats.go
seanlinsley: How did you test this, exactly? I had just copied this check from `input/postgres/relation_bloat.go` because it also uses `pg_stats`. Though it's slightly different from that version because of @lfittl's comment earlier: https://github.com/pganalyze/collector/pull/192#discussion_r689674107

uhoh-itsmaciek: Ah, I missed Lukas' comment--I just tested locally: I created a new role in my local Postgres and checked if it could see pg_stats rows for a table from my default user (it could not). I granted `pg_monitor` to the new role and it could still not see those stats. I then granted my user to that role and it could see them. I think the check is right, and there's nothing else to do in the PR--we just need to think about how to communicate that users need to add the helper if they want accurate stats for the index advisor.

PR #193, opened 2021-08-12
Vendor protoc to keep the version consistent

seanlinsley:
Since there isn't an easy way to keep us all on the same version of protobuf/protoc, I wrote a semi-automated script to do that for us.

Here's what you see when there's a version mismatch:

>⚠️  protoc version needed: 3.14.0 vs none installed
>ℹ️  Please download the correct protobuf binary for your OS from https://github.com/protocolbuffers/protobuf/releases/tag/v3.14.0
>ℹ️  Note the download's name will look like this: protoc-3.14.0-osx-x86_64.zip
>ℹ️  Copy the unzipped folder into this project, and rename it to "protoc"
>ℹ️  If this is macOS, you will need to try running the binary yourself, then go to Security & Privacy to explicitly allow it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Set up protoc
       uses: arduino/setup-protoc@v1.1.2
       with:
-        version: 3.x
+        version: 3.14.0
 
     - name: Check out code
       uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


uhoh-itsmaciek (commented):
Nice, I like that you pinned protoc and you're ensuring the right version. Since we do get occasional third-party (usually customer) contributions to the collector code, and since protobuf changes are relatively rare, it might be nice not to raise the bar for contributions there--i.e., to still provide a "fast path" workflow where you don't need to worry about protoc. Thougths?

seanlinsley:
Oh that's a good point. I'll add back the code that skipped protobuf regeneration, instead printing this:
>Warning: protoc not found, skipping protocol buffer regeneration

uhoh-itsmaciek (approved):

PR #196, opened 2021-09-27
Track custom data types

seanlinsley:
This PR collects the relevant information for custom data types (enums, domains, composite types) so we can recreate their DDL statement for use with Index Advisor.


state/postgres_types.go
```diff
@@ -1,4 +1,21 @@
...
+type PostgresType struct {
+	Oid             Oid
+	DatabaseOid     Oid
+	SchemaName      string
+	Name            string
+	Type            string
+	UnderlyingType  null.String
+	NotNull         bool
+	Default         null.String
+	Constraint      null.String
```
lfittl, line 18: I wonder if we should prefix these four attributes with `Domain`?

In my understanding they are not used with the other types, and that makes it easier to understand what needs to be looked at for a given type.

lfittl (commented):
I think this looks great!

For getting data out of the enum/composite value fields, I wonder if we should just use something more basic like `jsonb_agg` in Postgres (and then unmarshal in the Go code). That would avoid any escaping issues, and supported nested structures.

But it does have a bit of overhead for the unmarshaling, compared to `unpackPostgresStringArray` or a similar array-specific parser.

seanlinsley:
I went ahead with the JSON approach in https://github.com/pganalyze/collector/pull/196/commits/96ab4124b78c3ae7e503bd7da437dd1f4c4bee2d. I did at least write it in a way that we won't be parsing JSON twice for each row when there would only ever be a single result.

output/transform/postgres_types.go
```diff
@@ -0,0 +1,43 @@
...
+
+  for _, pgType := range transientState.Types {
+    var customType snapshot.CustomTypeInformation_Type
+    switch pgType.Type {
+    case "e":
+      customType = snapshot.CustomTypeInformation_ENUM
+    case "d":
+      customType = snapshot.CustomTypeInformation_DOMAIN
+    case "c":
+      customType = snapshot.CustomTypeInformation_COMPOSITE
```
lfittl, line 19: Thinking through this again, I think we either need to:

(1) Add all custom type types here (i.e. missing would be `p` for a pseudo-type, `r` for a range type, or `m` for a multirange type)

(2) Filter out these other types in the input query, like we filter out `b` today

The benefit of (1) would be that we know how many people are using range types, to determine whether its worth supporting it. But not sure if its worth the effort.

lfittl (approved):
One more minor comment re: range types - but otherwise this looks ready to ship :)

output/transform/postgres_types.go
seanlinsley: Tracking all custom types makes sense, like you said, so we have a better idea what types our customers are using. I'll also update it to store `b` types.

seanlinsley: Ah, one thing to note about the base types is that e.g. btree_gist types get captured.

Screen Shot 2021-10-03 at 9 52 37 AM

An earlier version of this PR filtered out types that aren't used by columns, but I removed that to speed up the query. They could be filtered by the Go code since at that point we have the columns, but that would mean looping over the columns an extra time.

Thoughts?

seanlinsley: I went ahead and kept base types excluded from the query for now.

lfittl (approved):
👍

PR #212, opened 2021-10-29
Change schema collection message to warning

uhoh-itsmaciek:
If we fail to collect schema metadata, a database can remain
permanently hidden and a lot of in-app functionality does not
work. Since verbose logs are not likely to be turned on for a running
collector, hiding this message makes this situation difficult to
discover.

Elevate the message to a warning.

### DIFF
--- a/input/postgres/schema.go
+++ b/input/postgres/schema.go
@@ -37,7 +37,12 @@ func CollectAllSchemas(server *state.Server, collectionOpts state.CollectionOpts
 		collected[dbName] = true
 		psNext, tsNext, databaseOid, err := collectOneSchema(server, collectionOpts, logger, ps, ts, dbName, ts.Version, systemType)
 		if err != nil {
-			logger.PrintWarning("Failed to collect schema metadata for database %s: %s", dbName, err)
+			warning := "Failed to collect schema metadata for database %s: %s"
+			if collectionOpts.TestRun {
+				logger.PrintWarning(warning, dbName, err)
+			} else {
+				logger.PrintVerbose(warning, dbName, err)
+			}
 			continue
 		}
 		ps = psNext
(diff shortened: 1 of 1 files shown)


input/postgres/schema.go
```diff
@@ -37,7 +37,7 @@ func CollectAllSchemas(server *state.Server, collectionOpts state.CollectionOpts
 		collected[dbName] = true
 		psNext, tsNext, databaseOid, err := collectOneSchema(server, collectionOpts, logger, ps, ts, dbName, ts.Version, systemType)
 		if err != nil {
-			logger.PrintVerbose("Failed to collect schema metadata for database %s: %s", dbName, err)
+			logger.PrintWarning("Failed to collect schema metadata for database %s: %s", dbName, err)
```
lfittl, line 40: I wonder if we should make this dependent on `collectionOpts.TestRun`? (i.e. only output it as a warning during the test run - otherwise this becomes very repetitive, e.g. in cases where we are missing a cloud provider internal DB from the `isCloudInternalDatabase` list)

uhoh-itsmaciek: Yeah, this was banged out quickly on our support triage call with Sean, and we talked about that issue a bit--I mostly opened the PR as a straw man. We think this particular customer may have been hitting https://github.com/pganalyze/collector/pull/211 but it's hard to diagnose because this error is only logged in verbose mode. I think it's pretty unlikely anyone is actually running collection with verbose mode turned on (and instructing customers on how to turn it on is also complicated, especially given variations for the different package/Docker/Heroku deployment mechanisms and that it's not obvious how a customer has deployed the collector).

I think restricting it to test runs for now is a good start. But do customers re-test the collector after upgrading it? I'm not arguing for this change specifically, but I do think we should offer more visibility into collection issues to give customers confidence things are set up correctly (and clear diagnostics when they're not).

seanlinsley (approved):
personally I think we should raise this as a warning all the time, but raising it during a test run is an improvement

uhoh-itsmaciek:
After sleeping on this, I think a warning all the time is probably not the right way to communicate this issue. We're treating it inconsistently right now: the collector just issues an easy-to-miss warning, but the app treats this as a drastic problem (to the point where you can't really use pganalyze on a database that encounters this issue). But we *know* the schema information is missing in-app, so we can do a better job of communicating the problem there. That's sure to be more effective than more log messages.

Given that, I'll merge this for now (with the change to only print a warning in a test run), but let's plan to improve this in terms of the overall experience. We've definitely seen this as a barrier to a smooth onboarding in support and it's important to have a better experience here.

PR #214, opened 2021-11-08
Add basic 'go fmt' linting to CI

uhoh-itsmaciek:
- Run go fmt ./...
- Add linting to CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,9 @@ SHELL := env PATH=$(PATH) /bin/sh
 PROTOC_VERSION_NEEDED := 3.14.0
 PROTOC_VERSION := $(shell which protoc && protoc --version)
 
-.PHONY: default build build_dist vendor test docker_latest packages integration_test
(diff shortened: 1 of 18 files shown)


Makefile
```diff
@@ -35,6 +35,9 @@ test: build
 	go test -coverprofile=coverage.out ./...
 	# go tool cover -html=coverage.out
 
+lint:
+	if [ "`go fmt ./... | wc -l`" -gt 0 ]; then exit 1; fi
```
uhoh-itsmaciek, line 39: On second thought, this should be `gofmt -l ./...`; will update.

uhoh-itsmaciek: Maybe I should just remove `lint` from the default target and just have CI run this?

uhoh-itsmaciek: I've opted to drop the Makefile changes and only run the check in CI.

seanlinsley (approved):

PR #215, opened 2021-11-08
Support multiple domain constraints in custom types

uhoh-itsmaciek:
Currently, when we find a custom domain type with multiple
constraints, we error out in the collection query, and consequently
fail to collect any schema information at all.

Instead, capture all domain constraints with an updated snapshot
format.

See also https://github.com/pganalyze/collector-snapshot/pull/16


seanlinsley (approved):
the code changes look good. where the indentation changes on purpose though?

uhoh-itsmaciek:
The indentation changes are just running `go fmt`, which my editor does by default. I can rebase this over #214 to make it cleaner, but I don't think avoiding `go fmt` is worthwhile.

seanlinsley:
ah okay, no worries then

PR #221, opened 2021-11-29
Table processing: Ensure we use -1 index for unknown parent/FK tables

lfittl:
Previously we would sometimes use 0 in these situations, which is a valid
index when referencing the very first table in the list.

Note that whilst its not always clear why bad references here happen, as
they should only happen when the schema_table_regexp logic is active,
its better to not send bogus data to the server.

### DIFF
--- a/state/util.go
+++ b/state/util.go
@@ -15,7 +15,11 @@ func (m OidToIdxMap) Put(dbOid, objOid Oid, idx int32) {
 
 func (m OidToIdxMap) Get(dbOid, objOid Oid) int32 {
 	if _, ok := m[dbOid]; !ok {
-		return 0
+		return -1
 	}
-	return m[dbOid][objOid]
+	idx, ok := m[dbOid][objOid]
+	if !ok {
+		return -1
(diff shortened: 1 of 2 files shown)


uhoh-itsmaciek (approved):

state/util.go
```diff
@@ -15,7 +15,11 @@ func (m OidToIdxMap) Put(dbOid, objOid Oid, idx int32) {
 
 func (m OidToIdxMap) Get(dbOid, objOid Oid) int32 {
 	if _, ok := m[dbOid]; !ok {
-		return 0
+		return -1
```
uhoh-itsmaciek, line 18: Other than finding it hard to believe I wrote something this bogus in the first place, the patch looks good.

seanlinsley (approved):

state/util.go
lfittl: I think I wrote the original problem code here in https://github.com/pganalyze/collector/pull/61, you just expanded on it :-)

PR #222, opened 2021-12-03
Cache RDS server IDs to reduce API requests

seanlinsley:
`FindRdsInstance` is being called 21 times every 10 minutes (once for the full snapshot, once every 30 seconds for log snapshots), which is an issue for customers with many servers.

This PR caches the API requests for log snapshots only, with a TTL of 5 minutes. This should bring us down to 3 requests every 10 minutes.

Edit: this now also caches API errors so we will pause this API request for 10 minutes if there is an outage / a server is being reprovisioned.

### DIFF
--- a/util/awsutil/rds.go
+++ b/util/awsutil/rds.go
@@ -11,7 +11,7 @@ import (
 	"github.com/pganalyze/collector/util"
 )
 
-var IdentifierMap *util.TTLMap = util.NewTTLMap(5*60, 30)
+var IdentifierMap *util.TTLMap = util.NewTTLMap(5*60)
 
 func FindRdsIdentifier(config config.ServerConfig, sess *session.Session) (identifier string, err error) {
 	identifier = IdentifierMap.Get(config.AwsDbInstanceID)
--- a/util/awsutil/rds.go
+++ b/util/awsutil/rds.go
@@ -11,6 +11,20 @@ import (
 	"github.com/pganalyze/collector/util"
 )
 
+var IdentifierMap *util.TTLMap = util.NewTTLMap(5 * 60, 30)
+
+func FindRdsIdentifier(config config.ServerConfig, sess *session.Session) (identifier string, err error) {
+	identifier = IdentifierMap.Get(config.AwsDbInstanceID)
+	if identifier == "" {
+		instance, err := FindRdsInstance(config, sess)
+		if err != nil {
+			identifier = *instance.DBInstanceIdentifier
+			IdentifierMap.Put(config.AwsDbInstanceID, identifier)
+		}
+	}
+	return
+}
+
 func FindRdsInstance(config config.ServerConfig, sess *session.Session) (instance *rds.DBInstance, err error) {
 	var resp *rds.DescribeDBInstancesOutput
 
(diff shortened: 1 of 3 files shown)


util/ttl_map.go
```diff
@@ -0,0 +1,56 @@
...
+	m = &TTLMap{m: make(map[string]*item)}
+	go func() {
+		for now := range time.Tick(time.Second * time.Duration(checkFrequency)) {
+			m.l.Lock()
+			for k, v := range m.m {
+				if now.Unix()-v.createdAt > maxTTL {
+					delete(m.m, k)
+				}
+			}
+			m.l.Unlock()
```
uhoh-itsmaciek, line 28: NBD, but I think idiomatic Go defers the unlock right after the lock to prevent leaked locks if, e.g., code maintenance introduces early returns. I.e.

```golang
m.l.Lock()
defer m.l.Unlock()
```

util/ttl_map.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+type TTLMap struct {
+	m map[string]*item
+	l sync.Mutex
+}
+
+func NewTTLMap(maxTTL int64, checkFrequency int64) (m *TTLMap) {
+	m = &TTLMap{m: make(map[string]*item)}
+	go func() {
```
uhoh-itsmaciek, line 20: I think we could avoid the goroutine if we just check the ttl in Get before returning. We'd leak stale items, though (at least until someone tries to fetch them), so this works, too.

uhoh-itsmaciek (approved):
Nice!

util/ttl_map.go
seanlinsley: Since we already have memory leak issues, it's probably going to be better to keep the goroutine.

lfittl: FWIW, I'm also a bit inclined to suggest simplifying this here and relying on checking the TTL when `Get` is called, and then deleting the item.

If we're concerned about memory usage we could check the TTL of all items when any item is retrieved or stored (and delete if TTL is exceeded) - whilst that increases access costs to the map slightly, it avoids a Go routine running in the background (which makes debugging harder, etc).

Another benefit of that would be that we can drop the `checkFrequency` parameter to NewTTLMap.

seanlinsley: Okay, I updated the PR with the simpler option. I don't myself know if `AwsDbInstanceID` is likely to change or if it's mostly constant. We can always revisit if there is a memory leak here.

util/awsutil/rds.go
```diff
@@ -11,6 +11,20 @@ import (
 	"github.com/pganalyze/collector/util"
 )
 
+var IdentifierMap *util.TTLMap = util.NewTTLMap(5*60, 30)
+
+func FindRdsIdentifier(config config.ServerConfig, sess *session.Session) (identifier string, err error) {
+	identifier = IdentifierMap.Get(config.AwsDbInstanceID)
+	if identifier == "" {
+		instance, err := FindRdsInstance(config, sess)
+		if err != nil {
```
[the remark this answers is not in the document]
seanlinsley: Good idea! I've added that. Though I went with a 10 minute TTL for errors instead of 1 hour because I'm worried that temporary network issues could cause us to drop an hour of logs fairly often.

uhoh-itsmaciek (approved):
:+1: on latest changes

seanlinsley:
I deployed this code to our collector instance and logs are still being collected (using that as a basic system test since it's hard to write unit tests for this).

PR #225, opened 2021-12-10
Handle unexpected schema / version in pg_stat_statements

seanlinsley:
It's fairly common for customers to upgrade Postgres without realizing they need to upgrade pg_stat_statements as well. In that case, we show them an unhelpful error message:

```
Error collecting pg_stat_statements: pq: column "total_exec_time" does not exist
```

This PR updates the collector to instead use the correct `pg_stat_statements` columns based on the installed version of the extension, instead of the version we assume is installed based on the Postgres version.

If we know the extension is out of date we'll log a helpful message: (only during test runs)

```
pg_stat_statements extension outdated (1.6 installed, 1.8 available). To update run `ALTER EXTENSION pg_stat_statements UPDATE`
```

This PR also prevents errors when the extension is installed in a non-public schema, instead calling it in whatever schema it's defined.

### DIFF
--- a/input/postgres/statements.go
+++ b/input/postgres/statements.go
@@ -104,7 +104,7 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
 
 		var extSchema string
 		var foundExtVersion float32
-		err = db.QueryRow("SELECT nspname, extversion FROM pg_extension pge INNER JOIN pg_namespace pgn ON pge.extnamespace = pgn.oid WHERE pge.extname = 'pg_stat_statements'").Scan(&extSchema, &foundExtVersion)
+		err = db.QueryRow(QueryMarkerSQL+"SELECT nspname, extversion FROM pg_extension pge INNER JOIN pg_namespace pgn ON pge.extnamespace = pgn.oid WHERE pge.extname = 'pg_stat_statements'").Scan(&extSchema, &foundExtVersion)
 		if err != nil && err != sql.ErrNoRows {
 			return nil, nil, nil, err
 		}
--- a/input/postgres/statements.go
+++ b/input/postgres/statements.go
@@ -83,28 +83,6 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
 	var totalTimeField string
 	var optionalFields string
 	var sourceTable string
-	var extVersion float32               // defined in pg_stat_statements.c
-	var maxKnownExtVersion float32 = 1.8 // keeps the collector from erroring out when new Postgres versions are released
-
-	if postgresVersion.Numeric >= state.PostgresVersion13 {
-		totalTimeField = statementSQLpg13TotalTimeField
-	} else {
-		totalTimeField = statementSQLDefaultTotalTimeField
-	}
-
-	if postgresVersion.Numeric >= state.PostgresVersion13 {
-		optionalFields = statementSQLpg13OptionalFields
-		extVersion = 1.8
-	} else if postgresVersion.Numeric >= state.PostgresVersion95 {
-		optionalFields = statementSQLpg95OptionalFields
-		extVersion = 1.3
-	} else if postgresVersion.Numeric >= state.PostgresVersion94 {
-		optionalFields = statementSQLpg94OptionalFields
-		extVersion = 1.2
-	} else {
-		optionalFields = statementSQLDefaultOptionalFields
-		extVersion = 1.1
-	}
 
 	usingStatsHelper := false
 
@@ -123,10 +101,59 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
 				" the monitoring helper functions (https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user)" +
 				" or connect as superuser, to get query statistics for all roles.")
 		}
-		if !showtext {
-			sourceTable = "public.pg_stat_statements(false)"
+
+		var extSchema string
+		var foundExtVersion float32
+		err = db.QueryRow("SELECT nspname, extversion FROM pg_extension pge INNER JOIN pg_namespace pgn ON pge.extnamespace = pgn.oid WHERE pge.extname = 'pg_stat_statements'").Scan(&extSchema, &foundExtVersion)
+		if err != nil && err != sql.ErrNoRows {
+			return nil, nil, nil, err
+		}
+
+		var extVersion float32 // defined in pg_stat_statements.c
+		if postgresVersion.Numeric >= state.PostgresVersion13 {
+			extVersion = 1.8
+		} else if postgresVersion.Numeric >= state.PostgresVersion95 {
+			extVersion = 1.3
+		} else if postgresVersion.Numeric >= state.PostgresVersion94 {
+			extVersion = 1.2
 		} else {
-			sourceTable = "public.pg_stat_statements"
+			extVersion = 1.1
+		}
+
+		if err == sql.ErrNoRows {
+			logger.PrintInfo("pg_stat_statements does not exist, trying to create extension...")
+			_, err = db.Exec(QueryMarkerSQL + "CREATE EXTENSION IF NOT EXISTS pg_stat_statements SCHEMA public")
+			if err != nil {
+				return nil, nil, nil, err
+			}
+			extSchema = "public"
+			foundExtVersion = extVersion
+		}
+
+		if foundExtVersion >= 1.8 {
+			optionalFields = statementSQLpg13OptionalFields
+		} else if foundExtVersion >= 1.3 {
+			optionalFields = statementSQLpg95OptionalFields
+		} else if foundExtVersion >= 1.2 {
+			optionalFields = statementSQLpg94OptionalFields
+		} else {
+			optionalFields = statementSQLDefaultOptionalFields
+		}
+
+		if foundExtVersion >= 1.8 {
+			totalTimeField = statementSQLpg13TotalTimeField
+		} else {
+			totalTimeField = statementSQLDefaultTotalTimeField
+		}
+
+		if foundExtVersion < extVersion {
+			logger.PrintInfo("pg_stat_statements extension outdated (%.1f installed, %.1f available). To update run `ALTER EXTENSION pg_stat_statements UPDATE`", foundExtVersion, extVersion)
+		}
+
+		if !showtext {
+			sourceTable = extSchema + ".pg_stat_statements(false)"
+		} else {
+			sourceTable = extSchema + ".pg_stat_statements"
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


input/postgres/statements.go
```diff
@@ -131,10 +137,14 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
 		var e *pq.Error
 		if !usingStatsHelper && errors.As(err, &e) && (e.Code == "42P01" || e.Code == "42883") { // undefined_table / undefined_function
 			var pgssSchema string
-			err = db.QueryRow("SELECT nspname FROM pg_extension pge INNER JOIN pg_namespace pgn ON pge.extnamespace = pgn.oid WHERE pge.extname = 'pg_stat_statements'").Scan(&pgssSchema)
+			var foundExtVersion float32
+			err = db.QueryRow("SELECT nspname, extversion FROM pg_extension pge INNER JOIN pg_namespace pgn ON pge.extnamespace = pgn.oid WHERE pge.extname = 'pg_stat_statements'").Scan(&pgssSchema, &foundExtVersion)
 			if err == nil && pgssSchema != "public" {
 				return nil, nil, nil, fmt.Errorf("pg_stat_statements must be created in schema \"public\"; found in schema \"%s\"", pgssSchema)
 			}
+			if err == nil && extVersion != foundExtVersion && foundExtVersion <= maxKnownExtVersion {
```
uhoh-itsmaciek, line 145: Ah, so if there's a version we don't know about but might work, we just keep our fingers crossed and go on, right?

Separately, I wonder if requiring the latest version for older Postgres versions could be a breaking change here. E.g., if currently we have some customers on say 1.7 for pg13 for whatever reason, but everything works fine because 1.7 is identical to 1.8 in terms of what we care about? I don't know whether this is theoretically even possible (i.e., I don't know if there are released versions of the extension with lower version numbers but still-compatible-for-us APIs), and even if so, whether it's a problem in practice, but maybe something to consider.

uhoh-itsmaciek (approved):
Nice, this looks good and should definitely improve the experience. I have some idle thoughts on compatibility inline, but my gut feeling is it's not a huge problem.

Do you have a sense of how hard it would be to (1) adapt to the existing extension version instead of requiring a specific one (perhaps with a warning to UPDATE) and (2) support a pg_stat_statements installed in other schemas?

(1) would be great to streamline onboarding. The downside is customers may be unaware they should to update the extension, but we could (separately) track this in the snapshot and warn about it in-app. (2) is a nice-to-have—we get occasional support requests for it.

input/postgres/statements.go
lfittl: Hm, that's a good point - it could be better to not error out here unnecessarily if the schema is compatible.

Possibly go with the approach where we decide which pg_stat_statements schema to expect based on the schema version, so we can even work without the extension upgrade?

seanlinsley: Sorry for the delay on this PR. @lfittl this is ready for review. I went ahead and added the schema detection code we discussed.

input/postgres/statements.go
```diff
@@ -117,10 +101,59 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
 				" the monitoring helper functions (https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user)" +
 				" or connect as superuser, to get query statistics for all roles.")
 		}
+
+		var extSchema string
+		var foundExtVersion float32
+		err = db.QueryRow(QueryMarkerSQL+"SELECT nspname, extversion FROM pg_extension pge INNER JOIN pg_namespace pgn ON pge.extnamespace = pgn.oid WHERE pge.extname = 'pg_stat_statements'").Scan(&extSchema, &foundExtVersion)
```
seanlinsley, line 107: IIRC, this will fail if more than one row is returned (that is: the extension is installed more than once). I'm not sure if that's possible; just trying to think through edge cases.

input/postgres/statements.go
```diff
@@ -117,10 +101,59 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
...
+		if err != nil && err != sql.ErrNoRows {
+			return nil, nil, nil, err
+		}
+
+		var extVersion float32 // defined in pg_stat_statements.c
```
lfittl, line 112: Hmm. I wonder a little bit if we can avoid treating the version as a float - and rather just use it as a text column. That appears how Postgres is treating it - i.e. versions are not meant to be compared based on their numeric value.

The branch below could then work like this, with explicitly listed version strings:

```golang
switch foundExtVersion {
case "1.0", "1.1":
  optionalFields = statementSQLDefaultOptionalFields
case "1.2":
  optionalFields = statementSQLpg94OptionalFields
case "1.3", "1.4", "1.5", "1.6", "1.7":
  optionalFields = statementSQLpg95OptionalFields
default: // 1.8 and newer
  optionalFields = statementSQLpg13OptionalFields
}
```

seanlinsley: It's true that `pg_extension.extversion` is a string, but that's likely just to allow third party developers to use whatever version format they want. This extension has always used a float for its version number AFAICT.

Despite that, I'm not opposed to treating them as strings -- if Postgres changes versioning scheme at some point in the future for this extension that would prevent a runtime error here. But then how would we implement the `pg_stat_statements extension outdated` log message if we can't compare them as if they were floats?

input/postgres/statements.go
```diff
@@ -117,10 +101,59 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
 				" the monitoring helper functions (https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user)" +
 				" or connect as superuser, to get query statistics for all roles.")
 		}
+
+		var extSchema string
```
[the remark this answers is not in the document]
seanlinsley: Sure thing, it couldn't hurt to run these checks even if the helper is installed.

input/postgres/statements.go
lfittl: For the outdated message, I think we could just fill in the missing versions, like this:

```golang
if postgresVersion.Numeric >= state.PostgresVersion14 {
  extVersion = "1.9"
} else if postgresVersion.Numeric >= state.PostgresVersion13 {
  extVersion = "1.8"
} else if postgresVersion.Numeric >= state.PostgresVersion12 {
  extVersion = "1.7"
} else if postgresVersion.Numeric >= state.PostgresVersion11 {
  extVersion = "1.6"
} else if postgresVersion.Numeric >= state.PostgresVersion10 {
  extVersion = "1.6"
} else if postgresVersion.Numeric >= state.PostgresVersion96 {
  extVersion = "1.4"
} else if postgresVersion.Numeric >= state.PostgresVersion95 {
  extVersion = "1.3"
} else if postgresVersion.Numeric >= state.PostgresVersion94 {
  extVersion = "1.2"
} else {
  extVersion = "1.1"
}
```

My caution here btw comes from Postgres itself not making any guarantees about this, despite the versioning scheme clearly having a pattern. It seems that when Postgres does version comparisons, it looks at the version upgrade scripts only (e.g. "pg_stat_statements--1.5--1.6.sql" indicates you can upgrade from 1.5 to 1.6) and doesn't consider the numeric sorting to be relevant.

seanlinsley: When I brought up the outdated message, I meant the if condition that decides whether to print it:
```go
if foundExtVersion < extVersion {

Comment (lfittl) [reply input/postgres/statements.go:112]: Ah - but for that I think we can just do an equality comparison, right?

Comment (seanlinsley) [reply input/postgres/statements.go:112]: That would cause us to incorrectly print the message if the `foundExtVersion` is higher than what their version of the collector knows to expect from Postgres. That is: they upgraded Postgres and the extension already, but not the collector.

IMO this isn't worth the effort; it's unlikely Postgres is going to change the version format out of the blue and even if they did we'd likely be releasing a new version of the collector to support new features regardless.

Comment (lfittl) [reply input/postgres/statements.go:112]: Fair - we could limit the check to known Postgres versions, but agreed that this ends up being a whole lot more complicated.

I'm okay with keeping the float check for now, and adjusting if needed.

**Review by lfittl — APPROVED** [review]
:+1: Nicely done!

File: input/postgres/statements.go [hunk 5]
```diff
@@ -83,25 +83,56 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
...
 	}
 
-	usingStatsHelper := false
+	if foundExtVersion >= 1.8 {
+		totalTimeField = statementSQLpg13TotalTimeField
+	} else {
+		totalTimeField = statementSQLDefaultTotalTimeField
+	}
+
+	if foundExtVersion < extVersion {
```
lfittl, line 131: @seanlinsley Oh, one more thing I realize - should we only output this during test runs?

Because right now with this change we'll output it once a minute, which may be a bit frustrating.

seanlinsley: Is there a reason they wouldn't update the extension after first seeing the log message? I'd tend to assume most people only use the test run feature when something's broken. Hiding this message would make it likely most users wouldn't upgrade the extension, causing us to have limited access to new data.

Of course looking farther out, this type of message should be in-app instead of being hidden in the logs.

lfittl: In practice we're not really missing any data at this point (we don't rely on the newer fields), so I think it would actually be okay to not tell people all the time.

I think most people don't look at the logs (or the test output) in regular operations. But if there is a support ticket, we'd then get them to share the test output - thus the info/warning is helpful there, but I don't think it would prompt action just because its output a lot in the regular logs.

We could also push this to next week and get @uhoh-itsmaciek's opinion - not a strong preference from my end.

input/postgres/statements.go
```diff
@@ -83,25 +83,56 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
...
 
-	if postgresVersion.Numeric >= state.PostgresVersion13 {
+	err = db.QueryRow(QueryMarkerSQL+"SELECT nspname, extversion FROM pg_extension pge INNER JOIN pg_namespace pgn ON pge.extnamespace = pgn.oid WHERE pge.extname = 'pg_stat_statements'").Scan(&extSchema, &foundExtVersion)
+	if err != nil && err != sql.ErrNoRows {
+		return nil, nil, nil, err
+	}
+
+	if err == sql.ErrNoRows {
+		logger.PrintInfo("pg_stat_statements does not exist, trying to create extension...")
+		_, err = db.Exec(QueryMarkerSQL + "CREATE EXTENSION IF NOT EXISTS pg_stat_statements SCHEMA public")
```
uhoh-itsmaciek, line 107: This is outside of the scope of the PR, but as a side note, we should stop doing this. It rarely works (since the monitoring user typically doesn't have permissions to create extensions), and when it does, do we really want this behavior as a side effect of running the collector?

uhoh-itsmaciek (approved):
Re-reviewed the updated code: looks good. Thanks for working on this.

input/postgres/statements.go
seanlinsley: Okay, well it's easy to remove. @lfittl thoughts?

seanlinsley: Will go ahead and merge this as-is. We can follow up with that change in a future PR

lfittl: Yeah, lets discuss separately - I remember there being cases where this matters (maybe it was Heroku?) -- essentially when there is no way to create a separate monitoring user, this ensures that you don't have to run an extra step during setup (since you'll have to run the collector with database owner permissions anyway, due to platform limitations)

PR #226, opened 2021-12-14
Fix handling of custom domains with no constraints

uhoh-itsmaciek:
Currently, we always expect domains to have at least one constraint
defined, and collecting custom type information breaks when that's not
the case.

Handle the case of domains with no constraints.


lfittl (approved):
:+1:

PR #233, opened 2021-12-21
Fix crash when reloading empty config file, add integration test

lfittl:
(+ update the integration tests to Go 1.17, and support building them on ARM64)


integration_test/Makefile
```diff
@@ -199,9 +198,31 @@ pg13:
...
+# Note that we reload twice after adding a server to the config, so we can test the
+# "write state file when reload occurs" logic, that requires the server to be there
+# *before* the reload
+reload:
+	docker build -f Dockerfile.test-reload $(DOCKER_BUILD_OPTS)
+	$(call docker_run_cmd,)
+	docker exec --user pganalyze --detach pganalyze-collector-test bash -c "pganalyze-collector 2> /tmp/collector.log"
+	sleep 5
+	docker exec pganalyze-collector-test pganalyze-collector --reload
+	docker exec pganalyze-collector-test cat /tmp/collector.log | grep "Reloading configuration..."
```
uhoh-itsmaciek, line 212: I think this could technically be racy (since I think the signal is processed asynchronously?), but I don't think we don't need to worry about it.

uhoh-itsmaciek (approved):

integration_test/Makefile
lfittl: Yeah, good point - I think its fine, but agree that there is a theoretical race condition here.

PR #234, opened 2021-12-21
Build multi-arch Docker image for x86/amd64 and ARM64

lfittl:
This utilizes Docker's manifest handling to contain both architectures
within the same Docker tag.

In passing improve the Makefile task and rename it to "docker_release"
for clarity.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,10 +41,6 @@ integration_test:
 packages:
 	make -C packages
 
-docker_latest:
-	docker build -t quay.io/pganalyze/collector:latest .
-	docker push quay.io/pganalyze/collector:latest
-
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -45,6 +45,17 @@ docker_latest:
 	docker build -t quay.io/pganalyze/collector:latest .
 	docker push quay.io/pganalyze/collector:latest
```
uhoh-itsmaciek, line 46: The PR description says "rename" but this code is not deleted--is this still useful?

uhoh-itsmaciek (approved):

Makefile
lfittl: Whoops, good point.

PR #239, opened 2022-01-05
Log streams: Support processing primary and secondary lines out of order

lfittl:
This is in particular useful for GCP, where Pub/Sub does not guarantee
that associated lines will be received in order, e.g. when the DETAIL
line comes before the LOG line for log_min_duration_statement output,
causing parameters to be missed.

In passing, fix a bug in logline readiness handling that likely caused
some problems with Heroku logs, as we were considering logs to be "ready"
too often, due to a buggy comparison of the "logLine" loop variable
which we kept a long-lived pointer too (and then compared the next
iteration against that pointer, which would always succeed).

### DIFF
--- a/logs/stream/stream.go
+++ b/logs/stream/stream.go
@@ -44,39 +44,79 @@ import (
 func findReadyLogLines(logLines []state.LogLine, threshold time.Duration) ([]state.LogLine, []state.LogLine) {
 	var readyLogLines []state.LogLine
 	var tooFreshLogLines []state.LogLine
-	var lastPrimaryLogLine *state.LogLine
+	var lastLogLineWithPrefixPid int32 = -1
+	var lastMainLogLinePid int32 = -1
 
 	now := time.Now()
 
 	for _, logLine := range logLines {
+		ready := false
 		if logLine.LogLevel == pganalyze_collector.LogLineInformation_UNKNOWN {
 			if logLine.BackendPid != 0 {
 				// Heroku case (Unknown log lines have PIDs assigned)
-				if lastPrimaryLogLine != nil && logLine.BackendPid == lastPrimaryLogLine.BackendPid {
-					readyLogLines = append(readyLogLines, logLine)
-				} else {
-					tooFreshLogLines = append(tooFreshLogLines, logLine)
-				}
+				ready = lastLogLineWithPrefixPid != -1 && logLine.BackendPid == lastLogLineWithPrefixPid
 			} else {
-				if lastPrimaryLogLine != nil {
-					// Self-managed and Google Cloud SQL case - we always stitch unknown log levels to the prior line
-					readyLogLines = append(readyLogLines, logLine)
-				} else if now.Sub(logLine.CollectedAt) <= threshold {
-					tooFreshLogLines = append(tooFreshLogLines, logLine)
+				// Self-managed case (subsequent multi-line lines do not have PIDs)
+				if lastLogLineWithPrefixPid != -1 {
+					// We always stitch unknown log levels to the prior line
+					ready = true
+				} else if now.Sub(logLine.CollectedAt) > threshold {
+					// This is ready, but we can't associate it to anything within the threshold - throw it away
+					continue
 				}
 			}
 		} else {
 			if now.Sub(logLine.CollectedAt) > threshold {
-				readyLogLines = append(readyLogLines, logLine)
-				lastPrimaryLogLine = &logLine
+				ready = true
+				lastLogLineWithPrefixPid = logLine.BackendPid
+				if isPostgresErrorLevel(logLine.LogLevel) {
+					lastMainLogLinePid = logLine.BackendPid
+				} else {
+					lastMainLogLinePid = -1
+				}
+			} else if isAdditionalLineLevel(logLine.LogLevel) && logLine.BackendPid != 0 && lastMainLogLinePid != -1 && logLine.BackendPid == lastMainLogLinePid {
+				ready = true
+				lastLogLineWithPrefixPid = logLine.BackendPid
 			} else {
-				tooFreshLogLines = append(tooFreshLogLines, logLine)
-				lastPrimaryLogLine = nil
+				lastLogLineWithPrefixPid = -1
+				lastMainLogLinePid = -1
 			}
+		}
+		if ready {
+			readyLogLines = append(readyLogLines, logLine)
+		} else {
+			tooFreshLogLines = append(tooFreshLogLines, logLine)
 		}
 	}
 
 	return readyLogLines, tooFreshLogLines
+}
+
+/* Level lists current as of Postgres 14.1 */
+func isPostgresErrorLevel(str pganalyze_collector.LogLineInformation_LogLevel) bool {
+	switch str {
+	case pganalyze_collector.LogLineInformation_DEBUG,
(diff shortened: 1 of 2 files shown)


logs/stream/stream.go
```diff
@@ -44,41 +44,81 @@ import (
...
 	var tooFreshLogLines []state.LogLine
-	var lastPrimaryLogLine *state.LogLine
+	var lastLogLineWithPrefixPid int32 = -1
+	var lastMainLogLinePid int32 = -1
 
 	now := time.Now()
 
 	for _, logLine := range logLines {
+		ready := false
 		if logLine.LogLevel == pganalyze_collector.LogLineInformation_UNKNOWN {
```
uhoh-itsmaciek, line 54: Maybe this is overthinking it, but I wonder if, for readability, we shouldn't invert the condition here and handle known-LogLevel lines first? What happens in the first branch depends on what happened to related lines in the second branch, so maybe having that other branch come first makes more sense?

logs/stream/stream.go
```diff
@@ -44,41 +44,81 @@ import (
...
+		pganalyze_collector.LogLineInformation_WARNING,
+		pganalyze_collector.LogLineInformation_ERROR,
+		pganalyze_collector.LogLineInformation_LOG,
+		pganalyze_collector.LogLineInformation_FATAL,
+		pganalyze_collector.LogLineInformation_PANIC:
+		return true
+	}
+	return false
+}
+func isAdditionalLineLevel(str pganalyze_collector.LogLineInformation_LogLevel) bool {
```
uhoh-itsmaciek, line 110: These helpers are great, but the naming is a little inconsistent (the first name includes "Postgres" and the second doesn't, the first mentions errors even though several log levels are not errors, and the second includes "Line" but the first doesn't). Maybe "isMainLogLevel" and "isExtraLogLevel"?

logs/stream/stream.go
```diff
@@ -44,41 +44,81 @@ import (
...
+				} else {
+					lastMainLogLinePid = -1
+				}
+			} else if isAdditionalLineLevel(logLine.LogLevel) && logLine.BackendPid != 0 && lastMainLogLinePid != -1 && logLine.BackendPid == lastMainLogLinePid {
+				ready = true
+				lastLogLineWithPrefixPid = logLine.BackendPid
 			} else {
-				tooFreshLogLines = append(tooFreshLogLines, logLine)
-				lastPrimaryLogLine = nil
+				lastLogLineWithPrefixPid = -1
```
uhoh-itsmaciek, line 81: Might be worth commenting this branch? This is the case for a too-fresh main line, right?

uhoh-itsmaciek (commented):
Looks great. Sorry for the nit-picks, but I found the log handling code hard to follow (looking forward to reviewing the refactoring patch), so I think we should make an effort for readability here.

logs/stream/stream.go
```diff
@@ -44,41 +44,81 @@ import (
...
+		ready := false
 		if logLine.LogLevel == pganalyze_collector.LogLineInformation_UNKNOWN {
 			if logLine.BackendPid != 0 {
 				// Heroku case (Unknown log lines have PIDs assigned)
-				if lastPrimaryLogLine != nil && logLine.BackendPid == lastPrimaryLogLine.BackendPid {
-					readyLogLines = append(readyLogLines, logLine)
-				} else {
-					tooFreshLogLines = append(tooFreshLogLines, logLine)
-				}
+				ready = lastLogLineWithPrefixPid != -1 && logLine.BackendPid == lastLogLineWithPrefixPid
```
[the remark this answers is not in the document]
lfittl: Yeah, I was debating with myself about that - after a good night's sleep (or at least some sleep 😬) it seems reasonable to not have that check.

logs/stream/stream.go
```diff
@@ -44,41 +44,81 @@ import (
...
-					// Self-managed and Google Cloud SQL case - we always stitch unknown log levels to the prior line
-					readyLogLines = append(readyLogLines, logLine)
-				} else if now.Sub(logLine.CollectedAt) <= threshold {
-					tooFreshLogLines = append(tooFreshLogLines, logLine)
+				// Self-managed case (subsequent multi-line lines do not have PIDs)
+				if lastLogLineWithPrefixPid != -1 {
+					// We always stitch unknown log levels to the prior line
+					ready = true
+				} else if now.Sub(logLine.CollectedAt) > threshold {
+					// This is ready, but we can't associate it to anything within the threshold - throw it away
```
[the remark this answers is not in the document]
lfittl: Yeah, good point - I reworded to "We can't associate it to anything within the threshold, and its unlikely we ever will - throw it away"

logs/stream/stream.go
lfittl: Yeah, good point - adding a comment.

lfittl:
@uhoh-itsmaciek Ok, I think I found a way that this becomes much more readable - thank you for nudging me to attempt a better way of structuring this :)

logs/stream/stream.go
```diff
@@ -37,48 +38,131 @@ import (
...
+
+	// Part of a multi-line log line, where we don't have a log_line_prefix,
+	// or any other contextual information!
+	//
+	// This is the case for self-managed servers where we tail a log file,
+	// and so we can reasonably assume that subsequent unidentifiable lines
+	// directly belong to the line we saw right before.
+	//
+	// There are edge cases where we discard lines here, since we want to avoid
+	// keeping lines forever that we could never associate to anything.
```
uhoh-itsmaciek, line 86: I think this is out of scope for this PR, but could self-hosted log output like the following get us in trouble:

```
[1]: backend 2: some long line with
[2]:    newlines so this line is a continuation
[3]: backend 1: another long line with
[4]:    newlines so this line is also a continuation
```

Once we sort this, we'd get

```
[3]: backend 1: another long line with
[1]: backend 2: some long line with
[2]:    newlines so this line is a continuation
[4]:    newlines so this line is also a continuation
```

right? And here we'd lose line [4]? Or am I misunderstanding how this works?

logs/stream/stream.go
```diff
@@ -37,48 +38,131 @@ import (
...
+			return LogLineReady
+		}
+
+		return LogLineNoAction
+	}
+
+	// Part of a multi-line log line, where we don't have a log_line_prefix,
+	// but we have a PID, and so can make some assumptions of where this
+	// log line might fit. This is the case for Heroku and syslog receivers.
+	if logLine.BackendPid != 0 {
```
uhoh-itsmaciek, line 70: If we have InvalidPid as a constant, we could add UnknownPid as well.

logs/stream/stream.go
```diff
@@ -37,48 +38,131 @@ import (
...
+// - Always has the log line number, allowing association of related log lines
+// - Multi-line messages are already combined together
+//   (as of Sept 14, 2021 - see https://cloud.google.com/sql/docs/release-notes#September_14_2021)
+
+const InvalidPid int32 = -1
+
+type LogLineReadiness int
+
+const (
+	LogLineNoAction LogLineReadiness = iota
```
uhoh-itsmaciek, line 50: This is more like LogLineDefer, right?

uhoh-itsmaciek (approved):
Nice, I like breaking out the state analysis into a helper. A couple of additional comments, but feel free to ignore.

PR #240, opened 2022-01-05
Refactor log runner, use consistent 10s interval for streamed logs

lfittl:
This is a major refactoring to reduce a lot of duplicate Go routine
code in the log streaming logic, and centralizes all top-level log
handling in the runner package, which then calls input/logs/output
packages as needed.

Additionally, change the interval for streamed logs from (at most) once
every 3 seconds to every 10 seconds. Its unnecessary to send more often
and can cause tiny log file records to be created that put unnecessary
stress on the system.

For consistency, drop the scheduler logic use for log downloads, and
use a simple time.Ticker to schedule log downloads every 30 seconds, like
before.


input/system/google_cloudsql/logs.go
```diff
@@ -127,3 +138,46 @@ func SetupLogSubscriber(ctx context.Context, wg *sync.WaitGroup, globalCollectio
...
+				logLine, ok := logs.ParseLogLineWithPrefix("", in.Content+"\n")
+				if !ok {
+					logger.PrintError("Can't parse log line: \"%s\"", in.Content)
+					continue
+				}
+				logLine.OccurredAt = in.OccurredAt
+
+				// Ignore loglines which are outside our time window
+				nullTime := time.Time{}
+				if logLine.OccurredAt != nullTime && logLine.OccurredAt.Before(linesNewerThan) {
```
uhoh-itsmaciek, line 171: Same as above, we should probably use `IsZero` here instead.

input/system/heroku/logs.go
```diff
@@ -186,55 +183,63 @@ func logStreamItemToLogLine(item HerokuLogStreamItem, servers []*state.Server, n
...
+func setupLogTransformer(ctx context.Context, wg *sync.WaitGroup, servers []*state.Server, in <-chan HerokuLogStreamItem, out chan state.ParsedLogStreamItem, globalCollectionOpts state.CollectionOpts, logger *util.Logger) {
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		var logLinesByName map[string][]state.LogLine
+		var nameToServer map[string]*state.Server
+
+		logLinesByName = make(map[string][]state.LogLine)
+		nameToServer = make(map[string]*state.Server)
```
uhoh-itsmaciek, line 208: NBD, but maybe 'source' instead of 'name' in both of these?

main.go
```diff
@@ -92,7 +88,7 @@ func run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
 		if config.SystemType == "google_cloudsql" {
 			hasAnyGoogleCloudSQL = true
 		}
-		if config.SystemType == "heroku" {
+		if config.SystemType == "heroku" && os.Getenv("DYNO") != "" && os.Getenv("PORT") != "" {
```
uhoh-itsmaciek, line 91: ? (not objecting, just curious about the change)

runner/logs.go
```diff
@@ -1,57 +1,108 @@
...
+
+					if server.Config.AwsDbInstanceID == "" {
+						continue
+					}
+
+					innerWg.Add(1)
+					go downloadLogsForServerWithLocksAndCallbacks(&innerWg, server, globalCollectionOpts, logger)
+				}
+
+				innerWg.Wait()
```
uhoh-itsmaciek, line 99: This lets us avoid downloading new log files if previous log files are still being processed, right? Nice.

runner/logs.go
```diff
@@ -82,6 +133,157 @@ func downloadLogsFromOneServer(wg *sync.WaitGroup, server *state.Server, globalC
...
+				return
+			case t := <-ticker.C:
+				for identifier := range logLinesByServer {
+					if len(logLinesByServer[identifier]) > 0 {
+						server := &state.Server{}
+						for _, s := range servers {
+							if s.Config.Identifier == identifier {
+								server = s
+							}
+						}
```
uhoh-itsmaciek, line 181: NBD but since this is just mechanical, we could extract this into a helper to take the clutter out of the function body here:

```golang
func findServerById(servers []*state.Server, id string) {
  for _, s := range servers {
    if s.Config.Identifier == identifier {
      return s
    }
  }
}
```

runner/logs.go
```diff
@@ -111,73 +313,126 @@ func TestLogsForAllServers(servers []*state.Server, globalCollectionOpts state.C
...
-			continue
-		}
+	logs.EmitTestLogMsg(server, globalCollectionOpts, logger)
 
-		if server.Config.AwsDbInstanceID == "" {
-			continue
-		}
+	select {
+	case <-logTestSucceeded:
+		break // TODO: Is this correct?
```
uhoh-itsmaciek, line 366: ?

runner/logs.go
```diff
@@ -111,73 +313,126 @@ func TestLogsForAllServers(servers []*state.Server, globalCollectionOpts state.C
...
 
-	wg.Wait()
+	logs.EmitTestLogMsg(server, globalCollectionOpts, logger)
 
-	return
+	select {
+	case <-logTestSucceeded:
+		break // TODO: Is this correct?
+	case <-time.After(10 * time.Second):
+		logger.PrintError("ERROR - Azure Event Hub log tail timed out after 10 seconds - did not receive any log events")
```
uhoh-itsmaciek, line 406: I guess with Azure and GCP log tests, we still have the issue of potentially competing with a running collector on the subscription, right? Should we indicate that in the message here?

uhoh-itsmaciek (approved):
Nice, this is awesome. A bunch of smaller comments inline, but I think this improves things a lot.

input/system/azure/logs.go
```diff
@@ -140,3 +173,47 @@ func SetupLogSubscriber(ctx context.Context, wg *sync.WaitGroup, globalCollectio
...
+				if !ok {
+					logger.PrintError("Can't parse log line: \"%s\"", logLineContent)
+					continue
+				}
+				logLine.CollectedAt = time.Now()
+				logLine.UUID = uuid.NewV4()
+
+				// Ignore loglines which are outside our time window (except in test runs)
+				nullTime := time.Time{}
+				if logLine.OccurredAt != nullTime && logLine.OccurredAt.Before(linesNewerThan) && !globalCollectionOpts.TestRun {
```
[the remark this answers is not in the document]
lfittl: Oh, right, I forgot `IsZero()` is a thing :)

main.go
lfittl: We were previously making the same check further down - essentially it tries to protect against situation where we incorrectly identify something as being a Heroku system type.

That said, we don't actually need to verify that, because we have the same check already in https://github.com/pganalyze/collector/blob/main/config/read.go#L473 -- so I'll drop this.

runner/logs.go
lfittl: Yeah, right now we always finish the work for all servers before scheduling the work again (the same is done for other snapshot data). This was already in place before for log downloads as well, FWIW.

Not sure if its the right decision, as it could lead to a slow server causing all other servers on a collector to not have collections at regular intervals. But for now this is consistent this way, and we may re-evaluate this at some point.

runner/logs.go
lfittl: Ah, thanks for the reminder, I forgot that failure scenario.

I'll add a hint message like the following: "HINT - This error may be a false positive if the collector is running in the background and consumes the same [provider name] stream" -- but we probably want to find a better way to do this in general.

*thinks about this briefly*

We could signal the collector running in the background (like we do for a reload) to not acknowledge the GCP Pub Sub messages - that way they should still be available for the test collector to be retrieved. e.g. when the collector receives a `SIGUSR1` signal it doesn't acknowledge any messages for the next 10 seconds. We'd have to test whether the acknowledge handling actually behaves that way, but that could work (at least for GCP where there is an explicit ack from the client).

runner/logs.go
lfittl: Good catch, that was a left over - I was trying to reason whether its safe to have `select` here without a `for` loop, but the answer is yes, this code is fine as is (see e.g. https://stackoverflow.com/questions/39212333/how-can-i-use-time-after-and-default-in-golang)

runner/logs.go
uhoh-itsmaciek: Ah interesting—I was thinking about signaling, too, but didn't think of using that to not ack. That could work, although maybe that would lead to duplicate messages? Maybe instead we could use the signal to somehow share the received messages (e.g., an ephemeral localhost web server?). But yeah, for now, I think just a HINT in the message works.

PR #242, opened 2022-01-14
Log processing: Handle non-UTC/non-local log_timezone values correctly

lfittl:
Previously we supported handling timezone values in three cases:

(1) No timezone specified
(2) UTC/GMT timezone explicitly specified
(3) Timezone specified that matches the collector system timezone

However we did not handle a log_timezone setting that is different from
the timezone thats configured on the system the collector is running on.

This was caused by incorrect assumptions of what time.Parse() does in
this case. As we have no other way to get the UTC timestamp of a logline,
we rely on the system's timezone database to help us parse the log line
in the correct timezone for this edge case.

### DIFF
--- a/logs/parse.go
+++ b/logs/parse.go
@@ -401,10 +401,28 @@ func ParseLogLineWithPrefix(prefix string, line string) (logLine state.LogLine, 
 		logLine.OccurredAt, err = time.Parse(timeFormat, timePart)
 		if err != nil {
 			if timeFormatAlt != "" {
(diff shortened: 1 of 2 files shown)


uhoh-itsmaciek (approved):
When we format these for log snapshots, is that done as UTC, or do we need to audit time zone handling on the receiving end as well?

lfittl:
> When we format these for log snapshots, is that done as UTC, or do we need to audit time zone handling on the receiving end as well?

Thats all done based on the UNIX timestamp, to my understanding.

PR #244, opened 2022-01-22
Fix fingerprinting of unparsable queries

lfittl:
```
Avoid generating unique fingerprints for truncated/unparsable queries 

This provides little value, and causes a lot of unnecessary unique query
records to be stored by pganalyze. Match the logic to how NormalizeQuery
treats these query texts, and support the old handling when the config
setting "filter_query_text" is set to "none".

In passing, move all special query texts into a new central util file,
for easier overview and explanation.
```

```
Normalization: Attempt auto-fixing truncated queries 

This adjusts the normalization logic to use the same "fix unparsable
queries" logic that already existed for fingerprinting, and is mainly
intended for data in pg_stat_activity (that is often truncated).

This logic provides a better level of grouping for these truncated
queries.

Add the new "/* truncated-query */ " magic comment to make it clear to the
viewer that they are looking at a fixed truncated query text, and not
the full query text as sent by the application.
```


output/transform/transform_test.go
```diff
@@ -22,10 +22,10 @@ func TestStatements(t *testing.T) {
 
 	q1 := "SELECT 1"
 	q2 := "SELECT * FROM test"
-	fp1 := util.FingerprintQuery(q1)
+	fp1 := util.FingerprintQuery(q1, "none", -1)
```
uhoh-itsmaciek, line 25: It might be nice to avoid magic strings here, but that's an existing problem so we don't need to do that in this patch.

uhoh-itsmaciek (approved):

seanlinsley (approved):
This looks to be working as expected on staging

output/transform/transform_test.go
lfittl: Agree, I'll not do this here for now, but good cleanup to do some time.

PR #246, opened 2022-01-25
Handle booleans in config more consistently

### LINKED ISSUE #245: true:false vs 1:0
Hi,
it is confusing to use in documentation value `true` if `false` doesn't represent opposite.
See here:
https://pganalyze.com/docs/explain/setup/log_explain/02_enable_log_explain

While setting `false` doesn't disable the feature because it is not "" nor `0`:
https://github.com/pganalyze/collector/blob/main/config/read.go#L62

Note that we are using kubernetes yaml, where we cannot use boolean values but only "true"/"false" as strings.

uhoh-itsmaciek:
Right now, we parse boolean values in our config files with go-ini,
but environment-based config goes through some ad-hoc parsing. For
most boolean values, any value other than the empty string and '0' is
considered true (for the DB_ALL_NAMES setting, any value except '1' is
considered false).

Unify this to treat case-insensitive false, off, no, 'f', and 'n' as
false in addition to zero. All other values are still considered true
to preserve backward compatibility (this breaks backward compatibility
for DB_ALL_NAMES, but that setting is unlikely to be set when
disabling it, and the unified behavior is worth the change).

This is closer to what go-ini does for parsing our config files [1],
and to the Postgres config file handling [2].

Fixes #245

[1]: https://ini.unknwon.io/docs/howto/work_with_values
[2]: https://www.postgresql.org/docs/current/config-setting.html

### DIFF
--- a/config/read.go
+++ b/config/read.go
@@ -23,6 +23,14 @@ import (
 
 const DefaultAPIBaseURL = "https://api.pganalyze.com"
 
+func parseConfigBool(value string) bool {
+	if value == "" || value == "0" || value == "off" || value == "false" || value == "no" || value == "f" || value == "n" {
+		return false
+	}
+
+	return true
+}
+
 func getDefaultConfig() *ServerConfig {
 	config := &ServerConfig{
 		APIBaseURL:              DefaultAPIBaseURL,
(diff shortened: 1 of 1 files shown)


danielkucera:
The patch looks good. Thank you.

config/read.go
```diff
@@ -23,6 +23,14 @@ import (
 
 const DefaultAPIBaseURL = "https://api.pganalyze.com"
 
+func parseConfigBool(value string) bool {
+	if value == "" || value == "0" || value == "off" || value == "false" || value == "no" || value == "f" || value == "n" {
```
lfittl, line 27: Worth doing this comparison against the lower-cased version of string? (both go-ini and Postgres seem to handle this case-insensitively)

lfittl (approved):
:+1:, one minor comment but overall looks good.

config/read.go
uhoh-itsmaciek: Yes! I had that change locally (the commit message even mentions it), but neglected to push it. Updated.

PR #251, opened 2022-02-10
Ensure temp files are cleaned up properly for RDS log processing

uhoh-itsmaciek:
You can see this minimized in https://go.dev/play/p/pZo9NWTiN2n versus https://go.dev/play/p/Z2ODTVDTgSR . The problem is that we copy the struct at the site of the defer. When we later replace the LogFiles slice, the copy captured by the defer is not updated, so when we run Cleanup, it cleans up only the empty slice in the captured copy. Adding a wrapper function means the struct is only copied when the deferred function is executed.

### DIFF
--- a/runner/logs.go
+++ b/runner/logs.go
@@ -140,7 +140,9 @@ func downloadLogsForServer(server *state.Server, globalCollectionOpts state.Coll
 	}
 
 	transientLogState := state.TransientLogState{CollectedAt: time.Now()}
-	defer transientLogState.Cleanup()
+	defer func() {
+		transientLogState.Cleanup()
+	}()
 
 	var newLogState state.PersistedLogState
 	newLogState, transientLogState.LogFiles, transientLogState.QuerySamples, err = system.DownloadLogFiles(server.LogPrevState, server.Config, logger)
--- a/runner/logs.go
+++ b/runner/logs.go
@@ -140,9 +140,7 @@ func downloadLogsForServer(server *state.Server, globalCollectionOpts state.Coll
 	}
 
 	transientLogState := state.TransientLogState{CollectedAt: time.Now()}
-	defer func() {
-		transientLogState.Cleanup()
-	}()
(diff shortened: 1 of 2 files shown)


runner/logs.go
```diff
@@ -140,7 +140,9 @@ func downloadLogsForServer(server *state.Server, globalCollectionOpts state.Coll
 	}
 
 	transientLogState := state.TransientLogState{CollectedAt: time.Now()}
-	defer transientLogState.Cleanup()
+	defer func() {
```
lfittl, line 143: I wonder if we should add a note here, that's saying something like:

```golang
// We wrap the function call in an anonymous function to avoid a copy of the stack-allocated
// struct occurring at this point - causing the log file list pointed to by the state to be incorrect
// (as it gets replaced in a subsequent function call, causing the pointer to change)
```

(I've also briefly considered whether we can fix this by using a pointer, but that seems more complicated?)

lfittl (approved):

lfittl:
> Given that this issue may have left a bunch of temp files lying around, I wonder if we should try to just delete all temp files created by the pganalyze user on collector startup? That would not play nice with multiple collectors running on the same system, but that's not really a supported use case anyway.

Yeah, good question. I think the problem might be that we can't identify which are our temporary files, vs those of other programs. In the case of log files, the temp files just contain the log content, so I don't think there is any obvious indicator...  except maybe the user that owns it, in the case of package installed collectors? (for Docker this shouldn't matter anyway, since there the container restarts on an upgrade, clearing the temp files)

Instead of solving this inside the collector code, we could solve it in the packaging scripts on upgrade, as there we have a reasonable assumption that the pganalyze user was created (as its maintained by the packaging logic).

runner/logs.go
uhoh-itsmaciek: Updated the code to fix this with pointer receivers instead. Confirmed by running an instrumented build on our staging collector.

PR #252, opened 2022-02-15
Release 0.42.2

uhoh-itsmaciek:


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.42.2      2022-02-15
+
+* Fix cleanup of temporary files used when processing logs
```
lfittl, line 5: Do we want users to take manual action to clean up historic log files? (if so, we should probably note that here)

Alternatively, if we want to handle it in the packaging, we should get a PR in that does that before making the release (I think that needs changes in https://github.com/pganalyze/collector/blob/main/packages/src/rpm-systemd/post.sh and https://github.com/pganalyze/collector/blob/main/packages/src/deb-systemd/after-install.sh)

uhoh-itsmaciek: Good point. Thinking about this again, I think it's probably safest not to try to be too clever here, especially since these files will be in the temp directory. I'll add a note on manual cleanup.

lfittl (approved):
:+1:

PR #253, opened 2022-02-19
Filter out vacuum records we cannot match to a table name

uhoh-itsmaciek:
Currently we ensure that we have a v.relid, but that does not
guarantee that we match an entry in pg_class or in our
pg_stat_activity query.

If we have no table name to report (e.g., because of a manual VACUUM
running in a different database), omit the record from the vacuum
progress output.

Alternately, we could try to match manual vacuums in pg_stat_activity
in addition to autovacuums--is there a reason we're not doing that?

### DIFF
--- a/input/postgres/vacuum_progress.go
+++ b/input/postgres/vacuum_progress.go
@@ -80,7 +80,7 @@ SELECT (query_start_epoch || padded_pid)::bigint AS vacuum_identity,
 			 JOIN activity a USING (pid)
 			 LEFT JOIN pg_catalog.pg_class c ON (c.oid = v.relid)
 			 LEFT JOIN pg_catalog.pg_namespace n ON (n.oid = c.relnamespace)
- WHERE v.relid IS NOT NULL OR (a.query <> '<insufficient privilege>' AND a.nspname IS NOT NULL AND a.relname IS NOT NULL)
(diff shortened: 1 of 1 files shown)


lfittl (approved):
:+1:, that makes sense (had to re-read that OR condition a few times to make sure we're not making the LEFT JOIN useless, but its correct)

For manual vacuums, we could try to match them, but the challenge is that they may not follow a certain standard pattern - e.g. someone might have a query comment in there, specify multiple tables at the same time, or just have `VACUUM` to capture the whole database.

AFAIK, the pg_stat_activity matching logic should only apply when we don't have permission to read `pg_stat_progress_vacuum.relid`, because otherwise we can directly join with pg_class/pg_namespace - I wonder how often thats actually a problem these days, with the existence of `pg_monitor` on most providers.

uhoh-itsmaciek:
@lfittl wouldn't a vacuum in a different database also cause a problem? pg_stat_progress_vacuum is a system-wide view, but pg_class is database-local.

lfittl:
> @lfittl wouldn't a vacuum in a different database also cause a problem? pg_stat_progress_vacuum is a system-wide view, but pg_class is database-local.

(took me a while to check back here, oops!)

That's an excellent point - I forgot about pg_class being database-local when reviewing this.

I think the change as is here makes sense, and +1 on merging this.

Issue #255, opened 2022-03-18
Citus relation stats: don't error out when shard_replication_factor > 1


seanlinsley:
`citus_table_size()` errors out if `shard_replication_factor > 1` https://docs.citusdata.com/en/v10.2/admin_guide/table_management.html#table-size

To avoid dropping the stats for all tables, we should ignore this error. One way to do that would be to add a function in the `pganalyze` schema that calls the original Citus function wrapped in a try/catch block.

PR #256, opened 2022-03-18
Check citus.shard_replication_factor before querying citus_table_size

uhoh-itsmaciek:
The citus_table_size function errors out if
citus.shard_replication_factor is greater than 1. Since checking this
is part of the extended stats collection, a shard_replication_factor
greater than 1 means schema collection will fail.

Check the citus.shard_replication_factor parameter before invoking the
function. Fall back to 0 since the field is nut nullable in our
snapshot.

### DIFF
--- a/input/postgres/relation_stats_ext.go
+++ b/input/postgres/relation_stats_ext.go
@@ -9,7 +9,11 @@ import (
 
 const citusRelationSizeSQL = `
 SELECT logicalrelid::oid,
-       pg_catalog.citus_table_size(logicalrelid)
+			 CASE
+			   WHEN coalesce(current_setting('citus.shard_replication_factor')::integer, 1) = 1
+				 THEN pg_catalog.citus_table_size(logicalrelid)
+				 ELSE 0
+			 END AS citus_table_size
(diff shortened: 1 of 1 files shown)


lfittl (approved):
:+1:

seanlinsley (commented):
I tried to reproduce the error locally, but it would still succeed even after increasing the replication factor. Maybe that's just because I'm using their example Docker image which doesn't have any replication? (trying to call [replicate_table_shards](https://docs.citusdata.com/en/v10.2/develop/api_udf.html#replicate-table-shards) fails)

uhoh-itsmaciek:
@seanlinsley hmm, interesting. Thanks for trying that out, but not sure where to go from there. What we have right now is clearly broken in some cases, although apparently maybe not as many as we think. Basically, the downside of the fix is that it might report the table size as 0 in more situations than necessary, but I think that's better than failing schema collection. Maybe we can go with this for now, and follow up with an additional patch if we get reports of unexpected behavior. Thoughts?

seanlinsley:
agree, it should be fine to merge as-is and then update in the future if we get more information on the failure case

seanlinsley:
closes #255

PR #262, opened 2022-04-19
logging: add option for structured JSON logs

### LINKED ISSUE #261: RFC: Structured logging
Hello, I'd like to have the pganalyze collector emit structured logs. The reason I want this is that occasionally the collector will fail to parse a query plan. Such query plans tend to be thousand-line monstrosities and the default logging agent for GKE dutifully logs each and every line.

image
 
As I see it, the general options are:

1. Add a flag like `--json-logging` that sets the logging destination to one of the many structured json loggers on the market.
2. Do nothing. I imagine structured logging is a reasonably common thing these days, so hopefully I'm not the only one that wants it.

My favorite is option 1 and I'm happy to send a PR your way. Just tell me your preferred logging library. Alternately, the format is simple enough that we could probably just use a basic struct and the stdlib json encoder if you want to avoid a dependency.

jschaf:
Add support for emitting logs as newline-delimited JSON with a bit of
structure matching the GCP stackdriver conventions [1]. I didn't use a library
like zerolog or zap because it seemed way overkill since there's already
a logging facade with util.logger, and we just need a way to dump output
to stderr.

[1]: https://cloud.google.com/logging/docs/structured-logging

Output looks like:

```json
{"severity":"INFO","message":"Running collector test with pganalyze-collector 0.43.0","time":"2022-04-19T12:31:05.100489-07:00"}
```

Future work could include:

- Using a more allocation efficient JSON encoder in place of Go's standard JSON
  library. The format is simple enough that it's tempting to roll our own.

- Capturing more structured data like source-code location.

Fixes #261.

### DIFF
--- a/main.go
+++ b/main.go
@@ -238,6 +238,7 @@ func main() {
 	var writeHeapProfile bool
 	var testRunAndTrace bool
 	var logToSyslog bool
+	var logToJSON bool
 	var logNoTimestamps bool
 	var reloadRun bool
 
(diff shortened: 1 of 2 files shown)


lfittl (approved):
Excellent, thanks for the contribution!

jschaf:
Anything more need to merge?

lfittl:
@jschaf Done - thanks again!

FWIW, we'll probably make a new collector release either this or next week, where this will be included.

PR #264, opened 2022-04-29
Log database name in column-stats-helper-missing warning

uhoh-itsmaciek:
This helper must be installed in each database separately, but the
current warning does not say where the problem is.

Include the database name in the warning.

### DIFF
--- a/input/postgres/relation_stats.go
+++ b/input/postgres/relation_stats.go
@@ -182,9 +182,11 @@ func GetColumnStats(logger *util.Logger, db *sql.DB, globalCollectionOpts state.
 		sourceTable = "pganalyze.get_column_stats()"
 	} else {
 		if systemType != "heroku" && !connectedAsSuperUser(db, systemType) && globalCollectionOpts.TestRun {
-			logger.PrintInfo("Warning: Limited access to table column statistics detected. Please setup" +
-				" the monitoring helper function pganalyze.get_column_stats (https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user)" +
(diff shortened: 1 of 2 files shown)


input/postgres/relation_stats.go
```diff
@@ -182,9 +182,11 @@ func GetColumnStats(logger *util.Logger, db *sql.DB, globalCollectionOpts state.
 		sourceTable = "pganalyze.get_column_stats()"
 	} else {
 		if systemType != "heroku" && !connectedAsSuperUser(db, systemType) && globalCollectionOpts.TestRun {
-			logger.PrintInfo("Warning: Limited access to table column statistics detected. Please setup" +
-				" the monitoring helper function pganalyze.get_column_stats (https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user)" +
-				" or connect as superuser, to get column statistics for all tables.")
+			var currentDb string
+			db.QueryRow(QueryMarkerSQL + "SELECT current_database()").Scan(&currentDb)
```
lfittl, line 186: This works, but we could also just pass the database name down from the original call that knows which `dbName` we are on, in `collectOneSchema` (which calls `collectSchemaData`, which calls `GetColumnStats`) -- since this might be useful for other error reporting in the future.

uhoh-itsmaciek: Yeah, I wasn't sure which was the lesser evil--the argument list is pretty long as is. It might be useful to refactor to pass more context around more easily, but I didn't want to do that here. I'll refactor to pass the name down as you suggest.

lfittl (approved):
:+1:

PR #265, opened 2022-04-30
Log processing: Add support for Azure DB for Postgres - Flexible Server

lfittl:
There are some differences in how Flexible Server sends data into the
Azure Event Hub as compared to Single Server, causing log lines to not
be processed. This attempts to parse lines for Flexible Server first,
and automatically falls back to the old special handling for
Single Server if that fails.

### DIFF
--- a/runner/logs.go
+++ b/runner/logs.go
@@ -3,6 +3,7 @@ package runner
 import (
 	"context"
 	"io/ioutil"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


runner/logs.go
```diff
@@ -408,6 +409,9 @@ func testAzureLogStream(ctx context.Context, wg *sync.WaitGroup, server *state.S
 	err := azure.SetupLogSubscriber(ctx, wg, globalCollectionOpts, logger, []*state.Server{server}, parsedLogStream)
 	if err != nil {
 		logger.PrintError("ERROR - Could not get logs through Azure Event Hub: %s", err)
+		if strings.HasPrefix(err.Error(), "failed to configure Azure AD JWT provider: failed to refersh token") { // Note that "refersh" is a typo in the Azure SDK logic
```
uhoh-itsmaciek, line 412: Should we maybe future-proof against them fixing this?

uhoh-itsmaciek (approved):

runner/logs.go
lfittl: Yeah, maybe I'll just skip the last two words in that error message, that way if they fix the typo, it'll still work.

PR #266, opened 2022-05-02
Postgres 14: Add support for new autovacuum and autoanalyze log event syntax

lfittl:


logs/analyze_test.go
```diff
@@ -1430,6 +1430,112 @@ var tests = []testpair{
...
+		[]state.LogLine{{
+			Content: "automatic vacuum of table \"mydb.myschema.mytable\": index scans: 0\n" +
+				"	pages: 0 removed, 14761 remain, 0 skipped due to pins, 12461 skipped frozen\n" +
+				"	tuples: 0 removed, 122038 remain, 14433 are dead but not yet removable, oldest xmin: 538040633\n" +
+				"	index scan bypassed: 255 pages from table (1.73% of total) have 661 dead item identifiers\n" +
+				"	I/O timings: read: 0.000 ms, write: 0.000 ms\n" +
+				"	avg read rate: 0.000 MB/s, avg write rate: 0.000 MB/s\n" +
+				"	buffer usage: 4420 hits, 0 misses, 0 dirtied\n" +
+				"	WAL usage: 1 records, 0 full page images, 245 bytes\n" +
+				"	system usage: CPU: user: 0.00 s, system: 0.00 s, elapsed: 0.01 s",
```
uhoh-itsmaciek, line 1443: Given the large number of capture groups we're working, should we have distinct values for the various captured values to ensure the test is validating we're referencing the correct groups? I think we should also consider named captured groups here, but we can save that for another day.

uhoh-itsmaciek (approved):

logs/analyze_test.go
lfittl: Yeah, fair point, I'll adjust the values to make sure the test captures it all.

And agree re: named capture groups, I was initially concerned about performance when using them, but I think the regexp is the bigger performance impact here anyway (and we do prefix matching in the log parser first, before we check the regexp anyway, so unless its a high recurring event like connections, its unlikely that using named captures would be a problem).

uhoh-itsmaciek: Hmm, yeah. If perf does end up being a problem, we could compromise on just constants referencing the indexes of the capture groups--that has no perf impact but still avoids relying on carefully lining up magic numbers.

PR #269, opened 2022-05-03
logging: keep UseJSON when creating prefixed loggers

jschaf:
Otherwise, prefixed loggers won't use the JSON output.

### DIFF
--- a/util/logger.go
+++ b/util/logger.go
@@ -19,11 +19,11 @@ type Logger struct {
 }
 
 func (logger *Logger) WithPrefix(prefix string) *Logger {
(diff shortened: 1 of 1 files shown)


lfittl (approved):
Good catch!

jschaf:
Do you all have a scheduled release cadence? I'd like to update to the latest for this fix.

lfittl:
> Do you all have a scheduled release cadence? I'd like to update to the latest for this fix.

We don't have a set release schedule, but should have this out within the next 1-2 weeks I think. Feel free to ping again if that is not the case - thanks for your patience!

PR #271, opened 2022-05-14
Improve the "too many tables" error message

uhoh-itsmaciek:
This error message does not make it clear (1) what the consequences are and (2) what you can do about it.

Attempt to address both of these by clarifying the impact and including a link.

### DIFF
--- a/input/postgres/schema.go
+++ b/input/postgres/schema.go
@@ -54,7 +54,7 @@ func CollectAllSchemas(server *state.Server, collectionOpts state.CollectionOpts
 		schemaTableLimit = defaultSchemaTableLimit
 	}
 	if relCount := len(ps.Relations); relCount > schemaTableLimit {
-		logger.PrintWarning("Too many tables: got %d, but only %d can be monitored per server; use ignore_schema_regexp config setting to filter", relCount, schemaTableLimit)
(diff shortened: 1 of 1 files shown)


input/postgres/schema.go
```diff
@@ -54,7 +54,7 @@ func CollectAllSchemas(server *state.Server, collectionOpts state.CollectionOpts
 		schemaTableLimit = defaultSchemaTableLimit
 	}
 	if relCount := len(ps.Relations); relCount > schemaTableLimit {
-		logger.PrintWarning("Too many tables: got %d, but only %d can be monitored per server; use ignore_schema_regexp config setting to filter", relCount, schemaTableLimit)
+		logger.PrintWarning("Too many tables: got %d, but only %d can be monitored per server; schema information will be lost; learn more at https://pganalyze.com/docs/collector/settings#schema-filter-settings", relCount, schemaTableLimit)
```
uhoh-itsmaciek, line 57: I concede this is getting a little wordy. Should we split this into two messages (they may get interleaved with output from other servers in high-concurrency situations), wrap the message to a new line (not sure if we're doing that anywhere else and may not be a good idea since dealing with log messages that include newlines can be tricky), or just reword? Thoughts?

lfittl: I think this is okay-ish - we could consider using the hint-style error message like the other one I just added, but that may not be ideal if this gets output a lot.

On the wording itself, maybe "schema information will not be sent" instead of "lost"

lfittl (approved):

PR #276, opened 2022-06-02
Heroku log processing: Avoid repeating the same line over and over again

lfittl:
This is a bug that was introduced in the refactoring in
735d95cc4e6d4ab012a2a5a92bc282bd180c2fb1, which kept the old logic that
remembers log lines across processing iterations. That is no longer needed
since we now handle that the same way for all streamed log providers in
setupLogStreamer.

### DIFF
--- a/input/system/heroku/logs.go
+++ b/input/system/heroku/logs.go
@@ -201,7 +201,6 @@ func setupLogTransformer(ctx context.Context, wg *sync.WaitGroup, servers []*sta
 	go func() {
 		defer wg.Done()
 
-		logLinesBySource := make(map[string][]state.LogLine)
 		sourceToServer := make(map[string]*state.Server)
 
 		for {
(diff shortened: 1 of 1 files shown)


input/system/heroku/logs.go
```diff
@@ -215,27 +214,22 @@ func setupLogTransformer(ctx context.Context, wg *sync.WaitGroup, servers []*sta
...
-					}
-
-					for _, logLine := range logLines {
-						logLine.Username = server.Config.GetDbUsername()
-						logLine.Database = server.Config.GetDbName()
-						out <- state.ParsedLogStreamItem{Identifier: server.Config.Identifier, LogLine: logLine}
-					}
+				server, exists := sourceToServer[sourceName]
+				if !exists {
+					logger.PrintInfo("Ignoring log line since server can't be matched yet - if this keeps showing up you have a configuration error for %s", sourceName)
```
uhoh-itsmaciek, line 226: Not really something we should fix here (especially since I don't have any better suggestions), but this is a pretty bad error message: "you have a configuration" error doesn't really tell you anything, since there's very little configuration for a Heroku-based collector.

uhoh-itsmaciek (approved):
Tested this together with #277 and it appears to fix the issue.

PR #277, opened 2022-06-02
Add support for new Heroku log_line_prefix

uhoh-itsmaciek:
Update the collector to support new Heroku log_line_prefix

Heroku recently updated the log_line_prefix on their new databases
[1]. Through regular maintenance, this new prefix will eventually come
to old databases as well. The new prefix is rejected by the collector
test and not parsed correctly by the log processing code.

Update the Heroku-specific code path to accept this new prefix as
well.

[1]: https://devcenter.heroku.com/changelog-items/2419

### DIFF
--- a/input/system/heroku/logs.go
+++ b/input/system/heroku/logs.go
@@ -154,24 +154,24 @@ func logStreamItemToLogLine(item HerokuLogStreamItem, servers []*state.Server, s
 	}
 	backendPid, _ := strconv.ParseInt(parts[1], 10, 32)
 
-	contentParts := regexp.MustCompile(`^\[(\w+)\] \[(\d+)-(\d+)\] (.+)`).FindStringSubmatch(string(item.Content))
-	if len(contentParts) != 5 {
+	lineParts := regexp.MustCompile(`^\[(\w+)\] \[(\d+)-(\d+)\] (.+)`).FindStringSubmatch(string(item.Content))
+	if len(lineParts) != 5 {
 		fmt.Printf("ERR: %s\n", string(item.Content))
 		return sourceToServer, nil, ""
 	}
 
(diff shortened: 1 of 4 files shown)


input/system/heroku/logs.go
```diff
@@ -173,7 +173,7 @@ func logStreamItemToLogLine(item HerokuLogStreamItem, servers []*state.Server, s
 	}
 	backendPid, _ := strconv.ParseInt(parts[1], 10, 32)
 
-	contentParts := regexp.MustCompile(`^\[(\w+)\] \[(\d+)-(\d+)\] ( sql_error_code = \w+ (\w+):  )?(.+)`).FindStringSubmatch(string(item.Content))
+	contentParts := regexp.MustCompile(`^\[(\w+)\] \[(\d+)-(\d+)\] (?:( sql_error_code = \w+ (\w+):  )|( sql_error_code = \w+ time_ms = ".+?" pid=".+?" proc_start_time=".+?" session_id=".+?" vtid=".+?" tid=".+?" log_line=".+?" .+?database=".+?" connection_source=".+?" user=".+?" application_name=".+?" ))?(.+)`).FindStringSubmatch(string(item.Content))
```
uhoh-itsmaciek, line 176: Not sure about this, since it seems like an ad-hoc copy of the regex in `parse.go`. Testing it now with a Heroku DB. Unfortunately the Saascontrol primary has already been updated, so I'm not sure if we have an old Heroku database we can test with to ensure this does not break the old prefix handling.

uhoh-itsmaciek:
Updated this and tested it on Heroku with a db using the new prefix. We're still getting some duplicate lines (see #276), but otherwise this works.

lfittl (approved):

PR #278, opened 2022-06-03
Google Cloud SQL: Re-enable log stitching for messages

lfittl:
Whilst the GCP release notes mention that this is no longer a problem as of
Sept 2021, log events can still be split up into multiple messages if they
exceed a threshold around 1000-2000 lines, or ~100kb.

In order to fix message like this, re-enable the logic that was previously
turned off in commit 735d95cc4e6d4ab012a2a5a92bc282bd180c2fb1.

This can be reproduced by a query like this on a pgbench database:

SELECT aid FROM pgbench_accounts, pg_sleep(1)
UNION SELECT aid FROM pgbench_accounts --- repeat this line 100 times
ORDER BY 1 LIMIT 1

### DIFF
--- a/input/system/google_cloudsql/logs.go
+++ b/input/system/google_cloudsql/logs.go
@@ -156,14 +156,12 @@ func setupLogTransformer(ctx context.Context, wg *sync.WaitGroup, servers []*sta
 					return
 				}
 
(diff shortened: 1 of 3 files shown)


uhoh-itsmaciek (approved):
Linting problems, but otherwise looks good.

PR #281, opened 2022-06-17
Fix rare bug with duplicate pg_settings values on Aurora Postgres

lfittl:
On some systems, notably Aurora Postgres 12.6, it appears possible to have
duplicate values in the pg_settings output, specifically as seen with the
"ssl_min_protocol_version" and "ssl_max_protocol_version" settings.

This is most likely a bug in the modified Postgres engine.

Since this causes other problems in the processing of the settings,
ensure that at most we get one result row per setting name.

### DIFF
--- a/input/postgres/settings.go
+++ b/input/postgres/settings.go
@@ -27,7 +27,7 @@ SELECT DISTINCT ON (name)
 	   sourcefile,
 	   sourceline
   FROM pg_catalog.pg_settings
- ORDER BY name, sourceline DESC NULLS LAST`
+ ORDER BY name, CASE source WHEN 'default' THEN 1 ELSE 0 END`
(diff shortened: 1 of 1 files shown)


seanlinsley (approved):
Without more information I'd wonder if perhaps `sourceline` is present or null when we don't expect it to be. A more explicit handling might use a `CASE source WHEN 'default' THEN 1 ELSE 0 END` in the order by expression.

lfittl:
> Without more information I'd wonder if perhaps `sourceline` is present or null when we don't expect it to be. A more explicit handling might use a `CASE source WHEN 'default' THEN 1 ELSE 0 END` in the order by expression.

Good idea - I'll adjust it to that.

PR #284, opened 2022-06-27
Add optional normalization of sensitive fields in EXPLAIN plans

lfittl:
This adds the new "normalize" option for the "filter_query_sample"
configuration setting. When this option is selected, EXPLAIN plans that
are collected from either auto_explain or log-based EXPLAIN are analyzed
on a field-by-field basis. Each field known to contain Postgres expressions
(which may contain constant values, which could contain PII), such as
"Filter" or "Index Cond", is ran through the same normalization function
applied to query texts. Unknown EXPLAIN fields are discarded when this
option is active.

Turning on this setting will also cause all query samples (whether they
have an EXPLAIN attached or not) to have their query text normalized
and their parameters marked as "\<removed\>".

This setting is recommended when EXPLAIN plans may contain sensitive
data that should not be stored. Please verify that the logic works
as expected with your workload and log output.


logs/querysample/querysample.go
```diff
@@ -0,0 +1,406 @@
...
+		return planRaw, err
+	}
+	normalizePlanNodeFields(plan)
+	return json.Marshal(plan)
+}
+
+func transformExplainJSONToQuerySample(logLine state.LogLine, explainText string, queryRuntimeMs float64, normalize bool) (state.PostgresQuerySample, error) {
+	var planDetails PlanContainer
+
+	if err := json.Unmarshal([]byte(explainText), &planDetails); err != nil {
```
uhoh-itsmaciek, line 282: Does this mean we're parsing and re-formatting the JSON even if we're not normalizing? We're not doing this today, are we? Do we need to do that? I don't know if this is still an issue, but I believe the `encoding/json` performance was [not great](https://github.com/json-iterator/go#benchmark). Could this be a performance issue for users with large plans?

logs/querysample/querysample.go
```diff
@@ -0,0 +1,406 @@
...
+		Database:      logLine.Database,
+		LogLineUUID:   logLine.UUID,
+		HasExplain:    true,
+		ExplainSource: pganalyze_collector.QuerySample_AUTO_EXPLAIN_EXPLAIN_SOURCE,
+		ExplainFormat: pganalyze_collector.QuerySample_TEXT_EXPLAIN_FORMAT,
+		ExplainOutput: explainParts[2],
+	}, nil
+}
+
+func TransformLogMinDurationStatementToQuerySample(logLine state.LogLine, queryText string, queryRuntime string, queryProtocolStep string, parameterParts [][]string) (state.PostgresQuerySample, bool) {
```
uhoh-itsmaciek, line 372: Maybe we should document the semantics of the boolean return value here, at least with a name?

logs/querysample/querysample.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	queryText = strings.TrimSpace(queryText)
+	if queryText == "" {
+		return state.PostgresQuerySample{}, false
+	}
+
+	// Note that we intentionally don't normalize here, because this may be
+	// passed on to log-based EXPLAIN still. We instead normalize this at a
+	// later step (in the output functions).
```
uhoh-itsmaciek, line 386: Ah, I take it this is why we normalize during explain? Could we instead have a separate `OriginalQuery` field or something in the struct that we use during explain (but not serialize to PB)? It feels brittle to depend on normalization to happen later. Also, does this work if explain is not enabled?

output/compact_logs.go
```diff
@@ -16,6 +17,18 @@ func UploadAndSendLogs(server *state.Server, grant state.GrantLogs, collectionOp
 	if server.Config.FilterQuerySample == "all" {
 		logState.QuerySamples = []state.PostgresQuerySample{}
 	}
+	if server.Config.FilterQuerySample == "normalize" {
+		for idx, sample := range logState.QuerySamples {
+			// EXPLAIN was already handled separately, here we just need to normalize non-EXPLAIN samples
```
uhoh-itsmaciek, line 22: Okay, I see--this case is handled after all. Given this package is called "output", though, I'm not sure if this is the best place for it...

uhoh-itsmaciek (commented):
+1 on the general idea and the filtering mechanism itself, but I think it might be better to do the filtering in one place, and allow log-based EXPLAIN access to the unfiltered query instead. Right now the filtering happens in a few different places and it's tricky to reason about. Thoughts?

logs/querysample/querysample.go
lfittl: Yeah, we already do this - in fact this approach is now slightly improving on the performance of it, because we no longer unmarshal the actual plan (we keep that as json.RawMessage) unless the `normalize` setting is active.

In my understanding there are two reasons we want to do this:

* Avoid generating samples that have broken JSON (due to truncated messages)
* Remove the `Query Text` field from the top level node, since its duplicative of what we already remember separately

We could consider dropping this step (since we already parse the JSON on the server-side anyway). I'll look into doing this together with refactoring where we do the normalization.

lfittl: Ah. After some more digging, we can't actually avoid this, because the query text itself is only contained within the JSON. We need to parse the JSON so we can get the query, so we can then fingerprint it (and associate the query sample and log line). I'll thus keep the JSON parsing.

lfittl:
> +1 on the general idea and the filtering mechanism itself, but I think it might be better to do the filtering in one place, and allow log-based EXPLAIN access to the unfiltered query instead. Right now the filtering happens in a few different places and it's tricky to reason about. Thoughts?

Yep, agreed, and thanks for calling this out (I was 50/50 on whether refactoring it was worth it, but I think it is).

I've now reworked it to flow more clearly, by having normalization triggered in the logs runner, in the existing post-processing function. For reviewing looking at this commit-by-commit might be easiest.

input/postgres/explain.go
```diff
@@ -109,6 +109,23 @@ func runDbExplain(db *sql.DB, inputs []state.PostgresQuerySample, useHelper bool
 				}
 			}
 		}
+		// Ensure we always normalize the query text (when sample normalization is on), even if EXPLAIN errors out
+		if normalize {
+			sample.Query = util.NormalizeQuery(sample.Query, "unparseable", -1)
```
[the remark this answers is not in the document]
lfittl: Fixed in the refactored version (this now happens in the logs runner post-processing)

logs/querysample/querysample.go
lfittl: This difference (between log-based EXPLAIN and auto_explain normalization) is now gone in the refactoring, as normalization always runs as an explicit step in the logs runner post-processing.

I didn't go with the `OriginalQuery` field idea, since if we wanted to implement this throughout we'd need a few duplicate fields, and I don't think thats necessary since the post-processing is centralized in a function and so its easy to audit that all sensitive fields are taken care of.

logs/querysample/normalize.go
```diff
@@ -0,0 +1,242 @@
...
+	*expr = strings.TrimPrefix(res, "SELECT ORDER BY ")
+}
+
+func normalizeExprArray(exprArray *[]string) {
+	if exprArray == nil {
+		return
+	}
+	for idx, expr := range *exprArray {
+		normalizeExpr(&expr)
+		(*exprArray)[idx] = expr
```
uhoh-itsmaciek, line 161: Is this still necessary? You're updating the expr by reference, no? But NBD.

uhoh-itsmaciek (approved):

logs/querysample/normalize.go
lfittl: Yeah, this is one of the Go-isms that I don't like..

This is necessary (and in fact a test fails if this line is not present), because `&expr` references the location of the `expr` variable, which is a copy of the list element - not the list element itself. Essentially `normalizeExpr` just does `expr = ...` but that won't modify the list itself (since expr is not a pointer, but a simple string).

sj26:
(This is amazing, thank you!)

PR #285, opened 2022-06-27
Custom types: Correctly track custom type reference for array types

lfittl:
Array types are technically separate in pg_type, but for our use case
all we need to make sure is that we reference the base type on the column
to allow downstream logic to know which custom type definition needs to
be created before running a CREATE TABLE that references said type.

### DIFF
--- a/input/postgres/types.go
+++ b/input/postgres/types.go
@@ -9,6 +9,7 @@ import (
 
 const typesSQL string = `
 SELECT t.oid,
+       t.typarray AS arrayoid,
        n.nspname AS schema,
        t.typname AS name,
        t.typtype AS type,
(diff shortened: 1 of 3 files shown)


uhoh-itsmaciek (approved):
I'm not 100% clear on the impact here: does this affect schema definitions, or only dependencies? If the former, I wonder if we're not being too clever here (if just the latter, this seems reasonable).

lfittl:
> I'm not 100% clear on the impact here: does this affect schema definitions, or only dependencies? If the former, I wonder if we're not being too clever here (if just the latter, this seems reasonable).

Only dependencies - the schema definitions come from the data type name that we track separately.

PR #286, opened 2022-06-28
Add RDS instance role hint when NoCredentialProviders error is hit

uhoh-itsmaciek:


runner/logs.go
```diff
@@ -122,6 +122,10 @@ func downloadLogsForServerWithLocksAndCallbacks(wg *sync.WaitGroup, server *stat
 	if err != nil {
 		server.LogStateMutex.Unlock()
 		prefixedLogger.PrintError("Could not collect logs for server: %s", err)
+		msg := err.Error()
+		if server.Config.SystemType == "amazon_rds" && strings.Contains(msg, "NoCredentialProviders") {
+			prefixedLogger.PrintInfo("HINT - This may occur if you have not assigned an IAM role to the collector instance")
```
lfittl, line 127: I think that makes sense, but maybe slight reword:

```
HINT - This may occur if you have not assigned an IAM role to the collector EC2 instance, or provided AWS credentials through another method
```

(since its perfectly legitimate to not use the IAM role setting)

lfittl (approved):

runner/logs.go
uhoh-itsmaciek: I thought we should steer people toward the common IAM role config, but on reflection, I think you're right--we should try to keep the message more generally applicable, even if it's more complex.

uhoh-itsmaciek: Reworded slightly for clarity:
>HINT - This may occur if you have not assigned an IAM role to the collector EC2 instance, and have not provided AWS credentials through another method

PR #287, opened 2022-06-28
Add support for Aiven system type

uhoh-itsmaciek:
See also https://github.com/pganalyze/collector-snapshot/pull/18


config/read.go
```diff
@@ -364,6 +364,13 @@ func preprocessConfig(config *ServerConfig) (*ServerConfig, error) {
 				config.CrunchyBridgeClusterID = parts[1]
 			}
 		}
+	} else if strings.HasSuffix(host, ".aivencloud.com") {
+		parts := strings.SplitN(host, ".", 2)
```
lfittl, line 368: It appears that the Aiven format is actually `serviceId-projectId.aivencloud.com`.

For example, with a project named `myexample12345` and a service named `pg-1234abcd`, the hostname thats generated is `pg-1234abcd-myexample12345.aivencloud.com`.

I think its consistent that aiven uses a single dash in the service ID (and you can't modify that), so I wonder if we should split that here? (everything after the second dash is the project ID, everything before is the service ID)

If we were to call the API in the future, that also requires this split: https://api.aiven.io/doc/#operation/PGServiceQueryStatistics

config/identify_system.go
```diff
@@ -46,6 +46,11 @@ func identifySystem(config ServerConfig) (systemType string, systemScope string,
 		if systemID == "" {
 			systemID = config.CrunchyBridgeClusterID
 		}
+	} else if (config.AivenServiceID != "") || systemType == "aiven" {
+		systemType = "aiven"
+		if systemID == "" {
+			systemID = config.AivenServiceID
```
lfittl, line 52: I wonder if we should be setting something like `systemTypeFallback` and `systemIDFallback` here.

For comparison, we have similar logic earlier in this file for AWS, since we changed how we identify AWS systems, and we use the fallback logic to make sure that the API first checks if there is an existing server that it should lookup based on the fallback values, and only then would go ahead and do the regular logic.

That fallback logic doesn't support the ID and Type fields yet, but there is no reason it couldn't. And that way we would have a seamless transition for servers that are currently identified as "Self-managed" but are actually Aiven. Otherwise upgrading the collector would cause them to get a new server entry in pganalyze (with all new statistics going to that new server entry).

uhoh-itsmaciek: That's a good point. I copied the Crunchy integration here, but you're right, this case is different. I'll look into that.

config/read.go
uhoh-itsmaciek: Yeah, I noticed that but figured the identifier was opaque and didn't want to risk a format change from Aiven breaking things on our end. But future API integration is a good point. Should we track the project name separately then? Or can we ignore it?

lfittl: I think tracking it separately seems like the right way to go.

uhoh-itsmaciek: Done that way in 2f21961a40467d4d1837bdd5f2d34341f142756f

config/identify_system.go
```diff
@@ -6,12 +6,14 @@ import (
...
-func identifySystem(config ServerConfig) (systemType string, systemScope string, systemScopeFallback string, systemID string) {
+func identifySystem(config ServerConfig) (systemID string, systemType string, systemScope string, systemIDFallback string, systemTypeFallback string, systemScopeFallback string) {
 	// Allow overrides from config or env variables
+	systemID = config.SystemID
 	systemType = config.SystemType
 	systemScope = config.SystemScope
+	systemIDFallback = config.SystemIDFallback
+	systemTypeFallback = config.SystemTypeFallback
 	systemScopeFallback = config.SystemScopeFallback
-	systemID = config.SystemID
```
uhoh-itsmaciek, line 14: I ended up rearranging this since we are not always consistent about how we order these. In most places we have (id, type, scope), which makes sense to me, so that's what I tried to standardize on. I think it also makes sense to keep the fallback values together, rather than intersperse them with the main identity values. In some places (mostly protobuf definitions), we don't have the flexibility to do that, but most cases can be consistent.

config/identify_system.go
```diff
@@ -6,12 +6,14 @@ import (
 )
 
 // Figure out if we're self-hosted or on RDS, as well as what ID we can use - Heroku is treated separately
-func identifySystem(config ServerConfig) (systemType string, systemScope string, systemScopeFallback string, systemID string) {
+func identifySystem(config ServerConfig) (systemID string, systemType string, systemScope string, systemIDFallback string, systemTypeFallback string, systemScopeFallback string) {
```
uhoh-itsmaciek, line 9: This is kind of a lot of return values, but I don't have any great ideas here. We could use a struct, but given this is only called in two places, and the values are immediately assigned to fields of a Config object, I don't know if that's helpful.

config/identify_system.go
```diff
@@ -46,10 +48,15 @@ func identifySystem(config ServerConfig) (systemType string, systemScope string,
 		if systemID == "" {
 			systemID = config.CrunchyBridgeClusterID
 		}
-	} else if (config.AivenServiceID != "") || systemType == "aiven" {
-		systemType = "aiven"
+	} else if (config.AivenProjectID != "" && config.AivenServiceID != "") || systemType == "aiven" {
+		systemTypeFallback = "aiven"
```
uhoh-itsmaciek, line 52: Looking at how this works for AWS, this is not quite right--still reviewing, but I'll update.

config/identify_system.go
lfittl: I think that's fine for now - a struct seems like the way to go if we wanted to improve on it (possible a struct for the set of id+type+scope, so its two return values?)

config/identify_system.go
lfittl: Yep, and I just tested this and it worked as expected :)

input/collector.go
```diff
@@ -103,9 +103,12 @@ func getCollectorConfig(c config.ServerConfig) state.CollectorConfig {
 		GcpHasCredentialsFile:      c.GcpCredentialsFile != "",
 		GcpProjectId:               c.GcpProjectID,
 		CrunchyBridgeClusterId:     c.CrunchyBridgeClusterID,
+		AivenServiceId:             c.AivenServiceID,
```
lfittl, line 106: Do we need `AivenProjectId` here?

lfittl (approved):
:+1:

config/identify_system.go
uhoh-itsmaciek: Great--I ended up doing that as well, and it worked too.

PR #290, opened 2022-06-29
Fix Heroku log handling when consuming logs for multiple databases

uhoh-itsmaciek:
Currently, we track which database logs belong to through
server.Config.Identifier, but that's not set for servers when running
on Heroku, which leads to all logs being sent to an arbitrary server.

Define server.Config.Identifier for Heroku servers like we do when
reading the config file.

This may have been broken in the refactoring in #240


lfittl (approved):
:+1: Excellent catch!

PR #292, opened 2022-07-08
Improve shutdown handling

uhoh-itsmaciek:
Signal the wait group passed to the log transform setup once the context is canceled.

### DIFF
--- a/input/system/selfhosted/logs.go
+++ b/input/system/selfhosted/logs.go
@@ -182,11 +182,11 @@ func tailFile(ctx context.Context, path string, out chan<- SelfHostedLogStreamIt
 	TailLoop:
 		for {
 			select {
+			case line := <-t.Lines():
+				out <- SelfHostedLogStreamItem{Line: line.String()}
 			case <-ctx.Done():
 				prefixedLogger.PrintVerbose("Stopping log tail for %s (stop requested)", path)
 				break TailLoop
-			case line := <-t.Lines():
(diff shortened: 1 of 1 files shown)


lfittl (approved):
:+1:

uhoh-itsmaciek:
Hmm, so the integration test here fails with

```
 2022/07/08 01:17:47 I [server1] Testing log collection (local)...
2022/07/08 01:17:47 I [server1] Setting up log tail for /var/log/postgresql/postgresql-10-main.log
2022/07/08 01:17:51 I [server1]   Local log test successful
panic: close of closed channel
goroutine 38 [running]:
github.com/pganalyze/collector/input/system/selfhosted.setupLogTransformer.func1()
	/collector/input/system/selfhosted/logs.go:390 +0x4dd
created by github.com/pganalyze/collector/input/system/selfhosted.setupLogTransformer
	/collector/input/system/selfhosted/logs.go:379 +0x134
  ✗ step failed: test command failed: exit status 2
```

I'm not sure how this channel can be closed twice...

uhoh-itsmaciek:
@lfittl can you take another look? I'm not sure how my original solution failed to hit the issue in local testing that CI ran into, but I think the new approach is safer.

lfittl:
> @lfittl can you take another look? I'm not sure how my original solution failed to hit the issue in local testing that CI ran into, but I think the new approach is safer.

Ah, yes - good catch. New fix looks good.

lfittl (approved):

PR #293, opened 2022-07-08
Track extensions that are installed on each database 

lfittl:
[Track extensions that are installed on each database](https://github.com/pganalyze/collector/commit/73ccbf9b29a42bd6a3ce6804dffaa45e62c44b19)

This is helpful to ensure that the necessary schema definitions are
loaded by pganalyze, e.g. for use by the Index Advisor.

Depends on https://github.com/pganalyze/collector-snapshot/pull/19

---

[Ignore objects that are provided by extensions, as determined by pg_depend](https://github.com/pganalyze/collector/commit/bf42042ba25166aefcdf18ccd6e95a3d069f6065)

Since we are now tracking which extensions are installed, its unnecessary
to keep track of functions/types/views defined by extensions, as we can
infer that from the extension and its version. Filter out these extension
owned objects through the pg_depend catalog information.


uhoh-itsmaciek (approved):
The code looks good, but I don't understand the CI failure--I can't tell what the actual error is.

lfittl:
> The code looks good, but I don't understand the CI failure--I can't tell what the actual error is.

Not sure, I couldn't explain it either - looks like it didn't occur on a re-run.

PR #295, opened 2022-07-16
Self-hosted system: Support disk statistics for software RAID devices

lfittl:
These statistics are summarized across all component disk devices and
then tracked for the parent software RAID device as one. Note that this
is only done in case these statistics are not set yet (which is the case
for the typical Linux software RAID setup).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -50,6 +50,8 @@ require (
 	gopkg.in/mcuadros/go-syslog.v2 v2.3.0
 )
 
+require github.com/prometheus/procfs v0.7.3
+
 require (
 	github.com/Azure/go-amqp v0.16.0 // indirect
 	github.com/Azure/go-autorest v14.2.0+incompatible // indirect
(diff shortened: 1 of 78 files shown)


go.mod
```diff
@@ -50,6 +50,8 @@ require (
 	gopkg.in/mcuadros/go-syslog.v2 v2.3.0
 )
 
+require github.com/prometheus/procfs v0.7.3
```
lfittl, line 53: Note that we could take a smaller portion of this package (we only need the `/proc/mdstat` handling), but this could actually be interesting to switch to in the future - gopsutils is pretty large, and this is a more targeted package for gathering Linux system statistics from /proc and /sys.

uhoh-itsmaciek (approved):
:+1:

uhoh-itsmaciek:
Tried this locally on Ubuntu 20.04. I only have root and swap partitions, but it did not seem to break anything.

lfittl:
@uhoh-itsmaciek Thanks! I'll do one more test on a non-Linux system (to confirm the use of /proc here didn't break anything), and then merge.

lfittl:
Works as expected on FreeBSD (disk statistics are still reported, even though /proc doesn't do anything on FreeBSD)

lfittl:
Works as expected on FreeBSD (disk statistics are still reported, even though /proc doesn't do anything on FreeBSD)

PR #298, opened 2022-07-26
Stop pretty-printing constraintdefs on collection

uhoh-itsmaciek:
See 63be3a0d7b2b71a59fe1287b849d3159eb00383f -- a similar problem can
occur with constraintdefs, if those defs are foreign key table references.


lfittl (approved):
:+1:

PR #301, opened 2022-07-29
Log Insights: Filter out log_statement=all and log_duration=on log lines

lfittl:
This replaces the previous behaviour that prevented all log collection
for servers that had either log_statement=all or log_duration=on enabled.

With the new logic, we continue ignoring these high-frequency events
(which would cause downstream problems), but accept all other log events,
including threshold-based auto_explain events.

Note that using log_min_duration_statement in combination with
log_duration=on is not supported, and would result in both types of events
being ignored.

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -282,6 +282,7 @@ func (s *Server) SetLogIgnoreFlags(ignoreStatement bool, ignoreDuration bool) {
 	if ignoreDuration {
 		newFlags |= LOG_IGNORE_DURATION
 	}
+	atomic.StoreUint32(&s.LogIgnoreFlags, newFlags)
 }
 
 // IgnoreLogLine - helper function that lets callers determine whether a log
--- a/state/state.go
+++ b/state/state.go
@@ -295,6 +295,6 @@ func (s *Server) SetLogIgnoreFlags(ignoreStatement bool, ignoreDuration bool) {
 func (s *Server) IgnoreLogLine(content string) bool {
 	flags := atomic.LoadUint32(&s.LogIgnoreFlags)
 
-	return (flags&LOG_IGNORE_STATEMENT != 0 && strings.HasPrefix(content, "statement: ")) ||
-		(flags&LOG_IGNORE_DURATION != 0 && strings.HasPrefix(content, "duration: "))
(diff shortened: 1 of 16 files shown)


state/state.go
```diff
@@ -257,4 +259,40 @@ type Server struct {
...
+const (
+	LOG_IGNORE_STATEMENT uint32 = 1 << iota
+	LOG_IGNORE_DURATION
+)
+
+func (s *Server) SetLogIgnoreFlags(ignoreStatement bool, ignoreDuration bool) {
+	var newFlags uint32
+	if ignoreStatement {
+		newFlags |= LOG_IGNORE_STATEMENT
+	}
```
uhoh-itsmaciek, line 280: This will set the flags if they're not set, but the old code was explicitly written to also clear them when they are set (so that you don't need to restart the collector after adjusting settings). I think this should be something like (ignoring the concurrency issue for now):

```golang
var newFlags uint32 = s.LogIgnoreFlags
if ignoreStatement {
  newFlags |= LOG_IGNORE_STATEMENT
} else {
  newFlags &^= LOG_IGNORE_STATEMENT
}
if ignoreDuration {
  newFlags |= LOG_IGNORE_DURATION
} else {
  newFlags &^= LOG_IGNORE_DURATION
}

s.LogIgnoreFlags = newFlags
```

state/state.go
```diff
@@ -257,4 +259,40 @@ type Server struct {
...
+	if ignoreDuration {
+		newFlags |= LOG_IGNORE_DURATION
+	}
+}
```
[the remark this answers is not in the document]
lfittl: Good catch, I had dropped that in a refactoring by accident.

state/state.go
lfittl: The idea was to always reset the flags, i.e. `s.LogIgnoreFlags = newFlags` (or rather the atomic version of it), and thus if the passed in flags are false, there is no need to update the flags variable.

uhoh-itsmaciek: Ah, got it. Yeah, assuming you're passing in the full state each time (which we are here), that works.

uhoh-itsmaciek (approved):

PR #306, opened 2022-08-05
Show RDS instance role hint when running collector test

uhoh-itsmaciek:
In fd65610fd60b0aaa2b2784ba23bb7f7e09c7f71e, we added a hint to
instruct users to check the instance role when hitting RDS credential
erros in log download. This works, but only in the live code path,
because the test code logs the download error from a different entry
point.

To work around this, add the same warning to the test code path.

Alternately, we could push the logging down into the download method
itself (in code that both code paths end up calling), but that breaks
up the logging success/failure cases in a weird way, especially for
the test path. Or we could build an error that includes the hint,
though there's less flexibility in formatting (the hint cannot be its
own line) if we do that. Finally, we could also use an error type that
includes separate message and hint fields, but that would require a
larger refactoring.

### DIFF
--- a/runner/logs.go
+++ b/runner/logs.go
@@ -127,11 +127,7 @@ func downloadLogsForServerWithLocksAndCallbacks(wg *sync.WaitGroup, server *stat
 	newLogState, success, err := downloadLogsForServer(server, globalCollectionOpts, prefixedLogger)
 	if err != nil {
 		server.LogStateMutex.Unlock()
-		prefixedLogger.PrintError("Could not collect logs for server: %s", err)
-		msg := err.Error()
-		if server.Config.SystemType == "amazon_rds" && strings.Contains(msg, "NoCredentialProviders") {
-			prefixedLogger.PrintInfo("HINT - This may occur if you have not assigned an IAM role to the collector EC2 instance, and have not provided AWS credentials through another method")
-		}
+		printLogDownloadError(server, err, prefixedLogger)
 		if server.Config.ErrorCallback != "" {
 			go runCompletionCallback("error", server.Config.ErrorCallback, server.Config.SectionName, "logs", err, prefixedLogger)
 		}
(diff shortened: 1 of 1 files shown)


uhoh-itsmaciek:
Not really thrilled with this solution, but I don't see great alternatives without a larger refactoring. Thoughts?

lfittl:
> Not really thrilled with this solution, but I don't see great alternatives without a larger refactoring. Thoughts?

I think this solution works for now - the duplication isn't great, but it appears we already have some other duplication for the other hint in the file (related to the collector running in the background).

The only immediate way I can think of, that you didn't mention, is to just have an explicit error handling method that does the error output - that way we could have the same code and avoid the message duplication. i.e. something like this:

```golang
func printLogDownloadError(prefixedLogger *util.Logger, error err) {
  prefixedLogger.PrintError("ERROR - Could not download logs: %s", err)
  msg := err.Error()
  if server.Config.SystemType == "amazon_rds" && strings.Contains(msg, "NoCredentialProviders") {
    prefixedLogger.PrintInfo("HINT - This may occur if you have not assigned an IAM role to the collector EC2 instance, and have not provided AWS credentials through another method")
  }
}
```

and then we modify the relevant part in `downloadLogsForServerWithLocksAndCallbacks` to look like this:

```golang
func downloadLogsForServerWithLocksAndCallbacks(wg *sync.WaitGroup, server *state.Server, globalCollectionOpts state.CollectionOpts, logger *util.Logger) {
        ...
	newLogState, success, err := downloadLogsForServer(server, globalCollectionOpts, prefixedLogger)
	if err != nil {
		server.LogStateMutex.Unlock()
		printLogDownloadError(prefixedLogger, err)
		if server.Config.ErrorCallback != "" {
			go runCompletionCallback("error", server.Config.ErrorCallback, server.Config.SectionName, "logs", err, prefixedLogger)
		}
	} else {
		server.LogPrevState = newLogState
		server.LogStateMutex.Unlock()
		if success && server.Config.SuccessCallback != "" {
			go runCompletionCallback("success", server.Config.SuccessCallback, server.Config.SectionName, "logs", nil, prefixedLogger)
		}
	}
}
```

and `testLogDownload` to look like this:

```golang
func testLogDownload(ctx context.Context, wg *sync.WaitGroup, server *state.Server, globalCollectionOpts state.CollectionOpts, prefixedLogger *util.Logger) bool {
	prefixedLogger.PrintInfo("Testing log download...")
	_, _, err := downloadLogsForServer(server, globalCollectionOpts, prefixedLogger)
	if err != nil {
		printLogDownloadError(prefixedLogger, err)
		return false
	}

	prefixedLogger.PrintInfo("  Log test successful")
	return true
}
```

lfittl (approved):

uhoh-itsmaciek:
I rejected that initially because it seemed like more ceremony without much benefit, but thinking about it again, it does at least make the duplication clear and should make it easier to refactor in the future. I'll go with that; thanks.

PR #307, opened 2022-08-06
AWS Aurora: Auto detect writer instance for log downloads / system metrics

lfittl:
Depends on https://github.com/pganalyze/collector-snapshot/pull/20


input/system/rds/system.go
```diff
@@ -36,11 +36,6 @@ func GetSystemState(config config.ServerConfig, logger *util.Logger) (system sta
 		return
 	}
 
-	if instance == nil {
-		logger.PrintWarning("Could not find RDS instance in AWS, skipping system data")
-		return
-	}
```
lfittl, line 42: This code is no longer needed since we now always return an error when we can't find the instance, instead of only returning `instance = nil`.

util/awsutil/rds.go
```diff
@@ -37,44 +38,99 @@ func FindRdsIdentifier(config config.ServerConfig, sess *session.Session) (ident
 	return
 }
 
-func FindRdsInstance(config config.ServerConfig, sess *session.Session) (instance *rds.DBInstance, err error) {
-	var resp *rds.DescribeDBInstancesOutput
+func findRdsClusterByIdentifier(clusterIdentifier string, svc *rds.RDS) (*rds.DBCluster, error) {
```
lfittl, line 41: Side effect of the way this gets called through `FindRdsIdentifier` => `FindRdsInstance`, is that in the case of a failover it will take us up to 5 minutes to update the mapping thats cached in `FindRdsIdentifier`.

This might be confusing/unexpected, since it'd show the wrong logs in pganalyze for those 5 minutes, instead of switching over.

We could consider having a shorter, separate TTL for the cluster => instance mapping, and handle that separately?

(e.g. assuming we download logs every 30 seconds, it might be reasonable to a `DescribeDBClusters` that often, but we could keep the `DescribeDBInstances` result cached)

uhoh-itsmaciek: No strong feelings here. This could be confusing, but it should be very rare, and I don't know that we want to block these changes on figuring out a good solution.

uhoh-itsmaciek (approved):

util/awsutil/rds.go
lfittl: Yeah, for now I'll leave this to avoid rate limiting problems.

PR #316, opened 2022-08-15
Quote PATH when spawning shells in Makefile

overdrivenpotato:
Fixes builds where `$PATH` has a space, e.g. `..:/Applications/VMware Fusion.app/Contents/Public:..`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ OUTFILE := pganalyze-collector
 PROTOBUF_FILES := $(wildcard protobuf/*.proto) $(wildcard protobuf/reports/*.proto)
 
 PATH := $(PWD)/protoc/bin:$(PWD)/bin:$(PATH)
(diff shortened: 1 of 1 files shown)


overdrivenpotato:
[`#12 0.208 /bin/sh: 1: go: not found`](https://github.com/pganalyze/collector/runs/7847857656?check_suite_focus=true#step:5:810)
I'm not sure how the diff could cause this. Maybe a spurious issue?

lfittl:
@overdrivenpotato Maybe a difference between `make` on Linux and `make` on macOS?

overdrivenpotato:
Looks like this only works for GNU Make 3.81, other versions like 4+ change the behavior of setting `SHELL := ` with quoted arguments. Closing this as there doesn't seem to be a good version-portable way to solve the problem.

PR #317, opened 2022-08-24
Collect RDS tags in snapshots

msepga:


input/system/rds/system.go
```diff
@@ -62,6 +62,13 @@ func GetSystemState(config config.ServerConfig, logger *util.Logger) (system sta
 		IsAuroraPostgres:           isAurora,
 	}
 
+	tags := make(map[string]string)
+	for _, tag := range instance.TagList {
+		tags[util.StringPtrToString(tag.Key)] = util.StringPtrToString(tag.Value)
```
msakrejda, line 67: We expect the `.Key` and `.Value` pointers to be non-nil here, right? I noticed `StringPtrToString` just turns nil pointers to empty strings, which is probably not useful for us here. Should we just dereference the pointers here instead (intentionally crashing if we run into a nil) or skip nil pointers?

msakrejda (approved):
I had a question about how to treat nil tag keys and values, but otherwise looks good.

input/system/rds/system.go
msepga: I've updated `Key` to be a straight dereference, but left `Value` as-is. It might be possible for `Tag.Value` to be `nil` here; the docs on that field state `A value is the optional value of the tag.`

lfittl (approved):

PR #318, opened 2022-08-29
Docker: Allow setting CONFIG_CONTENTS to pass ini-style configuration

lfittl:
This allows easier configuration of multiple servers to be monitored by
the same Docker container. Previously this required use of a volume
mount, which can be harder to make work successfully.

CONFIG_CONTENTS needs to match the regular configuration file format that
uses separate sections for each server.

This can be combined with environment-variable style configuration for
settings that apply to all servers (e.g. PGA_API_KEY) but all
server-specific configuration should only be passed in through the
CONFIG_CONTENTS variable.

### DIFF
--- a/contrib/docker-entrypoint.sh
+++ b/contrib/docker-entrypoint.sh
@@ -2,6 +2,10 @@
 
 set -e
 
+if [ -n "$CONFIG_CONTENTS" ]; then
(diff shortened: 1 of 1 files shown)


msakrejda (approved):
Nice!

PR #322, opened 2022-09-28
Log streaming: Discard logs after consistent failures to upload

lfittl:
Previously we kept logs around after the grant or upload failed, but this actually causes more trouble in situations where the upstream server is not available. In such situations the collector will try reprocessing more and more data until the server accepts data again, or the collector crashes.

Since we already have HTTP client retries enabled since 314edc86e90b1f584c364b8665c42c3e643f2d86, rely on that to workaround transient failures, and otherwise discard the logs. This is is similar to log downloads where we also discard logs if we can't send them (after the HTTP retry limit is reached).


keiko713 (approved):
change makes sense 👍

seanlinsley (approved):

PR #324, opened 2022-10-06
Prevent synchronous runs of the full snapshot collector

seanlinsley:
When the Postgres database is overloaded or the collector instance is under-provisioned, it's possible for a full snapshot to not finish in time before the next full snapshot is scheduled to begin. That can result in corrupted snapshots as well as resource contention leading to an out of memory event.

It would definitely be good to report this in the next full snapshot so we can alert customers to this issue. Looks like `CollectorErrors` is filled in by `logger.PrintError`, but it's not clear if the same is true for `prefixedLogger.PrintError`.

The correctness of this code definitely needs review. For example I think the early `return` is actually wrong because `wg.Done()` is never called.

### DIFF
--- a/runner/full.go
+++ b/runner/full.go
@@ -191,6 +191,17 @@ func CollectAllServers(servers []*state.Server, globalCollectionOpts state.Colle
 				prefixedLogger.PrintInfo("Testing statistics collection...")
 			}
 
+			server.CollectionStatusMutex.Lock()
+			if server.CollectionStatus.FullSnapshotInProgress {
+				prefixedLogger.PrintError("Could not collect full snapshot for server because the last one is still in progress")
+				allSuccessful = false
+				server.CollectionStatusMutex.Unlock()
+				return
+			} else {
+				server.CollectionStatus.FullSnapshotInProgress = true
+				server.CollectionStatusMutex.Unlock()
+			}
+
 			server.StateMutex.Lock()
 			newState, grant, newCollectionStatus, err := processServer(server, globalCollectionOpts, prefixedLogger)
 			if err != nil {
@@ -242,6 +253,9 @@ func CollectAllServers(servers []*state.Server, globalCollectionOpts state.Colle
 					go runCompletionCallback("success", server.Config.SuccessCallback, server.Config.SectionName, "full", nil, prefixedLogger)
 				}
 			}
+			server.CollectionStatusMutex.Lock()
+			server.CollectionStatus.FullSnapshotInProgress = false
+			server.CollectionStatusMutex.Unlock()
 			wg.Done()
 		}(servers[idx])
 	}
--- a/state/state.go
+++ b/state/state.go
@@ -247,6 +247,7 @@ type CollectionStatus struct {
 	CollectionDisabledReason  string
 	LogSnapshotDisabled       bool
 	LogSnapshotDisabledReason string
+	FullSnapshotInProgress    bool
 }
 
 type Server struct {


msakrejda (commented):
Right, I think we can't skip that `wg.Done()`. Plus I just opened #325 to illustrate what I was talking about with copied mutexes. If we want to add more locking for correctness, I think we should fix those issues first or it's going to be really hard to reason about our locks.

lfittl (commented):
I don't think this is necessary, because [the scheduler](https://github.com/pganalyze/collector/blob/main/scheduler/scheduler.go#L15) currently works by only scheduling the next run if the previous one completed.

(and `CollectAllServers` ensures that all per-server go routines have completed, through the use of wait groups, before returning to the caller)

If we're concerned about concurrent activity, I would review the interaction between the `query_stats` and the `stats` scheduler groups. They are supposed to exclude each other (through the logic in `ScheduleSecondary` skipping query stats gathering on the full 10 minute mark), but in case `query_stats` takes longer than 60 seconds that would not be sufficient and `CollectAllServers` and `GatherQueryStatsFromAllServers` would then run concurrently.

However, there are mutexes in place already to handle this (`server.StateMutex`) and so we should be fine in this situation. But maybe we missed something.

seanlinsley:
> I don't think this is necessary, because [the scheduler](https://github.com/pganalyze/collector/blob/main/scheduler/scheduler.go#L15) currently works by only scheduling the next run if the previous one completed.

Can you go into more detail about how that happens? From my reading, if a full snapshot takes over 10 minutes there's nothing that prevents it from overlapping with the next scheduled run. We have seen duplicated query stats that suggests memory corruption caused by concurrent full snapshot runs.

> If we're concerned about concurrent activity, I would review the interaction between the `query_stats` and the `stats` scheduler groups. They are supposed to exclude each other (through the logic in `ScheduleSecondary` skipping query stats gathering on the full 10 minute mark), but in case `query_stats` takes longer than 60 seconds that would not be sufficient and `CollectAllServers` and `GatherQueryStatsFromAllServers` would then run concurrently.

Good point, that's also worth looking into.

Adding a bit of context: I don't know exactly what is overlapping, but I've seen our own collector instance fall over in a way that suggested this issue. It was running multiple threads in parallel that were each attempting to use as much CPU time as possible. Eventually the system ran out of memory and the process rebooted. This was resolved by giving it more CPU and RAM, but ideally we would prevent so many things from running in parallel on resource-constrained systems.

lfittl:
> > I don't think this is necessary, because [the scheduler](https://github.com/pganalyze/collector/blob/main/scheduler/scheduler.go#L15) currently works by only scheduling the next run if the previous one completed.
> 
> Can you go into more detail about how that happens? From my reading, if a full snapshot takes over 10 minutes there's nothing that prevents it from overlapping with the next scheduled run. We have seen duplicated query stats that suggests memory corruption caused by concurrent full snapshot runs.

In my understanding:

- In `Schedule` ([source](https://github.com/pganalyze/collector/blob/main/scheduler/scheduler.go#L15)) the `runner()` is a synchronous (as in, not called via a goroutine) function call in the scheduler - thus the `Schedule` function will wait for it to return
- The anonymous runner function ([source](https://github.com/pganalyze/collector/blob/main/main.go#L182)) is synchronously calling `runner.CollectAllServers`
- `CollectAllServers` ([source](https://github.com/pganalyze/collector/blob/bf892a45f1a0dee10a858b854fd8dc33302dc32a/runner/full.go#L249)) does a `wg.Add()` before starting a go routine for each server, and then does a `wg.Wait()` at the end waiting for those servers to have completed processing (and thus should be blocking on the slowest server to be processed)

Therefore I'm pretty certain that this particular code path does not run concurrently, unless I'm missing something.

> Adding a bit of context: I don't know exactly what is overlapping, but I've seen our own collector instance fall over in a way that suggested this issue. It was running multiple threads in parallel that were each attempting to use as much CPU time as possible. Eventually the system ran out of memory and the process rebooted. This was resolved by giving it more CPU and RAM, but ideally we would prevent so many things from running in parallel on resource-constrained systems.

I think that today the main source of concurrency should be from having multiple servers being configured (as noted above, we start a go routine per server, so they do run in parallel) - otherwise we shouldn't be running concurrently except for different unrelated kinds of snapshots (e.g. an activity snapshot will be collected at the same time as a full snapshot, but I don't think that would be a problem per se).

PR #325, opened 2022-10-06
Add go vet linting to Makefile

msakrejda:
Currently, this fails with a lot of 'copies lock by value' errors
(see, e.g.,
https://medium.com/golangspec/detect-locks-passed-by-value-in-go-efb4ac9a3f2b
). I think many of them are moot, but we should address these so we
can add go vet since it can find real issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,7 @@ vendor:
 
 test: build
 	go test -coverprofile=coverage.out ./...
+	go vet ./...
 	# go tool cover -html=coverage.out
 
 integration_test:


lfittl (commented):
:+1: to adding this.

It looks like all of the current complaints are false positives, since they relate to how we handle the protobuf struct (and that never happens concurrently, I believe, since that struct is only generated in the output handling which runs within the same Go routine).

Whilst we could try to handle this better, maybe its best to just ignore them for now?

msakrejda:
Thanks for the analysis. I'm not sure there's a good way to exclude those other than omitting all of 'output' from linting. The idea of comments for that has been [discussed and rejected](https://github.com/golang/go/issues/18432) by the Go team. We could I guess omit all of `output` from `go vet` but that seems like a pretty heavyweight solution...

msakrejda:
Closing this for now since I don't plan to work on it in the near term. I think the right way to handle the generated code checks would be to put that in its own directory, and skip that in `go vet` (I don't think `go vet` supports more sophisticated ignore rules, and it'd be a shame to skip all of `output`).

PR #326, opened 2022-10-06
Collect lock information with compact activity snapshot

keiko713:
With this PR, we collect lock information and add it to the compact activity snapshot. To collect lock information, we use `pg_blocking_pids()`.

This function is introduced PG9.6, so the lock information collection is only supported from PG9.6, as well as when `no-postgres-locks` option is not true (default is false, so people need to opt-out the lock info collection if they wish).

This collects 4 new fields:

* blocked_by_pids: pids that this backend is blocked by
* blocking_pids: pids that this backend is blocking
* indirectly_blocked_by_pids: pids that this backend is indirectly blocked by
* indirectly_blocking_pids: pids that this backend is indirectly blocking

`blocked_by_pids` is pretty straight forward to collect as it's exactly what `pg_blocking_pids()` gives, but other ones are calculated/generated based on that info manually. I'm super open to change the logic of how to calculate these PIDs, I don't believe my logic is the best one (but... it works).

Related PR in snapshot side: https://github.com/pganalyze/collector-snapshot/pull/24


input/postgres/backends.go
```diff
@@ -5,23 +5,38 @@ import (
...
-				backend_start, xact_start, query_start, state_change, %s, state, query
-	 FROM %s
-	WHERE pid IS NOT NULL`
+const activitySQLpg96OptionalFields = `COALESCE(wait_event_type, '') = 'Lock' as waiting, backend_xid, backend_xmin, wait_event_type, wait_event, NULL`
+const activitySQLpg10OptionalFields = `COALESCE(wait_event_type, '') = 'Lock' as waiting, backend_xid, backend_xmin, wait_event_type, wait_event, backend_type`
+
+const pgBlockingPidsField = `
+CASE
+	WHEN COALESCE(wait_event_type, '') = 'Lock' THEN pg_blocking_pids(pid)
+	ELSE NULL
```
msakrejda, line 21:
```suggestion
```
`NULL` is the default value when no `WHEN` clauses of a `CASE` statement are matched, so you could just omit this line.

msakrejda (commented):
I have thoughts on the indirectly blocking calculation, but what's the advantage of deriving that here, versus doing it on the frontend when a user is looking at this data, or in the backend when processing a snapshot?

keiko713:
> I have thoughts on the indirectly blocking calculation, but what's the advantage of deriving that here, versus doing it on the frontend when a user is looking at this data, or in the backend when processing a snapshot?

Great point. I think initially I moved this logic to the collector because I needed to calculate `blocking_pids` in 2 different places in backend/frontend side, then it was easier to add that in snapshot (the data was also small).
However, later on, we decided to add more info (indirectly ones). Indeed it's making less sense to have this logic here, as it's going to add more info in snapshot that we could calculate later on 🤔 
I'm gonna make a version of only adding `blocked_by_pids`!

lfittl:
> > I have thoughts on the indirectly blocking calculation, but what's the advantage of deriving that here, versus doing it on the frontend when a user is looking at this data, or in the backend when processing a snapshot?
> 
> Great point. I think initially I moved this logic to the collector because I needed to calculate `blocking_pids` in 2 different places in backend/frontend side, then it was easier to add that in snapshot (the data was also small). However, later on, we decided to add more info (indirectly ones). Indeed it's making less sense to have this logic here, as it's going to add more info in snapshot that we could calculate later on 🤔 I'm gonna make a version of only adding `blocked_by_pids`!

Another motivation here was to be able to say "Alert me when a query blocks **10 or more other connections** (either directly or indirectly)" - calculating this in the collector makes it easy for us to hook an alert into the collected data on the statistics processing side without having to do any extra calculations.

That said, there is no reason this handling couldn't happen on the processing side (in the pganalyze) instead of in the collector - I think either approach should work.

input/postgres/backends.go
```diff
@@ -73,8 +96,71 @@ func GetBackends(logger *util.Logger, db *sql.DB, postgresVersion state.Postgres
...
+		pid := int64(activity.Pid)
+		activities[i].BlockingPids = findBlocking(blockingGraph, pid)
+		activities[i].IndirectlyBlockingPids = findIndirectlyBlocking(blockingGraph, activities[i].BlockingPids)
+		activities[i].IndirectlyBlockedByPids = findIndirectlyBlockedBy(blockingGraph, activities[i].BlockedByPids)
+	}
+
 	return activities, nil
 }
+
+func findBlocking(graph []Edge, pid int64) []int64 {
```
lfittl, line 119: We could move this logic into the `util` folder for better abstraction (together with the test), since its fairly independent. That assumes we keep the logic in the collector.

lfittl (commented):
I think overall this looks good!

I don't have a clear preference on this approach vs the one taken in #327 - as mentioned the main consideration I would add is that we may want to alert on some of the calculated data.

Additionally, as discussed, we should do a pgbench test to ensure this doesn't add noticable overhead (it shouldn't, since its only running pg_blocking_pids when there is a lock wait event already), but that would help us determine if we need to do any kind of caching in the initial implementation (assuming we keep this turned on by default, like done here).

keiko713:
> Another motivation here was to be able to say "Alert me when a query blocks 10 or more other connections (either directly or indirectly)" - calculating this in the collector makes it easy for us to hook an alert into the collected data on the statistics processing side without having to do any extra calculations.

Great reminder! I think we indeed still have a chance to make this possible in statistics processing side. With that, I'd say that let's keep the collector simple (only collecting the lock info) and go with https://github.com/pganalyze/collector/pull/327

> Additionally, as discussed, we should do a pgbench test to ensure this doesn't add noticable overhead (it shouldn't, since its only running pg_blocking_pids when there is a lock wait event already), but that would help us determine if we need to do any kind of caching in the initial implementation (assuming we keep this turned on by default, like done here).

Agreed, we should do some testing before merging. Will do that next week 👍

keiko713:
closing in favor of https://github.com/pganalyze/collector/pull/327

PR #327, opened 2022-10-07
Collect lock information with compact activity snapshot

keiko713:
With this PR, we collect lock information and add it to the compact activity snapshot. To collect lock information, we use `pg_blocking_pids()`.

This function is introduced PG9.6, so the lock information collection is only supported from PG9.6, as well as when `no-postgres-locks` option is not true (default is false, so people need to opt-out the lock info collection if they wish).

Related PR in snapshot side: https://github.com/pganalyze/collector-snapshot/pull/25

This PR also updates pq version from 1.3.0 to 1.10.7 (to support reading int32 array, introduced in v1.9.0). Here is the release note, it doesn't look like there is any breaking changes: https://github.com/lib/pq/releases

### TODO
- [x] run some testing to ensure this change won't add noticable overhead (see https://github.com/pganalyze/collector/pull/326#pullrequestreview-1133824215)
- [x] remove note from command flag (depending on the result of above testing, we may change the default) https://github.com/pganalyze/collector/blob/1ddeb5c191970a46786ceb7d8f5a4a5c9059215c/main.go#L274


msakrejda (approved):
I think this approach makes more sense to me than #326--I feel like the collector should be focused on gathering low-level information, and anything we can derive from that we should do in later processing steps, rather than inline with the collection, unless there are significant benefits to doing the latter.

input/postgres/backends.go
```diff
@@ -63,7 +77,8 @@ func GetBackends(logger *util.Logger, db *sql.DB, postgresVersion state.Postgres
 			&row.RoleOid, &row.RoleName, &row.Pid, &row.ApplicationName, &row.ClientAddr,
 			&row.ClientPort, &row.BackendStart, &row.XactStart, &row.QueryStart,
 			&row.StateChange, &row.Waiting, &row.BackendXid, &row.BackendXmin,
-			&row.WaitEventType, &row.WaitEvent, &row.BackendType, &row.State, &row.Query)
+			&row.WaitEventType, &row.WaitEvent, &row.BackendType, pq.Array(&row.BlockedByPids),
+			&row.State, &row.Query)
```
seanlinsley, line 81: It looks like [pg_blocking_pids](https://github.com/postgres/postgres/blob/9fd45870c1436b477264c0c82eb195df52bc0919/src/backend/utils/adt/lockfuncs.c#L429-L437) includes duplicates for parallel queries. Maybe we should remove those duplicates? (In Go since Postgres doesn't seem to have a good function to dedup an array)

keiko713: thanks for sharing that! given that the duplicates count won't big (only for parallel queries), also we are going to remove dups in the backend/frontend part when presenting, I think we don't need to focus on removing duplicates here, so I'm going to leave it as is 👍

PR #329, opened 2022-10-21
Introduce benchmark flag to support running collector without submitting stats

keiko713:
The new flag `--benchmark` allows you to run collector without submitting stats. With this, you can run collector without connecting to pganalyze, which will be helpful for benchmarking as you can skip the setup, as well as ignore the overhead of that part.

Difference between other flags:

* `test`: this only runs once, and also submit the stats to the server (setup is needed)
* `dry-run`: this also only runs once, and print the JSON data to stdout (which is not needed for benchmarking)

### DIFF
--- a/main.go
+++ b/main.go
@@ -284,7 +284,7 @@ func main() {
 	flag.StringVar(&configFilename, "config", defaultConfigFile, "Specify alternative path for config file")
 	flag.StringVar(&stateFilename, "statefile", defaultStateFile, "Specify alternative path for state file")
 	flag.StringVar(&pidFilename, "pidfile", "", "Specifies a path that a pidfile should be written to (default is no pidfile being written)")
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -283,6 +284,7 @@ func main() {
 	flag.StringVar(&configFilename, "config", defaultConfigFile, "Specify alternative path for config file")
 	flag.StringVar(&stateFilename, "statefile", defaultStateFile, "Specify alternative path for state file")
 	flag.StringVar(&pidFilename, "pidfile", "", "Specifies a path that a pidfile should be written to (default is no pidfile being written)")
+	flag.BoolVar(&benchmark, "benchmark", false, "Runs as a benchmark mode (skip submitting the statistics to the server)")
```
lfittl, line 287: I think "Runs in benchmark mode" or "Runs collector in benchmark mode" would be slightly more grammatically correct.

lfittl (approved):
:+1:

PR #331, opened 2022-10-21
Statements: Fix Postgres 15 compatibility due to version check bug

lfittl:
There was a bug in our version check mechanism that caused the minor version "1.10" to be treated as smaller than "1.8", effectively causing us to look for the wrong fields.

Fix the relevant version comparisons to use integer instead of floats, and in passing cleanup some old references to Postgres version numbers as compared to pg_stat_statements version numbers.

In passing add the version strings for Postgres 14 and 15, in case we add version-specific checks in the future.

### DIFF
--- a/input/postgres/statements.go
+++ b/input/postgres/statements.go
@@ -13,12 +13,18 @@ import (
 	"github.com/pganalyze/collector/util"
 )
 
-const statementSQLDefaultOptionalFields = "NULL, NULL, NULL, NULL, NULL"
-const statementSQLpg94OptionalFields = "queryid, NULL, NULL, NULL, NULL"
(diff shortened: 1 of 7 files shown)


msakrejda (approved):
This will fail once pg_stat_statements hits 2.0, but for now I was able to reproduce the problem in `main` and confirm that the patch fixes it. I think this is okay.

lfittl:
> This will fail once pg_stat_statements hits 2.0, but for now I was able to reproduce the problem in `main` and confirm that the patch fixes it. I think this is okay.

Fair point. I think given the lack of built-in extensions that use higher major versions, that is acceptable for now.

seanlinsley (approved):

PR #332, opened 2022-10-21
Packaging improvements

lfittl:


contrib/install.sh
```diff
@@ -94,11 +94,11 @@ then
   pkg=deb
   distribution=ubuntu
   version=$(grep VERSION_CODENAME /etc/os-release | cut -d= -f2)
-  if [ "$version" != focal ] && [ "$version" != bionic ] && [ "$version" != xenial ];
+  if [ "$version" != jammy ] && [ "$version" != focal ] && [ "$version" != bionic ] && [ "$version" != xenial ];
   then
     if confirm "Unsupported Ubuntu version; try Ubuntu Focal (20.04) package?";
```
msakrejda, line 99: We should change the string here as well.

contrib/install.sh
```diff
@@ -94,11 +94,11 @@ then
   pkg=deb
   distribution=ubuntu
   version=$(grep VERSION_CODENAME /etc/os-release | cut -d= -f2)
-  if [ "$version" != focal ] && [ "$version" != bionic ] && [ "$version" != xenial ];
+  if [ "$version" != jammy ] && [ "$version" != focal ] && [ "$version" != bionic ] && [ "$version" != xenial ];
```
msakrejda, line 97: The latest refresh of [Xenial went EOL about 18 months ago](https://endoflife.software/operating-systems/linux/ubuntu). No need to do anything now, but we should consider dropping it.

msakrejda (approved):
One small copy change required, but +1 otherwise.

PR #334, opened 2022-11-23
Collect Citus distributed index sizes

msakrejda:
Currently, when monitoring Citus databases, we collect distributed
table sizes with Citus' `citus_table_size` helper function, but we
ignore indexes on these distributed tables.

Add a similar calculation for index sizes. While there is a similar
`citus_relation_size` that is documented to work for distributed
indexes, it does not [1]. Instead, we use `run_command_on_shards` to
get this information from the underlying catalog objects and stats
directly.

[1]: I filed https://github.com/citusdata/citus/issues/6496 which
     seems to be getting some attention, so maybe there is hope for a
     better fix

### DIFF
--- a/input/postgres/relation_stats_ext.go
+++ b/input/postgres/relation_stats_ext.go
@@ -51,3 +51,80 @@ func handleRelationStatsExt(db *sql.DB, relStats state.PostgresRelationStatsMap,
 
 	return relStats, nil
 }
+
+const citusIndexSizeSQL = `
+WITH dist_idx_shard_stats_raw AS (
+	SELECT
+		dp.logicalrelid::regclass,
+		(pg_catalog.run_command_on_shards(logicalrelid::regclass::text, $$
+			SELECT
+				jsonb_agg(shard_idx_stats)
+			FROM (
+				SELECT
+					relnamespace::regnamespace AS idx_shard_schema,
+					indexrelid::regclass AS idx_shard_name,
+					COALESCE(pg_catalog.pg_relation_size(indexrelid), 0) AS idx_shard_bytes
+				FROM
+					pg_stat_user_indexes pgsui INNER JOIN pg_class pgc ON pgc.oid = pgsui.relid
+				WHERE
+					relid = '%s'::regclass
+			) AS shard_idx_stats
+		$$)).*
+	FROM
+		pg_dist_partition dp INNER JOIN pg_catalog.pg_class c ON (dp.logicalrelid::oid = c.oid)
+			 INNER JOIN pg_catalog.pg_namespace n ON (c.relnamespace = n.oid)
+	WHERE
+		($1 = '' OR (n.nspname || '.' || c.relname) !~* $1)
+), dist_idx_shard_stats AS (
+	SELECT
+		shardid,
+		bool_and(success) OVER() AS all_success,
+		jsonb_array_elements(result::jsonb) AS shard_info
+	FROM
+		dist_idx_shard_stats_raw
+)
+SELECT
+	pgc.oid,
+	sum((shard_info ->> 'idx_shard_bytes')::bigint) AS total_size_bytes
+FROM
+  pg_class pgc INNER JOIN dist_idx_shard_stats ON (
+		pgc.relkind = 'i'
+			AND pgc.relnamespace::regnamespace::text = (shard_info ->> 'idx_shard_schema')
+			AND pgc.relname = pg_catalog.regexp_replace(shard_info ->> 'idx_shard_name', '_' || shardid || '$', '')
+	)
+WHERE
+	all_success
+GROUP BY
+  oid;
+`
+
+func handleIndexStatsExt(db *sql.DB, idxStats state.PostgresIndexStatsMap, postgresVersion state.PostgresVersion, ignoreRegexp string) (state.PostgresIndexStatsMap, error) {
+	if postgresVersion.IsCitus {
+		stmt, err := db.Prepare(QueryMarkerSQL + citusIndexSizeSQL)
+		if err != nil {
+			return idxStats, fmt.Errorf("IndexStatsExt/Prepare: %s", err)
+		}
+		defer stmt.Close()
+
+		rows, err := stmt.Query(ignoreRegexp)
+		if err != nil {
+			return idxStats, fmt.Errorf("IndexStatsExt/Query: %s", err)
+		}
+		defer rows.Close()
+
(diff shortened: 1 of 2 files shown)


input/postgres/relation_stats_ext.go
```diff
@@ -51,3 +51,80 @@ func handleRelationStatsExt(db *sql.DB, relStats state.PostgresRelationStatsMap,
...
+		dist_idx_shard_stats_raw
+)
+SELECT
+	pgc.oid,
+	sum((shard_info ->> 'idx_shard_bytes')::bigint) AS total_size_bytes
+FROM
+  pg_class pgc INNER JOIN dist_idx_shard_stats ON (
+		pgc.relkind = 'i'
+			AND pgc.relnamespace::regnamespace::text = (shard_info ->> 'idx_shard_schema')
+			AND pgc.relname = pg_catalog.regexp_replace(shard_info ->> 'idx_shard_name', '_' || shardid || '$', '')
```
msakrejda, line 93: As far as I can tell, there's not a more direct way to match a shard index to its parent distributed index, so I'm relying on this approach.

input/postgres/relation_stats_ext.go
```diff
@@ -51,3 +51,80 @@ func handleRelationStatsExt(db *sql.DB, relStats state.PostgresRelationStatsMap,
...
 	return relStats, nil
 }
+
+const citusIndexSizeSQL = `
+WITH dist_idx_shard_stats_raw AS (
+	SELECT
+		dp.logicalrelid::regclass,
+		(pg_catalog.run_command_on_shards(logicalrelid::regclass::text, $$
+			SELECT
+				jsonb_agg(shard_idx_stats)
```
msakrejda, line 61: Because we need to run this once for each table, and because `run_command_on_shards` can only return a scalar value, pack that into jsonb to get info for all indexes at once for each table.

input/postgres/relation_stats_ext.go
```diff
@@ -51,3 +51,80 @@ func handleRelationStatsExt(db *sql.DB, relStats state.PostgresRelationStatsMap,
...
+			FROM (
+				SELECT
+					relnamespace::regnamespace AS idx_shard_schema,
+					indexrelid::regclass AS idx_shard_name,
+					COALESCE(pg_catalog.pg_relation_size(indexrelid), 0) AS idx_shard_bytes
```
msakrejda, line 66: Gathering other distributed stats here doesn't really make sense, right? They're not additive like size, and an average seems dubious as well...

msakrejda:
Hmm, test failed with

```
#6 [ 2/10] RUN apt-get update -qq && apt-get install -y -q build-essential git curl
#6 sha256:e98287d89549a4f13c89e1821b879e7c0f29d8f78283efde155eaefbe0a2766c
#6 4.302 W: The repository 'http://apt.postgresql.org/pub/repos/apt stretch-pgdg Release' does not have a Release file.
#6 4.302 E: Failed to fetch http://apt.postgresql.org/pub/repos/apt/dists/stretch-pgdg/main/binary-amd64/Packages  404  Not Found [IP: 87.238.57.227 80]
#6 4.302 E: Some index files failed to download. They have been ignored, or old ones used instead.
#6 ERROR: executor failed running [/bin/sh -c apt-get update -qq && apt-get install -y -q build-essential git curl]: exit code: 100
```

Looks like stretch has been [EOL since June](https://www.debian.org/releases/stretch/). Should we drop that?

input/postgres/relation_stats_ext.go
lfittl: Yeah, good question. For indexes specifically, it could be interesting to know "was this index scanned on any of the shards" to drive the unused index recommendations. Maybe we can also just aggregate that across the nodes? (i.e. `pg_stat_user_indexes.idx_scan`, and then we do our usual diffing logic on it)

As you note, averaging could tell the wrong story, so I wouldn't do other metrics (like cache hit ratio).

In the future, we may also want to collect some extra metrics for tables in the same style (e.g. total number of live rows across all shards), but I'd look at that on another day.

input/postgres/relation_stats_ext.go
lfittl: Makes sense, I think that works.

One thought that just crossed my mind: It may be interesting to consider switching this query to using `run_command_on_workers` in the future, and instead of doing one distributed query per table, doing a single distributed query for all tables, that returns the statistics from each worker node together.

But since our current implementation for table sizes already does one query per table (as `citus_table_size` does that under the hood), so this is matches that existing pattern, and I think its acceptable for most clusters.

msakrejda:
CI failures should be fixed by #335; will rebase once that lands.

input/postgres/relation_stats_ext.go
lfittl: Yeah, I think that seems like the right way to go.

lfittl (approved):
Nice, I think that looks good!

I've only reviewed the code and not tested this with an actual Citus cluster, but as I understand these APIs are very stable, and conceptually this looks correct to me.

msakrejda:
The `reload` test failed with

```
#9 226.1 E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/p/pinentry/pinentry-curses_1.1.0-3build1_amd64.deb  Connection failed [IP: 185.125.190.36 80]
#9 226.1 E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
#9 ERROR: executor failed running [/bin/sh -c apt-get update -qq && apt-get install -y -q build-essential git curl]: exit code: 100
------
 > [ 5/18] RUN apt-get update -qq && apt-get install -y -q build-essential git curl:
------
executor failed running [/bin/sh -c apt-get update -qq && apt-get install -y -q build-essential git curl]: exit code: 100
make[1]: *** [Makefile:211: reload] Error 1
make: *** [Makefile:39: integration_test] Error 2
``

This looks like it might be transient, so retrying...

input/postgres/relation_stats_ext.go
msakrejda: >For indexes specifically, it could be interesting to know "was this index scanned on any of the shards" to drive the unused index recommendations. Maybe we can also just aggregate that across the nodes?

I thought I had asked this, but it looks like I never submitted my comment or it got lost: "aggregate" as in sum? Or average?

lfittl: FWIW, I mean aggregate on that :)

PR #335, opened 2022-11-24
Drop Debian Stretch and obsolete Postgres versions from CI

msakrejda:
Debian Stretch has been EOL since June 2022, and is no longer
available as an official Docker image.

Skip building packages for this distribution.

Also, since some of the original Postgres images for older versions
were based on Stretch, this broke some of our CI integration tests.
Drop some of the obsolete images based on Stretch, and update others
to use the still-available bullseye version.


lfittl (approved):
:+1: Thanks for cleaning this up!

msakrejda:
Hmm, `test-reload` is failing with
```
#24 writing image sha256:4bffe3ce0bce5ef9e0d4f330d5ce9d6031bdf575f259428b6f769e6250151f78 done
docker run --name pganalyze-collector-test --mount type=bind,source=/sys/fs/cgroup,target=/sys/fs/cgroup --mount type=bind,source=/sys/fs/fuse,target=/sys/fs/fuse --mount type=tmpfs,destination=/tmp --mount type=tmpfs,destination=/run --mount type=tmpfs,destination=/run/lock -v `pwd`/docker-entrypoint-initdb.d:/docker-entrypoint-initdb.d/ -e POSTGRES_HOST_AUTH_METHOD=trust -d pganalyze-collector-test postgres -c pg_stat_statements.track_utility=off
#24 naming to docker.io/library/pganalyze-collector-test done
#24 DONE 3.0s
89fbdb6d5b338d99b9fa4675719129f74ccf850748d5e6aa70e5ea6bb74e585f
sleep 10
docker exec pganalyze-collector-test systemctl start postgresql
Failed to connect to bus: No such file or directory
make[1]: *** [Makefile:204: guided-setup] Error 1
make: *** [Makefile:39: integration_test] Error 2
Error: Process completed with exit code 2.
```

Not sure what's up. Will investigate.

lfittl:
@msakrejda Looking at this once more with fresh eyes in the morning, the last few commits I added here still seem correct. Might be good if you review them (since I authored them), and then merge if it looks good :)

integration_test/Makefile
```diff
@@ -252,24 +186,22 @@ pg15:
...
+	docker exec pganalyze-collector-test systemctl start pganalyze-collector
 	docker exec pganalyze-collector-test pganalyze-collector --reload
-	docker exec pganalyze-collector-test cat /tmp/collector.log | grep "Reloading configuration..."
+	docker exec pganalyze-collector-test journalctl -u pganalyze-collector | grep "Reloading configuration..."
 	docker exec pganalyze-collector-test test \! -f /var/lib/pganalyze-collector/state
 	docker exec pganalyze-collector-test bash -c "echo 'db_url = postgres://127.0.0.1/postgres' >> /etc/pganalyze-collector.conf"
-	docker exec --user pganalyze pganalyze-collector-test bash -c "echo '' > /tmp/collector.log"
 	docker exec pganalyze-collector-test pganalyze-collector --reload
 	docker exec pganalyze-collector-test pganalyze-collector --reload
-	docker exec pganalyze-collector-test cat /tmp/collector.log | grep "Reloading configuration..."
```
msakrejda, line 264: I'm guessing we're dropping this second grep because there's no way for us to reset the journal, so the grep might find the older message instead? In that case, do we still want to keep the reloads right above?

lfittl: Yeah, exactly, it was dropped because its difficult to reset the journal (possible, but messy)

The reloads are still needed because we are checking for the presence of the state file, which, as the comment at the beginning of the target notes, only gets created when you reload once a config is present (and that's why two reloads are needed, and this still tests something useful 😄)

msakrejda: Sounds good. Since you commented all the interesting pieces here, there's not much value to keeping the history, so I'll just squash this for clarity and merge it as one commit.

PR #339, opened 2022-11-30
Track queries using Postgres internal query ID

seanlinsley:
This PR adds an internal query ID cache to improve our ability to match query stats to the correct fingerprint, even if that query happens to churn out of `pg_stat_statements` by the time the next full snapshot rolls around.

This also adds improved tracking of truncated queries using `pg_stat_activity.query_id` in Postgres 14 and above.


state/postgres_statement.go
```diff
@@ -104,3 +104,11 @@ func (stmt DiffedPostgresStatementStats) Add(other DiffedPostgresStatementStats)
...
 	}
 }
+
+type QueryIdentity struct {
+	QueryID int64
+	Fingerprint uint64
+	LastSeen time.Time
+}
+
+type QueryIdentityMap map[int64]QueryIdentity
```
seanlinsley, line 114: It looks like query IDs are globally unique, since the relation ID of referenced tables is used when generating the fingerprint:

https://github.com/postgres/postgres/blob/fc7852c6cb89a5384e0b4ad30874de92f63f88be/src/backend/utils/misc/queryjumble.c#L285

The question is: are duplicate query IDs across databases (that were for entirely different queries) a big enough issue that we need to track which database the query ID came from in `QueryIdentityMap`?

runner/full.go
```diff
@@ -254,3 +257,27 @@ func CollectAllServers(servers []*state.Server, globalCollectionOpts state.Colle
...
+		return oldMap
+	}
+	slice := make([]state.QueryIdentity, 0, len(oldMap))
+	for _, identity := range oldMap {
+		slice = append(slice, identity)
+	}
+	sort.Slice(slice, func(i, j int) bool {
+		return slice[i].LastSeen.Before(slice[j].LastSeen)
+	})
+	for _, identity := range slice[:min(len(oldMap), 100000)] {
```
keiko713, line 272: will this `min(len(oldMap), 100000)` ever be `len(oldMap)` given that you have the early condition of `if len(oldMap) < 100000 {`?

seanlinsley: good catch, it looks like `len(oldMap)` will never be used

state/postgres_statement.go
msakrejda: Checking pg_stat_statements in our own database, it looks like we have ~0.15% dupes. That's pretty small, but maybe not small enough that we can just ignore it. In our case, though, all the dupes are in one database. What's the impact of collisions here? We may associate stats with the incorrect query text?

output/transform/util.go
```diff
@@ -82,10 +91,15 @@ func upsertQueryReferenceAndInformationSimple(server *state.Server, refs []*snap
 	idx := int32(len(refs))
 	refs = append(refs, &newRef)
 
+	normalizedQuery := util.NormalizeQuery(originalQuery, server.Config.FilterQueryText, trackActivityQuerySize)
+	if normalizedQuery == util.QueryTextTruncated {
+		normalizedQuery = ""
+	}
```
msakrejda, line 97: I don't follow this change to the truncated query handling. I see you replace the marker text with the empty string, but I don't understand what impact that has. Should we add a comment here? Or should this be happening in NormalizeQuery, and we can explain it there?

msakrejda (commented):
I'm not sure what to do about the risk of dupes here. The chance is very small, but the consequences seem potentially very confusing, and there's not a great way for us to diagnose this, right? I guess we could log warnings if `count(queryid)` differs from `count(distinct queryid)` but I don't think we can make them actionable. Or we could disable this behavior in those cases? Or should we make this configurable via collector setting? I think even then, we'd need to highlight collisions somehow, since this could get pretty confusing.

Thoughts?

output/transform/util.go
seanlinsley: >I see you replace the marker text with the empty string, but I don't understand what impact that has.

The internal PR this is paired with skips over queries whose query text is empty. It could just as well be a new query text placeholder to self-document the behavior, I just didn't want to send extra bytes that are going to be discarded.

>Or should this be happening in NormalizeQuery, and we can explain it there?

I kept it here to avoid changing the full snapshot behavior, which also uses `util.NormalizeQuery`

state/postgres_statement.go
seanlinsley: Yes, the stats could end up being associated with the wrong query in that case. But we only fall back to this identity map for queries that have churned out of `pg_stat_statements` since the last full snapshot so it may not be a big deal.

output/transform/util.go
msakrejda: Got it. I think it's okay to use the empty string as a sentinel, but maybe we should add a comment here? It's pretty hard to tell why this is flagged as empty text.

state/postgres_statement.go
msakrejda: So you think altogether it's unlikely enough that we can just ignore it? I'm a little reluctant to go with that (because if this does bite you, it will be basically impossible to figure out), but given the odds, it doesn't seem unreasonable, and I don't have a great idea on how to surface the problem. (And I don't think it's worth talking about it as a potential problem, since it's so unlikely).

output/transform/util.go
seanlinsley: Reviewing this again, I'm not really sure why I made this change. Since we're discussing elsewhere that we may want to stop storing queries using placeholder text, it may be better to keep this as `<truncated query>`

state/postgres_statement.go
lfittl: Trying to think through this:

1) Queries on two different databases having the same queryid, e.g. `SELECT 1` would have the same queryid, despite it being on different databases (since it doesn't involve any table OID that'd be database specific)

2) A hash collision with the Postgres `queryid`, i.e. two entirely unrelated queries in two different databases having the same queryid (if it was the same database we wouldn't know about it, since they'd just end up being a single entry in pgss)

If I understand the conversation so far, you're talking about (2) here, though I'm not sure if the 0.15% statistic that @msakrejda referenced is for actual hash collisions? (I'd be surprised, I'm assuming its more likely these are actually identical queries).

Assuming the hash collision case (2), what would happen today:

1. database A has queryid 1, and querytext X
2. database B also has queryid 1, but querytext Y
3. we loop through the stat statements entries, fingerprinting each based on the querytext, and end up with two distinct query fingerprints (and thus associate to the correct query entry in pganalyze)

Now assuming that in a future iteration of this PR we actually check the cache first and don't fingerprint if needed, the problem here would be that we are now associating to the wrong query in the case of the collision.

I'm 50/50 on whether this is a problem in practice. Hash collisions could occur (and the standard Postgres hash function used in pgss isn't the best hash function either), but it does seem a quite unlikely edge case, and we are not in a security sensitive context here, the way I look at it.

The one thing I would be worried about from a security perspective is in case we consider caching query text here as well (or if this mapping is relied on elsewhere to get the text), because doing a simple mapping here could cause query text to cross a security boundary (in the pganalyze app you may be restricted to viewing a single database on a server).

Maybe its better to just use the full triplet (i.e. `(database_id, role_id, queryid)`) for the mapping here, to avoid any accidental errors?

input/postgres/statements.go
```diff
@@ -234,6 +235,20 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
 		if showtext {
 			statementTexts[key] = receivedQuery.String
 		}
+		if queryID.Valid && showtext {
+			if server.PrevState.QueryIdentities == nil {
+				server.PrevState.QueryIdentities = make(state.QueryIdentityMap)
+			}
+			if identity, ok := server.PrevState.QueryIdentities[queryID.Int64]; ok {
+				identity.LastSeen = time.Now()
+			} else {
```
lfittl, line 244: It may be worth caching `time.Now()` at the beginning of the function, or using the collected at time - getting the time ~5000 times (or more, depending on query count and pg_stat_statements.max) isn't free, I think.

input/postgres/statements.go
```diff
@@ -234,6 +235,20 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
...
+				server.PrevState.QueryIdentities[queryID.Int64] = state.QueryIdentity{
+					QueryID:     queryID.Int64,
+					Fingerprint: util.FingerprintQuery(receivedQuery.String, server.Config.FilterQueryText, -1),
+					LastSeen:    time.Now(),
+				}
```
lfittl, line 249: We generally don't mutate `server.PrevState` in the input functions - is there a way to write this, that avoids this?

(maybe `QueryIdentities` shouldn't be in state after all, but rather be a separate variable on the `Server` struct? That way we could also have a separate mutex for it, making it easier to have concurrent activity snapshots access its information)

output/transform/util.go
```diff
@@ -61,8 +62,16 @@ func upsertQueryReferenceAndInformation(s *snapshot.FullSnapshot, statementTexts
...
-func upsertQueryReferenceAndInformationSimple(server *state.Server, refs []*snapshot.QueryReference, infos []*snapshot.QueryInformation, roleIdx int32, databaseIdx int32, originalQuery string, trackActivityQuerySize int) (int32, []*snapshot.QueryReference, []*snapshot.QueryInformation) {
-	fingerprint := util.FingerprintQuery(originalQuery, server.Config.FilterQueryText, trackActivityQuerySize)
+func upsertQueryReferenceAndInformationSimple(server *state.Server, refs []*snapshot.QueryReference, infos []*snapshot.QueryInformation, roleIdx int32, databaseIdx int32, queryID null.Int, originalQuery string, trackActivityQuerySize int) (int32, []*snapshot.QueryReference, []*snapshot.QueryInformation) {
+	var fingerprint uint64
+	if server.PrevState.QueryIdentities != nil && queryID.Valid {
+		if identity, ok := server.PrevState.QueryIdentities[queryID.Int64]; ok {
+			fingerprint = identity.Fingerprint
+		}
+	}
+	if fingerprint == 0 {
```
lfittl, line 72: Are you sure `0` isn't a valid hash value?

runner/full.go
```diff
@@ -67,6 +68,8 @@ func collectDiffAndSubmit(server *state.Server, globalCollectionOpts state.Colle
 		newState.StatementStats = transientState.ResetStatementStats
 	}
 
+	newState.QueryIdentities = pruneQueryIdentities(newState.QueryIdentities)
```
lfittl, line 71: As noted in a parallel comment, I think it may be better to put `QueryIdentities` directly on the `Server` struct, since it doesn't follow the usual prev state => new state handling, and is used outside of full snapshots.

output/transform/util.go
```diff
@@ -61,8 +62,16 @@ func upsertQueryReferenceAndInformation(s *snapshot.FullSnapshot, statementTexts
 	return idx
 }
 
-func upsertQueryReferenceAndInformationSimple(server *state.Server, refs []*snapshot.QueryReference, infos []*snapshot.QueryInformation, roleIdx int32, databaseIdx int32, originalQuery string, trackActivityQuerySize int) (int32, []*snapshot.QueryReference, []*snapshot.QueryInformation) {
-	fingerprint := util.FingerprintQuery(originalQuery, server.Config.FilterQueryText, trackActivityQuerySize)
+func upsertQueryReferenceAndInformationSimple(server *state.Server, refs []*snapshot.QueryReference, infos []*snapshot.QueryInformation, roleIdx int32, databaseIdx int32, queryID null.Int, originalQuery string, trackActivityQuerySize int) (int32, []*snapshot.QueryReference, []*snapshot.QueryInformation) {
+	var fingerprint uint64
+	if server.PrevState.QueryIdentities != nil && queryID.Valid {
+		if identity, ok := server.PrevState.QueryIdentities[queryID.Int64]; ok {
```
lfittl, line 68: I think this may be accessing server.PrevState without holding the `server.StateMutex` (since activity snapshots don't acquire that mutex, only the separate `ActivityStateMutex`)

lfittl (commented):
Overall looks like the right direction - there is a concurrency issue here with accessing the `server.PrevState` in an activity snapshot whilst not holding the `server.StateMutex`.

I think we should resolve that by just giving this its own variable and mutex on the `Server` struct, and then lock/unlock the mutex as needed.

state/postgres_statement.go
seanlinsley: >Now assuming that in a future iteration of this PR we actually check the cache first and don't fingerprint if needed, the problem here would be that we are now associating to the wrong query in the case of the collision.
>
>I'm 50/50 on whether this is a problem in practice. Hash collisions could occur (and the standard Postgres hash function used in pgss isn't the best hash function either), but it does seem a quite unlikely edge case, and we are not in a security sensitive context here, the way I look at it.

I implemented that in a followup PR: #346. I don't think hash collisions in this case are likely enough to warrant a more complex implementation at this point in time.

>Maybe its better to just use the full triplet (i.e. (database_id, role_id, queryid)) for the mapping here, to avoid any accidental errors?

FWIW it looks like composite key maps aren't as fast as we'd expect (from the #346 screenshot):

Screenshot 2022-12-16 at 3 29 04 PM

lfittl: Can you run a more isolated benchmark on the performance difference here?

I think the added complexity isn't an issue, since using the full triplet would match what we're doing elsewhere with `PostgresStatementKey` -- and doing it differently actually makes it more complex because now we have to remember when to use a simple queryId and when to use the full `PostgresStatementKey` (and as mentioned we can't just change things to always use `queryId` as the key).

seanlinsley: Changing to a composite key would necessarily mean increasing the cache size limit. The collector already has an unresolved memory consumption issue, and I worry that increasing the cache size further is going to make that worse.

seanlinsley: I'll work on building a standalone benchmark script when I'm back from holidays. There might be an argument to drop `PostgresStatementKey` across the board in favor of the query ID on its own.

lfittl: > Changing to a composite key would necessarily mean increasing the cache size limit.

With cache size limit you mean the 100k entries limit on the map?

I'm not sure if that'd need to be raised if the key also includes the database ID and role ID - the overall number of entries shouldn't actually increase much, unless we have hash collisions (which would no longer be the same entry), or the same query without table references runs on multiple databases.

Adding the role ID on the other hand may have such an effect (e.g. imagine a workload using short-lived roles), so there may be an argument for only doing `(database_id, queryid)` without the role ID.

> There might be an argument to drop PostgresStatementKey across the board in favor of the query ID on its own.

I don't think we can drop it across the board because of the aforementioned issue with the same queryid having different text on different databases, and that being security-relevant (e.g. imagine db1 is for tenant1, and has a query comment like `SELECT 1 /* user.email: tenant1@example.com */` and db2 has `SELECT 1 /* user.email: tenant2@example.com */`)

Therefore we'd need to keep the `PostgresStatementKey` use for any maps mapping to query texts at the very least, and anything related to statistics is also problematic (since there you need it to count each database separately).

msakrejda: >If I understand the conversation so far, you're talking about (2) here, though I'm not sure if the 0.15% statistic that @msakrejda referenced is for actual hash collisions? (I'd be surprised, I'm assuming its more likely these are actually identical queries).

They were queries with identical queryids occurring multiple times in `pg_stat_statements`. However, investigating further, it looks like this is all due to different records across separate `userid`s in `pg_stat_statements`. If I filter down to a single user, we have no collisions.

seanlinsley: > Therefore we'd need to keep the `PostgresStatementKey` use for any maps mapping to query texts at the very least, and anything related to statistics is also problematic (since there you need it to count each database separately).

Good point.

Am I misremembering, or had we previously discussed no longer distinguishing between user roles for queries, since it creates a confusing user experience? If so we could improve the map hashing time by using an array/slice (that's indexed into based on `DatabaseIdx`) which itself contains a map of the query ID, and whatever data is being stored.

https://gist.github.com/seanlinsley/dd7b2bf8d09b6ba710d27044794f86c5#file-map_test-go-L182

Screenshot 2022-12-28 at 11 41 50 AM

But that's all a tangent. For this PR specifically, do you think `QueryIdentityMap` needs to be scoped per database?

input/postgres/statements.go
seanlinsley: Would we write the cache to a separate file or skip persisting it? I could see arguments for both: saving it would make missing fingerprints much less likely if the collector restarts regularly (OOM), though if there was some undiscovered bug it at least wouldn't stick around across collector version updates.

output/transform/util.go
seanlinsley: How would I confirm that? `PgQuery.fingerprint('')` in Ruby generates `d8d13f8b2da6c9ad`

input/postgres/statements.go
lfittl: Good question - I think we could do it either way. Probably doesn't hurt to persist it.

If we wanted to persist it, we could simply add it to the `StateOnDisk` struct and update `WriteStateFile` and `ReadStateFile` accordingly - i.e. still keep it as a separate value on the `Server` struct itself (not have it inside state, per the above mentioned mutex issue).

output/transform/util.go
lfittl: I think thats really a [XXH3](https://github.com/Cyan4973/xxHash) question - and since I couldn't find any particular documentation that states `0` is not a possible digest result, I think we should assume it could be.

seanlinsley: Even if it's theoretically possible, surely it wouldn't be a valid query?

But let's say it's a needed change. How would we have a null type here when numbers in Go default to zero? Use a `NullInt`?

lfittl: > Even if it's theoretically possible, surely it wouldn't be a valid query?

Why couldn't it be a valid query? I don't know the necessary tokens that'd need to be the input to the fingerprint function, but it's possible that the right combination of token values produces a hash value of `0` - it's just as likely as producing e.g. a value of `42` or `123`.

> But let's say it's a needed change. How would we have a null type here when numbers in Go default to zero? Use a NullInt? Add a local boolean variable to check if it's already been set?

I think in this particular case, if we just need to know inside the function, we could just use a local boolean variable that indicates whether we found any fingerprint.

seanlinsley:
Okay, this is ready for another round of review. Note I used a `RWMutex` so compact snapshots can read from the map without blocking (since only the full snapshot updates it).

input/postgres/statements.go
```diff
@@ -234,6 +236,22 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
 		if showtext {
 			statementTexts[key] = receivedQuery.String
 		}
+		if queryID.Valid && showtext {
+			server.QueryIdentitiesMutex.Lock()
```
lfittl, line 240: Two thoughts:

1) Should we be holding the mutex for the full loop, not each iteration? (the trade-off here is that we block other users for longer, but are reducing the per-loop overhead slightly)

2) If we are holding it separately for each iteration, should we first acquire the mutex for reading, and only elevate the mutex to writing if we didn't find the query ID? (that would be best in terms of letting others read too, but I'm not sure if that's necessary / helpful)

input/postgres/statements.go
```diff
@@ -234,6 +236,22 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
...
+			if server.QueryIdentities == nil {
+				server.QueryIdentities = make(state.QueryIdentityMap)
+			}
```
lfittl, line 243: Can we do this initialization once somewhere (possibly there is a place where we already initialize other server structs?), instead of rechecking for it in each loop iteration?

input/postgres/statements.go
```diff
@@ -234,6 +236,22 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
...
+			if identity, ok := server.QueryIdentities[queryID.Int64]; ok {
+				identity.LastSeen = now
+			} else {
+				server.QueryIdentities[queryID.Int64] = state.QueryIdentity{
+					QueryID:     queryID.Int64,
```
lfittl, line 248: Don't we already have the queryID in the key? (since storage is at a premium here, it seems better to keep the struct as small as possible)

output/transform/util.go
```diff
@@ -61,8 +62,20 @@ func upsertQueryReferenceAndInformation(s *snapshot.FullSnapshot, statementTexts
 	return idx
 }
 
-func upsertQueryReferenceAndInformationSimple(server *state.Server, refs []*snapshot.QueryReference, infos []*snapshot.QueryInformation, roleIdx int32, databaseIdx int32, originalQuery string, trackActivityQuerySize int) (int32, []*snapshot.QueryReference, []*snapshot.QueryInformation) {
-	fingerprint := util.FingerprintQuery(originalQuery, server.Config.FilterQueryText, trackActivityQuerySize)
+func upsertQueryReferenceAndInformationSimple(server *state.Server, refs []*snapshot.QueryReference, infos []*snapshot.QueryInformation, roleIdx int32, databaseIdx int32, queryID null.Int, originalQuery string, trackActivityQuerySize int) (int32, []*snapshot.QueryReference, []*snapshot.QueryInformation) {
+	var fingerprint uint64
+	var fingerprintFound bool
+	if server.QueryIdentities != nil && queryID.Valid {
```
lfittl, line 68: I think in order to read `server.QueryIdentities` here without holding the mutex we'd have to ensure it is set outside of a mutex (i.e. when a server struct is created) - right now this could theoretically be a bad pointer read (though maybe setting a pointer is an atomic operation in Go?), e.g. when an activity snapshot calls this whilst a full snapshot just starts the GetStatements loop.

runner/full.go
```diff
@@ -254,3 +257,31 @@ func CollectAllServers(servers []*state.Server, globalCollectionOpts state.Colle
 
 	return
 }
+
+func pruneQueryIdentities(server *state.Server) {
+	max := 100000
+	if len(server.QueryIdentities) < max {
```
lfittl, line 263: I think we need to hold the mutex here already (at least for reads).

state/postgres_statement.go
```diff
@@ -104,3 +104,11 @@ func (stmt DiffedPostgresStatementStats) Add(other DiffedPostgresStatementStats)
 		BlkWriteTime:      stmt.BlkWriteTime + other.BlkWriteTime,
 	}
 }
+
+type QueryIdentity struct {
+	QueryID     int64
```
lfittl, line 109: I don't think this field is necessary - are we using it anywhere? (its already in the key of the map)

state/postgres_statement.go
```diff
@@ -104,3 +104,11 @@ func (stmt DiffedPostgresStatementStats) Add(other DiffedPostgresStatementStats)
...
+	Fingerprint uint64
+	LastSeen    time.Time
```
lfittl, line 111: I was surprised to see that the size of time.Time is 24 bytes (https://stackoverflow.com/questions/56963226/how-many-bytes-are-in-a-golang-time-object), making this pretty expensive - maybe we can instead keep a uint64 as a unix timestamp field?

input/postgres/statements.go
seanlinsley: This `rows.Next()` loop takes at least 400ms per the screenshot in #346. It's hard to say which approach is better, I had just chosen this option to avoid moving `Lock()` to line 209 and `Unlock` to line 260.

Using a read-only lock doesn't make sense here because `LastSeen` is always updated.

input/postgres/statements.go
seanlinsley: Yes, I think it makes sense to move it to the `state.Server{...}` struct construction in `main.go`

input/postgres/statements.go
seanlinsley: Good point, there's no need to store the query ID twice.

output/transform/util.go
seanlinsley: I'll update the code so the identity map always exists, so this nil check isn't needed.

state/postgres_statement.go
seanlinsley: (marking as resolved since there's the same comment elsewhere)

state/postgres_statement.go
seanlinsley: Good point, will update.

input/postgres/statements.go
```diff
@@ -234,12 +237,23 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
...
+				}
+			}
+		}
 		if ignoreIOTiming(postgresVersion, receivedQuery) {
 			stats.BlkReadTime = 0
 			stats.BlkWriteTime = 0
 		}
 		statementStats[key] = stats
 	}
+	server.QueryIdentitiesMutex.Unlock()
```
lfittl, line 256: Thanks for moving these around. I think that looks good, however we also need to make sure to unlock the mutex in error cases where we do an early return out of the loop (i.e. line 222).

lfittl (approved):
Nice - I think this looks good!

One more minor comment re: a missing mutex unlock, but otherwise I think this is good to ship.

input/postgres/statements.go
seanlinsley: Ah, good catch

seanlinsley:
Due to the high rate of churn, this query ID cache is ineffective in practice. Reducing churn requires changes to how applications write their queries. We've added a message in-app to explain what customers can do:

Screenshot 2023-06-22 at 11 23 53 AM

PR #340, opened 2022-12-02
Bump pg_query_go version to 2.2.0

keiko713:
The `pg_query_go` 2.2.0 version includes the fix with the build issue with macOS SDK 13.0.
Bumping the version here too so that people won't get hit by this.

https://github.com/pganalyze/pg_query/issues/263


lfittl (approved):
:+1:

PR #341, opened 2022-12-08
Add `always_collect_system_data` config option

msepga:
Now we try to read the value of `always_collect_system_data` in the config file, as well as the original `PGA_ALWAYS_COLLECT_SYSTEM_DATA` environment variable.

This is useful for ~~non-container setups which connect to remote servers~~ package-based setups which monitor the local server by a non-local IP.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -37,10 +37,10 @@ type ServerConfig struct {
 	ErrorCallback   string `ini:"error_callback"`
 	SuccessCallback string `ini:"success_callback"`
 
-	EnableReports           bool `ini:"enable_reports"`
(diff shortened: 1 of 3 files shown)


lfittl:
> This is useful for non-container setups which connect to remote servers.

I believe that was meant to say "This is useful for package-based setups which monitor the local server by a non-local IP." (since the point of this is that we don't check if a particular private IP is local to the machine, instead we just check against `127.0.0.1`/`localhost`/etc -- this serves as a way to ensure the server is considered "local")

lfittl (approved):
:+1:

PR #342, opened 2022-12-14
Collect TXID related data

keiko713:
🍐 PR snapshot side: https://github.com/pganalyze/collector-snapshot/pull/26

We will start collecting the following data:

* server level: current TXID and next MXID
* database level: age of `datfrozenxid` and `datminmxid`, also `xact_commit` and `xact_rollback` (the diff from the prev snapshot)
* relation/table level: the age of `relfrozenxid` and `relminmxid`

Will leave PR comments for more details.

I'm dropping the logic that supports < PG 9.6 as we plan to retire these logics in https://github.com/pganalyze/collector/pull/343


input/postgres/databases.go
```diff
@@ -8,60 +8,75 @@ import (
 	"github.com/pganalyze/collector/util"
 )
 
-const databasesSQLDefaultOptionalFields = "1"
-const databasesSQLpg93OptionalFields = "datminmxid"
+const databasesSQLDefaultMxidFields = "1"
+const databasesSQLpg93MxidFields = "d.datminmxid"
+const databasesSQLDefaultMxidAgeFields = "0"
+const databasesSQLpg93MxidAgeFields = "CASE WHEN d.datminmxid <> '0' THEN mxid_age(d.datminmxid) ELSE 0 END"
```
keiko713, line 14: [note for reviewer]
I kept the "age" as 0 when
1. the data is not retrievable (before pg 9.3)
2. min xid or mxid is 0

I've seen the pattern that when min (frozen) xid is zero, the age is becoming some big number. This value being zero actually has a different meaning ([doc of pg_class](https://www.postgresql.org/docs/current/catalog-pg-class.html)):

> relfrozenxid: Zero (InvalidTransactionId) if the relation is not a table.

Technically, `datfrozenxid` is the minimum of the per-table `relminmxid` values, so it's less likely gonna be `0` (totally possible with `pg_class` side, but just not `pg_database` side), but adding that to be consistent with the relation side.

input/postgres/relations.go
```diff
@@ -10,8 +10,10 @@ import (
 	"github.com/pganalyze/collector/state"
 )
 
-const relationsSQLDefaultOptionalFields = "0"
-const relationsSQLpg93OptionalFields = "c.relminmxid"
+const relationsSQLDefaultMxidFields = "1"
```
keiko713, line 13: [note for reviewer]
This default value was previously `0`, but I'm going to change it to `1`, as I've been observing that it is actually the right default value. The database side was already having 1 as a default value, interestingly.

input/postgres/replication.go
```diff
@@ -76,12 +76,49 @@ SELECT client_addr,
 	FROM %s
  WHERE client_addr IS NOT NULL`
 
+const transactionIdSQLPg13 string = `
+SELECT
+	pg_current_xact_id(),
+	next_multixact_id
+FROM pg_catalog.pg_control_checkpoint()
```
keiko713, line 83: [note for reviewer]
The value of `pg_catalog.pg_control_checkpoint()` is only updated per checkpoint. The checkpoint runs every 5 mins (or wal size reaching to 1GB), so it's possible that it has pretty up-to-date value with full snapshot (runs every 10 mins). However, it is also common that the checkpoint timeout is set longer than 10 mins, and in that case, the `next_multixact_id` information can be stale.
As this value is more experimental one, we will accept this staleness and just treat it as "better than nothing".

input/postgres/replication.go
```diff
@@ -76,12 +76,49 @@ SELECT client_addr,
 	FROM %s
  WHERE client_addr IS NOT NULL`
 
+const transactionIdSQLPg13 string = `
+SELECT
+	pg_current_xact_id(),
```
keiko713, line 81: [note for reviewer]
`pg_current_xact_id()` will assign a new transaction ID if the current transaction does not currently have one. This means, it'll likely consume/assign a new transaction ID. There would be an argument of consuming such thing, but full snapshot only runs every 10 mins and it will likely be a small increase of transaction ID (144 per day) which will very less likely to be any issues to the Postgres server, so we choose to use this function.

input/postgres/replication.go
```diff
@@ -76,12 +76,51 @@ SELECT client_addr,
...
+
 func GetReplication(logger *util.Logger, db *sql.DB, postgresVersion state.PostgresVersion, systemType string) (state.PostgresReplication, error) {
 	var err error
 	var repl state.PostgresReplication
 	var sourceTable string
 	var replicationStandbySQL string
 	var replicationSQL string
+	var transactionIdSQL string
+
+	// Get Postgres Server stats first (even with Aurora, as long as it's not a replica)
```
keiko713, line 107: [note for reviewer]
Since `pg_current_xact_id()` can't be run in replicas (will return `ERROR:  cannot execute pg_current_xact_id() during recovery`), we should avoid running this if it's replica.
Previously, we were not collecting any replication info with Aurora, because of `Function pg_last_xlog_receive_location() is currently not supported for Aurora` error. However, there should be no problem getting the current xact info with Aurora, so not going to skip that part here (tested with Aurora that it works).

msakrejda (commented):
Looks good, but not sure about putting current xid and mxid under replication. That seems confusing, since these are relevant even without any replication happening. Thoughts?

input/postgres/databases.go
```diff
@@ -8,60 +8,75 @@ import (
...
+	d.datcollate,
+	d.datctype,
+	d.datistemplate,
+	d.datallowconn,
+	d.datconnlimit,
+	d.datfrozenxid,
+	%s,
+	CASE WHEN d.datfrozenxid <> '0' THEN age(d.datfrozenxid) ELSE 0 END,
+	%s,
+	(sd.xact_commit + sd.xact_rollback) AS transaction_count
```
lfittl, line 32: I think it'd be good if we keep these as separate counters in the collector and snapshot data - we may want to show them as separate graphs at some point, and it doesn't cost much in terms of snapshot size.

(we can still store them as one combined number in the application data model itself, but do that combination on the server side instead of collector side)

input/postgres/replication.go
keiko713: This has moved to server stats

input/postgres/databases.go
keiko713: We're not storing the transaction count itself to the pganalyze app, we're using this data to calculate TPS within the collector. Though, I've added `xact_commit` and `xact_rollback` separately to the message of database stats, and I'll simply not store these data in the pganalyze app side for now.

input/postgres/server_stats.go
```diff
@@ -0,0 +1,53 @@
...
+	txid_current(),
+	1
+`
+
+func GetServerStats(logger *util.Logger, db *sql.DB, postgresVersion state.PostgresVersion, systemType string) (state.PostgresServerStats, error) {
+	var stats state.PostgresServerStats
+	var transactionIdSQL string
+
+	// Only collect xact ID related stats with non-replicas
+	if isReplica, err := getIsReplica(db); err == nil && !isReplica {
```
lfittl, line 35: Not a big deal, but we could save a query if we get this once and remember it for both server stats and replication stats - similar to how we handle Postgres versions (they are fetched at the beginning of the input function).

output/transform/postgres.go
```diff
@@ -93,6 +94,19 @@ func transformPostgresDatabases(s snapshot.FullSnapshot, transientState state.Tr
...
 		s.DatabaseInformations = append(s.DatabaseInformations, &info)
+
+		diff, exist := diffState.DatabaseStats[database.Oid]
+		if exist {
+			stat := snapshot.DatabaseStatistic{
+				DatabaseIdx:        databaseOidToIdx[database.Oid],
+				FrozenXidAge:       database.FrozenXIDAge,
+				MinMxidAge:         database.MinMXIDAge,
+				XactCommit:         database.XactCommit,
+				XactRollback:       database.XactRollback,
```
lfittl, line 105: I think these also need to come out of the `diff.`, similar to TransactionsPerSecond.

state/postgres_database.go
```diff
@@ -21,4 +21,33 @@ type PostgresDatabase struct {
...
+type PostgresDatabaseStats struct {
+	TransactionCount uint64 // Transaction count of this database
+}
+
+// PostgresDatabaseStatsMap - Map of database statistics (key = database Oid)
+type PostgresDatabaseStatsMap map[Oid]PostgresDatabaseStats
+
+// DiffedPostgresDatabaseStat - Database statistics for a single database as a diff
+type DiffedPostgresDatabaseStats struct {
+	TransactionsPerSecond uint64
```
lfittl, line 42: I think we should also diff `XactCommit` and `XactRollback` the same way we do for TPS.

state/postgres_database.go
```diff
@@ -21,4 +21,33 @@ type PostgresDatabase struct {
...
+}
+
+// DiffedDatabaseStats - Map of diffed database statistics (key = database Oid)
+type DiffedPostgresDatabaseStatsMap map[Oid]DiffedPostgresDatabaseStats
+
+// DiffSince - Calculate the diff between two stats runs
+func (curr PostgresDatabaseStats) DiffSince(prev PostgresDatabaseStats, collectedIntervalSecs uint32) DiffedPostgresDatabaseStats {
+	return DiffedPostgresDatabaseStats{
+		TransactionsPerSecond: (curr.TransactionCount - prev.TransactionCount) / uint64(collectedIntervalSecs),
```
lfittl, line 51: Whilst we are using this pattern for the system statistics (to divide by `collectedIntervalSecs` in the collector), we don't do this for any other statistics - typically we've done this division on the application side, and just passed along the difference between the counter values. For system statistics its because some providers already give us a "per second" value, so we normalize all providers to work that way.

I have a preference to not the division here (i.e. not divide by collectedIntervalSecs as this is doing right now), because its sometimes helpful for debugging to see the difference value that a snapshot calculated - vs in this case its likely there is some rounding due to this devision, making debugging harder.

(e.g. when a counter wraps around, there are sometimes details that can be understood from the actual diff value between two full snapshot runs)

lfittl (approved):
Looks good - some minor comments, but assuming these are addressed we could merge this I think.

state/postgres_database.go
keiko713: Ah wait, I think I'm confused. So you wanted to have the diff, not the total count? If so, maybe it's better to clarify it in the naming and/or comment.

keiko713: Oh, actually, looking the code more, looks like others are not using the diff value also, but passing the absolute value and do the whole calculation by using the `snapshot.collected_interval_secs` + getting the last snapshot 🤔  Maybe you meant this?
I'll hold on tweaking this until the further clarification.

lfittl: Generally, the way it works today for most statistics, such as table statistics ("tuples updated", etc):

* First full snapshot:
  * We collect data point A
  * We don't send any data, since there is nothing to diff against
  * On the server side we don't see any data yet (since the diff was not done)
* Second full snapshot (~600 seconds later)
  * We collect data point B
  * We calculate "B - A" and send this as the diff to the server (this represents the activity that happened since the last full snapshot)
  * On the server side, we get the diff and then divide it by the time since the last snapshot (~600 seconds) to get a per second value

We do not pass in the raw values of B or A, since they don't really tell us anything - only the diff between the two values is meaningful, since it represents the amount of activity within a known time interval (i.e. between two full snapshots).

There is two things that I wanted to comment on in the context of this PR:
- I think it would be good to know "how many commits happened" separately from "how many rollbacks happened" (since it allows us to graph that in the future, even if for now we mostly care about the sum of both to understand TXID consumption) -- your initial code here sent the raw values (i.e. A or B), not the diff between two snapshots, i.e. B - A)
- To align with the other collector statistics code (with the exception of system statistics), I think we should not make the calculation of "transactions per second" on the collector side, but on the server side - and instead only send the difference in how many transactions occurred between snapshots

Does that clarify things? (happy to chat more on Slack/etc)

keiko713: Ohhh thanks for explaining! I didn't understand this part at all (TIL schema stats are holding diff info, not the total count)!
I think that's also why I'm missing the change to the `PostgresRelationStats` in this PR, the table level stats change should have been made there, not `PostgresRelation` (well technically `frozen_xid_age` doesn't need to be stored in `PersistedState` so it could still be there, but looks like we're doing it anyways with the stats like `NLiveTup` so we can put in `PostgresRelationStats`).

> To align with the other collector statistics code (with the exception of system statistics)

by now, I think it's clear that I was copy&pasting from the wrong place 🙃

input/postgres/server_stats.go
keiko713: Very good point. Currently `GetIsReplica` actually returns more than "if it's replica", therefore I had to introduce `getIsReplica`.
I think we ideally want to save this query indeed (and actually `GetIsAwsAurora` too, which is used in `GetIsReplica`), but it comes with "naming is hard" problem so I'm gonna just leave it as is this time.

keiko713:
@lfittl updated quite a bit based on the snapshot change, can you give me one more 👀 before merging when you have time?

input/full.go
```diff
@@ -81,6 +81,12 @@ func CollectFull(server *state.Server, connection *sql.DB, globalCollectionOpts
 		err = nil
 	}
 
+	ts.ServerStats, err = postgres.GetServerStats(logger, connection, ts.Version, systemType)
+	if err != nil {
+		logger.PrintWarning("Error collecting Postgres server statistics: %s", err)
+		return
```
lfittl, line 87: If we consider it a warning, we should probably drop the return (so the snapshot can be finished and sent) - otherwise we should make it an error to match existing code.

The downside of outputting a log message here but continuing is that you will also get a successful test run. If the error is very unlikely its probably best to use PrintError here.

state/postgres_server_stats.go
```diff
@@ -0,0 +1,7 @@
+package state
+
+// PostgresServerStats - Statistics for a Postgres server.
+type PostgresServerStats struct {
+	CurrentXactId   Xid
+	NextMultiXactId Xid
```
lfittl, line 6: Both of these should be 64-bit types (we could invent a new Xid8 type for use here, to match Postgres)

lfittl: Or maybe Xid32 for the existing Xid, and Xid64 for this here?

output/transform/postgres.go
```diff
@@ -59,7 +60,7 @@ func transformPostgresRoles(s snapshot.FullSnapshot, transientState state.Transi
 	return s, roleOidToIdx
 }
 
-func transformPostgresDatabases(s snapshot.FullSnapshot, transientState state.TransientState, roleOidToIdx OidToIdx) (snapshot.FullSnapshot, OidToIdx) {
+func transformPostgresDatabases(s snapshot.FullSnapshot, newState state.PersistedState, diffState state.DiffState, transientState state.TransientState, roleOidToIdx OidToIdx) (snapshot.FullSnapshot, OidToIdx) {
```
lfittl, line 63: I don't think the newState argument is used anymore (probably a leftover from an earlier version of the code)

output/transform/postgres.go
```diff
@@ -93,6 +94,18 @@ func transformPostgresDatabases(s snapshot.FullSnapshot, transientState state.Tr
 		}
 
 		s.DatabaseInformations = append(s.DatabaseInformations, &info)
+
+		diff, exist := diffState.DatabaseStats[database.Oid]
```
lfittl, line 98: It might be slightly clearer to understand the ages below not actually being a diff if we call this variable `stats` (instead of diff).

lfittl (approved):
Found a few more minor details, but I think once these are fixed we can ship it! 🚢

PR #343, opened 2022-12-15
Drop support for old Postgres versions

msakrejda:
In 06daccdacb25a1c8f37df472abd31c67f14df10a, we dropped CI testing for
Postgres versions older than than 9.6 (which itself has been EOL for over a
year).

Drop support in code to allow us to simplify some logic. Users who
need support for older Postgres releases can use an older collector
for the time being.


msakrejda:
@keiko713 yeah, I realized we should probably do this when I saw you were accounting for 9.3 in #342. And you're right--it looks like there's some README cleanup (we mention 9.3 there). I'll do that once this gets the green light.

state/postgres_version.go
```diff
@@ -13,8 +10,8 @@ const (
 	PostgresVersion14 = 140000
 	PostgresVersion15 = 150000
 
-	// MinRequiredPostgresVersion - We require PostgreSQL 9.3 or newer
-	MinRequiredPostgresVersion = PostgresVersion93
+	// MinRequiredPostgresVersion - We require PostgreSQL 9.6 or newer
+	MinRequiredPostgresVersion = PostgresVersion96
```
lfittl, line 14: There is a hard-coded "9.3 or newer is required." message on the top of `input/full.go` that also needs to be updated.

lfittl (approved):
:+1: Looks good, with two minor comments.

Overall, I think dropping 9.5 and older support is clearly the right choice - those versions are long gone.

9.6 and 10 there are still some late upgraders, so doesn't hurt to keep them around for a bit longer and then drop them when we either have a good reason, or more time has passed.

input/postgres/backends.go
```diff
@@ -35,10 +33,6 @@ func GetBackends(logger *util.Logger, db *sql.DB, postgresVersion state.Postgres
 
 	if postgresVersion.Numeric >= state.PostgresVersion10 {
 		optionalFields = activitySQLpg10OptionalFields
-	} else if postgresVersion.Numeric >= state.PostgresVersion96 {
```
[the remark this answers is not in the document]
msakrejda: It's not super clear from the diff, but 9.6 is now default, so that's the `else` branch.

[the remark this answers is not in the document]
lfittl: Ah, thanks for pointing that out - my bad :)

msakrejda:
I pushed an update cleaning up other mentions of older versions. The last commit gets rid of support for parsing pre-9.6 log lines (at least those that were documented as such). Let me know if we should be more conservative and I can just drop that commit, but I checked 9.6 xlog.c and vacuumlazy.c, and those parts of the messages are always present.

state/postgres_statement.go
```diff
@@ -20,34 +20,32 @@ type PostgresStatement struct {
...
+	LocalBlksDirtied  int64      // Total number of local blocks dirtied by the statement
+	LocalBlksWritten  int64      // Total number of local blocks written by the statement
+	TempBlksRead      int64      // Total number of temp blocks read by the statement
+	TempBlksWritten   int64      // Total number of temp blocks written by the statement
+	BlkReadTime       float64    // Total time the statement spent reading blocks, in milliseconds (if track_io_timing is enabled, otherwise zero)
+	BlkWriteTime      float64    // Total time the statement spent writing blocks, in milliseconds (if track_io_timing is enabled, otherwise zero)
+	MinTime           null.Float // Minimum time spent in the statement, in milliseconds
+	MaxTime           null.Float // Maximum time spent in the statement, in milliseconds
+	MeanTime          null.Float // Mean time spent in the statement, in milliseconds
+	StddevTime        null.Float // Population standard deviation of time spent in the statement, in milliseconds
```
lfittl, line 41: Should we turn these into regular `float64` values now? (or is there ever a case where they are NULL on 9.5+?)

state/postgres_statement.go
```diff
@@ -20,34 +20,32 @@ type PostgresStatement struct {
...
 }
 
 // PostgresStatementKey - Information that uniquely identifies a query
 type PostgresStatementKey struct {
 	DatabaseOid Oid   // OID of database in which the statement was executed
 	UserOid     Oid   // OID of user who executed the statement
-	QueryID     int64 // Postgres 9.4+: Internal hash code, computed from the statement's parse tree
+	QueryID     int64 // Internal hash code, computed from the statement's parse tree
```
lfittl, line 48: We have a fallback logic for calculating the QueryID in the statements input function (based on the query text) that we can probably drop now as well. In case we don't get a queryid a hard error is probably better than some special logic (we hash the text currently in such cases).

logs/analyze.go
```diff
@@ -1619,9 +1614,8 @@ func classifyAndSetDetails(logLine state.LogLine, statementLine state.LogLine, d
 			numIndexScans, _ := strconv.ParseInt(parts[4], 10, 64)
 			pagesRemoved, _ := strconv.ParseInt(parts[5], 10, 64)
 			relPages, _ := strconv.ParseInt(parts[6], 10, 64)
-
-			// Parts 7 and 8 (Pinskipped/Frozenskipped pages) are only present on Postgres 9.5/9.6+ and dealt with at the end
-
+			pinskippedPages, _ := strconv.ParseInt(parts[7], 10, 64)
+			frozenskippedPages, _ := strconv.ParseInt(parts[8], 10, 64)
```
lfittl, line 1618: We can drop this logic later as well:

```
			if parts[7] != "" {
				pinskippedPages, _ := strconv.ParseInt(parts[7], 10, 64)
				logLine.Details["pinskipped_pages"] = pinskippedPages
			}
			if parts[8] != "" {
				frozenskippedPages, _ := strconv.ParseInt(parts[8], 10, 64)
				logLine.Details["frozenskipped_pages"] = frozenskippedPages
			}
```

lfittl (approved):
Thanks for cleaning that up!

logs/analyze.go
msakrejda: Right, of course.

state/postgres_statement.go
msakrejda: 9.5+ should work. If someone is using a super-old un-upgraded 1.2 from Postgres 9.4, this would break. It's probably okay to drop support for that, too.

msakrejda: I changed these to raw floats, and to my surprise, nothing broke. Do we actually [do anything with these](https://github.com/pganalyze/collector/blob/main/output/transform/postgres_statements.go#L46-L66) right now?

keiko713: ... doesn't look like it's used at all? 🤔

msakrejda: Got it. So we could do something like plot the latest average, but that's probably not very useful because right now it becomes more and more weighted toward the past: the number grows in inertia and will stop responding meaningfully to even dramatic changes, right?

msakrejda:
I think this is ready to go now, but I can wait until #339 and #342 land to avoid causing unnecessary conflicts.

state/postgres_version.go
lfittl: Re-using this thread - On another note, I just realized we should probably raise this to Postgres 10 being the minimum - because that was the first version to use $n style replacement characters in pg_stat_statements.

See https://github.com/pganalyze/collector/pull/383, which will effectively make Postgres 9.6 not work with the collector (since we fail to parse / normalize the statements containing "?").

msakrejda: Makes sense. I'll update once #339 lands.

README.md
```diff
@@ -128,25 +116,6 @@ $$
     FROM pg_sequences WHERE schemaname = schema_name AND sequencename = sequence_name;
 $$ LANGUAGE sql VOLATILE SECURITY DEFINER;
 
---- For Postgres 9.6 and older, use this:
```
keiko713, line 131: There is "--- The following is needed for Postgres 10+:" above that we can remove.

lfittl (approved):
:+1: Re-reviewed and looks good

README.md
msakrejda: Eh, thinking about this, we presumably still need that (on all versions now) if you were to run the sequence report. Which we no longer really do. I think the README needs some TLC anyway (our docs on [pganalyze.com](https://pganalyze.com) are much more up-to-date), so I think I'll just leave as-is for now.

PR #346, opened 2022-12-16
Use query ID cache to skip query fingerprint calculation

seanlinsley:
Query fingerprinting takes around half of `GetStatements` runtime (0.9 of 1.9 seconds in the below pprof screenshot). Since #339 adds a query_id/fingerprint cache, we can rely on that to skip this expensive step.

Screenshot 2022-12-16 at 9 23 40 AM

### DIFF
--- a/input/postgres/statements.go
+++ b/input/postgres/statements.go
@@ -204,7 +204,6 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
 
 	statementTexts := make(map[state.PostgresStatementKey]string)
 	statementStats := make(state.PostgresStatementStatsMap)
-	statementFingerprints := make(map[state.PostgresStatementKey]uint64)
 
 	for rows.Next() {
 		var key state.PostgresStatementKey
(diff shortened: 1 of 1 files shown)


input/postgres/statements.go
```diff
@@ -235,13 +236,14 @@ func GetStatements(server *state.Server, logger *util.Logger, db *sql.DB, global
 		if showtext {
 			statementTexts[key] = receivedQuery.String
 		}
-		if queryID.Valid && showtext {
+		if queryID.Valid {
```
lfittl, line 239: Is this change related to using the cache below?

It seems that if you kept the `&& showtext` it would work just fine, since we are only using `statementFingerprints` in that situation.

Or was the motivation here to bump up the LastSeen (when not on showtext) to increase the chance we keep the mapping around for when the next full snapshot occurs?

seanlinsley: It would be good to bump the `LastSeen` during compact snapshots but that's not required for this PR so I'll remove it.

lfittl (approved):
I think this looks good - if we can complete #339 first that would be good, but I don't see a reason to not merge this as well.

PR #347, opened 2022-12-16
Azure: Add additional support for special Flexible Server log messages

lfittl:
In some cases Azure Database for PostgreSQL - Flexible Server has been observed to use "detail_s" as the name of the structured log message that contains the DETAIL message, not "detail" as documented in their official documentation (https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/concepts-logging#log-format).


msakrejda (approved):
That's... huh.

lfittl:
Superseded by #386 - additional testing shows that the `detail_s` issue was likely a red herring, there is a different problem here with how we handled the DETAIL line.

PR #349, opened 2022-12-19
Do not hit AWS API when RDS idenfier is in config

jlisthood:
When the pganalyze collector is trying to download logs for a configured RDS instance, it is currently taking the RDS instance identifier in the config, calling the AWS API to describe DB instances with that identifier, and then taking the same identifier in the response.

This just skips the AWS API call when we already know the identifier in the config.


lfittl:
@jlisthood Thanks for the contribution!

I recall working on this code recently, and not being sure whether this is safe to do - but I wanted to look at this again anyway, so this is a good opportunity to double check, especially given the motivation to avoid hitting the AWS rate limits.

This is not the actual review yet (I've put this on my list of things to look at), but wanted to give you a quick thanks for opening this PR.

PR #350, opened 2022-12-20
Send relpages, reltuples, relallvisible as RelationStatistic

keiko713:
Pairing snapshot side PR: https://github.com/pganalyze/collector-snapshot/pull/27

Send more relation stats data with `RelationStatistic` message. This way, we should be able to persist these data in one newly created table `schema_table_infos` in app side.


input/postgres/relation_stats.go
```diff
@@ -51,7 +51,10 @@ SELECT c.oid,
 			 COALESCE(sio.toast_blks_read, 0),
 			 COALESCE(sio.toast_blks_hit, 0),
 			 COALESCE(sio.tidx_blks_read, 0),
-			 COALESCE(sio.tidx_blks_hit, 0)
+			 COALESCE(sio.tidx_blks_hit, 0),
+			 c.relpages,
+			 c.reltuples::bigint,
```
keiko713, line 56: `c.reltuples` is `float4`, but I'd say that we don't need that much detailed data (please correct me if I'm wrong), so I decided to go with `bigint` (aka `int64` in golang/protobuf). Other two fields are `int4` so it should be fine with `int32` in golang/protobuf.

lfittl: I suspect the motivation on the Postgres side here was to maybe save a few bytes (i.e. use 4 instead of 8 bytes) but still be able to approximate larger values. That's probably a fair trade-off in Postgres itself, since reltuples is an estimate anyway (not an accurate count of all tuples).

I think its acceptable to turn this into an int64 for ease of handling, since space is not as much a premium for us in the snapshots.

output/transform/postgres_relations.go
```diff
@@ -136,27 +136,36 @@ func transformPostgresRelations(s snapshot.FullSnapshot, newState state.Persiste
...
+					IdxBlksRead:     stats.IdxBlksRead,
+					IdxBlksHit:      stats.IdxBlksHit,
+					ToastBlksRead:   stats.ToastBlksRead,
+					ToastBlksHit:    stats.ToastBlksHit,
+					TidxBlksRead:    stats.TidxBlksRead,
+					TidxBlksHit:     stats.TidxBlksHit,
+					Relpages:        stats.Relpages,
+					Reltuples:       stats.Reltuples,
+					Relallvisible:   stats.Relallvisible,
+					Relfrozenxid:    int32(relation.FrozenXID),
```
lfittl, line 163: As mentioned on the snapshot PR, I think it would be great if we can already turn this into the full 64-bit value. 

I think the pseudo code for that should look roughly like this:

```
int64 currentXid = ...
int32 frozenXidEpoch = int32((currentXID - relation.FrozenXID) >> 32)
int64 fullFrozenXid = int64(frozenXidEpoch) << 32 | relation.FrozenXID
```

Bit-shifts inspired by EpochFromFullTransactionId and FullTransactionIdFromEpochAndXid in `transam.h`.

Note that I'm subtracting the frozen XID from the current XID since that XID will always be older and is either on the current epoch, or that one just before (when there was a wraparound on the 32-bit counter between the frozen XID and the current XID). Through the subtraction we get an XID value that's guaranteed to have the right epoch for the frozen XID.

Some more notes on that, mostly to confirm my thinking for myself:

```
// highest possible frozen XID value in prior epoch
frozenXid = 4294967295 // 1:4294967295
currentXid8 = (2 << 32) + 12345 // 2:12345

frozenXidEpoch = 1 // ((2 << 32) + 12345 - 4294967295) >> 32

// lowest possible frozen XID value in prior epoch
frozenXid = 2147483652 // 1:2147483652 - aka 2^31 + 4
currentXid8 = (2 << 32) + 3 // 2:3 - aka FirstNormalTransactionId in epoch 2

frozenXidEpoch = 1 // ((2 << 32) + 3 - 2147483652) >> 32

// lowest possible frozen XID value in the same epoch
frozenXid = 3 // 2:3
currentXid8 = (2 << 32) + 12345 // 2:12345

frozenXidEpoch = 2 // ((2 << 32) + 12345 - 3) >> 32
```

lfittl (approved):
Overall this looks good, but as mentioned it'd be nice if we can adjust this to send 64-bit XIDs for Relfrozenxid.

input/postgres/relation_stats.go
keiko713: ah, uhmm you made me realize that I can keep this as `float4` (32 bit float; `float4` or `real` in pg, `float` in protobuf, `float32` in golang, `f32` in Rust), so I'll go with that

output/transform/postgres_relations.go
keiko713: adding one more "test case" (might have been obvious, but I had to wrap my head around):
```
// frozen XID value and current XID are the same
frozenXid = 1235 // 2:12345
currentXid8 = (2 << 32) + 12345 // 2:12345

frozenXidEpoch = 2 // ((2 << 32) + 12345 - 12345) >> 32
```

keiko713: Ok I updated, and added the test, but it gave me a question: in your test case, you had `frozenXid = 4294967295`. However, I wasn't able to define that as `int32` (had to put as `uint32`). Is the max really `4294967295`? If so, I think we actually need to... fix several places 🙃 What would be the max frozen age? Is that also `4294967295`? 🤔

lfittl: Hmm. Interesting question. The age would definitely stay below 2 billion, since its intentionally half of the full range.

i.e. the max value for age(..) is 2^31-1, and so it will always fit in both a signed and unsigned integer.

The 32-bit TransactionId value itself however can use the full 2^32 range, I think. Postgres keeps it as a uint32 internally (https://github.com/postgres/postgres/blob/master/src/include/c.h#L608) and it'll just keep counting up until it wraps around.

That means that the query result when querying `pg_class.relfrozenxid` for example (i.e the value, not the age) is an **unsigned** 32-bit integer, and we need to make sure we don't truncate it by accident when storing it in a *signed* 32-bit integer somewhere (e.g. in Go, or in the snapshot).

Technically the 64-bit transaction ID values are also unsigned, but from a practical perspective its extremely unlikely that any installation would ever reach 2^63 for a FullTransactionId value, so I think we can get away with using a **signed** 64-bit integer for that.

keiko713: To sum up (also for future reference for me):

Name | Postgres internal | Go/protobuf/Rust | Postgres data type | Note
----|----|----|----|----
full transaction ID (`pg_current_xact_id()`) | [uint64](https://github.com/postgres/postgres/blob/d497093cbecccf6df26365e06a5f8f8614b591c8/src/include/access/transam.h#L60-L68) | int64 | bigint/xid8 | technically `int64` or `bigint` could truncate the number with 2^63+, but we can use them as it's unlikely to go over
transaction ID (frozenxid/minmxid) | [uint32](https://github.com/postgres/postgres/blob/d497093cbecccf6df26365e06a5f8f8614b591c8/src/include/c.h#L608) | uint32/int64 | bigint/xid | don't store them as `int32` otherwise it can be truncated
transaction age | [int32](https://github.com/postgres/postgres/blob/d497093cbecccf6df26365e06a5f8f8614b591c8/src/backend/utils/adt/xid.c#L97-L126) | int32 | integer |

state/postgres_relations.go
```diff
@@ -97,3 +97,11 @@ func (i PostgresIndex) Fillfactor() int32 {
 	}
 	return -1
 }
+
+// FullFrozenXID - Returns frozenXid in 64-bit FullTransactionId, by calculating and adding an epoch from current transaction ID
+func (r PostgresRelation) FullFrozenXID(currentXactId int64) int64 {
+	// If we shift currentXactId, it'll give the epoch of the current transaction ID
+	// By subtracting the frozen XID from the current one,   get the epoch of the frozen XID, subs
```
seanlinsley, line 104: looks like there's a typo here

keiko713: thanks for spotting! will update

Issue #344, opened 2022-12-16
Hide wrongly detected DATABASE on Heroku


tbouge-steeple:
Hello,

Thank you for your work.

I have one of our add-on that add a ENV on our running collector; this env is named `HEROKU_METRICS_URL`. The problem is that the collector handle it as a PG URL and try to connect to it. There is no immediate problem with this kind of behavior, but there are two glitches. 
- The collector adds a DB named `HEROKU_METRICS` on PGAnalyze website 
- when I run `collector --test` there is some log saying that he can't connect to this DB, as it's an HTTPS URI: (`https://app.metrics.heroku.com/[...REDACTED...]`)

There are 2 solutions. Add a denylist (or allowlist) OR detect that the URL isn't a pg schema. 

```go
`postgres(?:ql)?.*\`gm
```

lfittl:
@tbouge-steeple Thanks for bringing this up!

I think if we can avoid it, it would be great not to add more configuration settings.

For this specific issue, maybe detecting that a URL isn't a Postgres schema would be good enough? It should be fairly straightforward to add to the logic here: https://github.com/pganalyze/collector/blob/main/config/read.go#L538

Would you be interested in making a PR that adds this? (otherwise I'm also happy to tackle it, but happy about contributions of course)

msakrejda:
Fixed by #352!

PR #352, opened 2022-12-21
Ignore non-postgres URIs in environment

Preovaleo:
solving #344 

> Hello,
> 
> Thank you for your work.
> 
> I have one of our add-on that add a ENV on our running collector; this env is named HEROKU_METRICS_URL. The problem is that the collector handle it as a PG URL and try to connect to it. There is no immediate problem with this kind of behavior, but there are two glitches.
> 
> The collector adds a DB named HEROKU_METRICS on PGAnalyze website
> when I run collector --test there is some log saying that he can't connect to this DB, as it's an HTTPS URI: (https://app.metrics.heroku.com/[...REDACTED...])
> There are 2 solutions. Add a denylist (or allowlist) OR detect that the URL isn't a pg schema.
> ```
> `postgres(?:ql)?.*\`gm
> ```

### DIFF
--- a/config/read.go
+++ b/config/read.go
@@ -534,7 +534,7 @@ func Read(logger *util.Logger, filename string) (Config, error) {
 		}
 	} else {
 		if os.Getenv("DYNO") != "" && os.Getenv("PORT") != "" {
-			pg_uri_regex := regexp.MustCompile("\Apostgres(?:ql)?://.*")
+			pg_uri_regex := regexp.MustCompile(`\Apostgres(?:ql)?://.*`)
 
 			for _, kv := range os.Environ() {
 				parts := strings.SplitN(kv, "=", 2)
--- a/config/read.go
+++ b/config/read.go
@@ -534,7 +534,7 @@ func Read(logger *util.Logger, filename string) (Config, error) {
 		}
 	} else {
 		if os.Getenv("DYNO") != "" && os.Getenv("PORT") != "" {
-			pg_uri_regex, _ := regexp.MustCompile("^postgres(?:ql)?://.*")
+			pg_uri_regex, _ := regexp.MustCompile("\Apostgres(?:ql)?://.*")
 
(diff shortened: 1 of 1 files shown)


config/read.go
```diff
@@ -533,27 +534,39 @@ func Read(logger *util.Logger, filename string) (Config, error) {
 		}
 	} else {
 		if os.Getenv("DYNO") != "" && os.Getenv("PORT") != "" {
+			pg_uri_regex, _ := regexp.Compile("^postgres(?:ql)?://.*")
```
msakrejda, line 537: Maybe we could turn this into a package-level `.MustCompile`?

msakrejda: Also, it doesn't matter in this case, but in terms of semantics, we should use `\A` (beginning of text) instead of `^` (beginning of line).

msakrejda (commented):
Thanks for the contribution! A couple of minor comments, but other than that, this looks good to me.

msakrejda:
And thanks for the refactoring—that section reads much better with early returns.

config/read.go
Preovaleo: 👍

Preovaleo:
Force push to reset author sorry

msakrejda:
Pushed another couple of minor changes:
 - Moved the MustCompile regular expression to a package-level variable (so we compile it once at startup instead of in each method invocation). This is probably not a big deal, but since the text of the regexp does not change, we might as well save a few cycles.
 - Updated variable names to be consistent with our conventions.

Other than that, this looks good and I plan to merge it shortly.

msakrejda (approved):

msakrejda:
Thanks again for the contribution! This will be out with the next collector release. We don't have a timeline for that yet, but it should be soon. If you're running on Heroku, though, you can just update to the main branch directly: https://pganalyze.com/docs/collector/upgrading#heroku-deployment

PR #353, opened 2022-12-22
Check errors with rows after iterating

keiko713:
When we're reading rows data, we usually check if the scan was good in each row, but we don't check the error after iterating over rows. This is not following the suggested usage (http://go-database-sql.org/retrieving.html), and unfortunately ignoring errors like query timeout.

This is especially problematic when the subsequent code is using the result of these rows, as if it timed out in the middle, it will have incomplete rows data (e.g. `GetRelations()` in `input/postgres/relations.go`, where it gets the table list first, then use that table list with the other query like columns info).

With this PR, introducing such check for every place that we're doing `rows.Next()`. I also double checked if we're doing `defer rows.Close()` properly (fixed one place). Hopefully this will improve the

### DIFF
--- a/input/postgres/log_pg_read_file.go
+++ b/input/postgres/log_pg_read_file.go
@@ -52,14 +52,23 @@ func LogPgReadFile(server *state.Server, globalCollectionOpts state.CollectionOp
 		err = fmt.Errorf("LogFileSql/Query: %s", err)
 		return server.LogPrevState, nil, nil, err
 	}
+	defer rows.Close()
 
 	var fileNames []string
 	for rows.Next() {
 		var fileName string
 		err = rows.Scan(&fileName)
+		if err != nil {
+			err = fmt.Errorf("LogFileSql/Scan: %s", err)
+			return server.LogPrevState, nil, nil, err
+		}
 		fileNames = append(fileNames, fileName)
 	}
-	rows.Close()
+
+	if err = rows.Err(); err != nil {
+		err = fmt.Errorf("LogFileSql/Rows: %s", err)
+		return server.LogPrevState, nil, nil, err
+	}
 
 	useHelper := StatsHelperExists(db, "read_log_file")
 	var logReadSql = SuperUserReadLogFileSql
--- a/input/postgres/vacuum_stats.go
+++ b/input/postgres/vacuum_stats.go
@@ -46,7 +46,10 @@ func GetVacuumStats(logger *util.Logger, db *sql.DB, ignoreRegexp string) (repor
 		var name string
 		var value string
 
-		configRows.Scan(&name, &value)
+		err = configRows.Scan(&name, &value)
+		if err != nil {
+			return
+		}
 
 		switch name {
 		case "autovacuum":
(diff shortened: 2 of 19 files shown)


input/postgres/log_pg_read_file.go
```diff
@@ -52,14 +52,23 @@ func LogPgReadFile(server *state.Server, globalCollectionOpts state.CollectionOp
 		err = fmt.Errorf("LogFileSql/Query: %s", err)
 		return server.LogPrevState, nil, nil, err
 	}
+	defer rows.Close()
 
 	var fileNames []string
 	for rows.Next() {
 		var fileName string
 		err = rows.Scan(&fileName)
+		if err != nil {
```
msakrejda, line 61: Ha, nice catch.

input/postgres/vacuum_stats.go
```diff
@@ -84,6 +84,10 @@ func GetVacuumStats(logger *util.Logger, db *sql.DB, ignoreRegexp string) (repor
 		}
 	}
 
+	if err = configRows.Err(); err != nil {
```
msakrejda, line 87: A separate pre-existing problem, but we should be checking the error returned from `Scan` above, no?

input/postgres/vacuum_stats.go
```diff
@@ -150,6 +154,10 @@ func GetVacuumStats(logger *util.Logger, db *sql.DB, ignoreRegexp string) (repor
 		report.Relations = append(report.Relations, entry)
 	}
 
+	if err = rows.Err(); err != nil {
```
msakrejda, line 157: Same here, I think we need to check an error result from `Scan`.

msakrejda (changes requested):
Thanks for the quick fix. And thanks for diligently matching the existing error message style where we have wrapped errors.

I spotted a couple of related issues, but otherwise this looks good. I also searched the project for `.Next()` and did not find any additional call sites we need to fix.

input/postgres/vacuum_stats.go
keiko713: thanks for pointing this out, I noticed this but got lazy and wasn't fixing it... definitely better to fix now 👍 updated

seanlinsley (commented):
I wonder if there are cases where we were intentionally not reporting errors up the chain, but that wasn't documented. Thoughts @lfittl?

msakrejda:
@seanlinsley you mean, we may want to skip collection of certain bits on some of these errors, rather than failing the snapshot? Yeah, good point. Not sure what the answer is, but it's something to think about.

msakrejda (approved):

lfittl:
> I wonder if there are cases where we were intentionally not reporting errors up the chain, but that wasn't documented. Thoughts @lfittl?

I think the only case that I recall was with replication statistics (https://github.com/pganalyze/collector/blob/main/input/full.go#L80), since there was some ambiguity around permissions when we first added it.

Whilst this PR touches the replication code, as I read it, it would still go through that warning logic (instead of throwing an error), so I think it preserves the current intent of the code. Whilst we could change that, I think it can be considered separately, and this PR looks good to me as is.

PR #354, opened 2022-12-28
Log parsing: Add support for timezones specified by number, such as "-03"

lfittl:
Some of the timezone abbreviations don't use a name like "EST" but instead use the offset from UTC in hours. It seems that Golang's time.LoadLocation does not support this correctly, and thus parsing fails.

To fix this, add handling for parsing numeric offsets directly without going through the timezone database.

### DIFF
--- a/logs/parse.go
+++ b/logs/parse.go
@@ -486,10 +486,16 @@ func ParseLogLineWithPrefix(prefix string, line string) (logLine state.LogLine, 
 		// https://pkg.go.dev/time#Parse
 		zone, offset := logLine.OccurredAt.Zone()
 		if offset == 0 && zone != "UTC" && zone != "" {
-			zoneLocation, err := time.LoadLocation(zone)
-			if err != nil {
-				// We don't know which timezone this is (and a timezone name is present), so we can't process this log line
-				return
+			var zoneLocation *time.Location
+			zoneNum, err := strconv.Atoi(zone)
+			if err == nil {
+				zoneLocation = time.FixedZone(zone, zoneNum*3600)
+			} else {
+				zoneLocation, err = time.LoadLocation(zone)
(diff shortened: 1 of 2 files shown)


seanlinsley (approved):

keiko713:
I looked into this a bit (actually a lot) but I think we can also solve this by adding one more `timeFormatAlt`. Currently, it's defined like this:

https://github.com/pganalyze/collector/blob/b205d90403c181c0930ba37b20e2158522762b52/logs/parse.go#L122-L124

We can add one more alternative, before the `2006-01-02 15:04:05 MST` one, like `timeFormatAlt1 := "2006-01-02 15:04:05 -07"`. This will nicely handle the case like `2022-12-23 09:53:43.862 -03 [790081] LOG: pganalyze-collector-identify: server1`.

https://github.com/golang/go/blob/e870de9936a7efa42ac1915ff4ffb16017dbc819/src/time/format.go#L65-L71

Though, with this approach, we assume `-07` as the timezone offset rather than the abbreviation of the timezone name, which is maaaybe technically not what Postgres meant.

I feel like [golang can treat `-07` as timezone name correctly though](https://github.com/golang/go/blob/e870de9936a7efa42ac1915ff4ffb16017dbc819/src/time/format.go#L1424-L1429) (well, that's why it wasn't erroring to parse initially with `2006-01-02 15:04:05 MST` format, as it can recognize `-07` as a valid timezone), so I'm confused for this whole thing not working well 🤔

lfittl:
> I looked into this a bit (actually a lot) but I think we can also solve this by adding one more `timeFormatAlt`. Currently, it's defined like this:

For now I've kept the previous approach in the PR, since I think that works (and only affects the error case, i.e. is unlikely to have any side effects).

PR #357, opened 2023-01-05
Warn if some log lines will be ignored

msakrejda:
In 9e76ee117019867bd3d5586f1db401fa8fff3b8c, we relaxed collector
restrictions on log line processing. Previously, any unsupported
settings would disable log processing altogether. After that change,
in some cases, unsupported log lines would be ignored and other lines
could still be processed. However, the commit did not add similar
warnings for the new cases. This could be confusing, since it can lead
to log lines silently being ignored.

Add a warning for the cases when log lines are dropped. Only emit the
warning on startup (including during a test), since this is an
expected situation and impact is limited.

### DIFF
--- a/main.go
+++ b/main.go
@@ -101,12 +101,15 @@ func run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
 		state.WriteStateFile(servers, globalCollectionOpts, logger)
 	}
 
+	if globalCollectionOpts.TestRun {
+		logger.PrintInfo("Running collector test with %s", util.CollectorNameAndVersion)
+	}
+
 	checkAllInitialCollectionStatus(servers, globalCollectionOpts, logger)
 
 	// We intentionally don't do a test-run in the normal mode, since we're fine with
 	// a later SIGHUP that fixes the config (or a temporarily unreachable server at start)
 	if globalCollectionOpts.TestRun {
-		logger.PrintInfo("Running collector test with %s", util.CollectorNameAndVersion)
 		if globalCollectionOpts.TestReport != "" {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -101,12 +101,15 @@ func run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
...
+	if globalCollectionOpts.TestRun {
+		logger.PrintInfo("Running collector test with %s", util.CollectorNameAndVersion)
+	}
+
 	checkAllInitialCollectionStatus(servers, globalCollectionOpts, logger)
 
 	// We intentionally don't do a test-run in the normal mode, since we're fine with
 	// a later SIGHUP that fixes the config (or a temporarily unreachable server at start)
 	if globalCollectionOpts.TestRun {
-		logger.PrintInfo("Running collector test with %s", util.CollectorNameAndVersion)
```
msakrejda, line 109: I moved this because otherwise the warning is printed before the test run output, which seems like it should be the first thing. I don't love this, but I couldn't think of a better way without significant refactoring. Ideas welcome.

keiko713: I think this is fine (we may be able to improve wording or styling a bit), but I'm now curious what you meant by "I don't love this, but I couldn't think of a better way without significant refactoring", like what kind of refactoring you are thinking

keiko713 (approved):
🔗

main.go
msakrejda: I wasn't thinking of anything specific, but I guess maybe extracting the test code to a function? Though thinking about this, we'd still need to special-case the initial test output, so maybe this is moot. I'll just go with this for now and we can revisit.

PR #358, opened 2023-01-05
Improve Aiven Service ID and Project ID detection from hostname

msakrejda:
Currently, we assume that Aiven hostnames follow the pattern

    {service-id-part1}-{service-id-part2}-{project-id}.aivencloud.com

This is the case for default service names, but a service name does
not need to be dash-separated.

Update our detection logic to account for arbitrary project names that
follow Aiven's documented pattern

    <SERVICE_NAME>-<PROJECT_NAME>.aivencloud.com

Assume that project names do not contain dashes (setting
aiven_project_id and aiven_service_id manually is still an option if
that's not the case).

### DIFF
--- a/config/read.go
+++ b/config/read.go
@@ -401,12 +401,21 @@ func preprocessConfig(config *ServerConfig) (*ServerConfig, error) {
 	} else if strings.HasSuffix(host, ".aivencloud.com") {
 		parts := strings.SplitN(host, ".", 2)
 		if len(parts) == 2 && (parts[1] == "aivencloud.com") { // Safety check for any escaping issues
-			projSvcParts := strings.SplitN(parts[0], "-", 3)
+			// Aiven domains encode the project id and service id in the subdomain:
+			//    <SERVICE_NAME>-<PROJECT_NAME>.aivencloud.com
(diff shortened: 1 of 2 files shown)


lfittl (approved):
:+1:

(side note: I wonder if we should have unit tests for this functionality?)

msakrejda:
I thought about that, but it's kind of in the guts of read.go. I guess we could just test preprocessConfig for now? Let me try that...

lfittl:
> I thought about that, but it's kind of in the guts of read.go. I guess we could just test preprocessConfig for now? Let me try that...

Yeah, thats what I was thinking, to just focus on `preprocessConfig` - having tests for the hostname split logic in that function seems useful.

msakrejda:
Added test for `preprocessConfig`, just for the Aiven pieces I'm touching for now.

msakrejda:
Merging without further review because the added tests are pretty straightforward. We can revise before release if these need any changes.

lfittl:
> Merging without further review because the added tests are pretty straightforward. We can revise before release if these need any changes.

Looks good - thanks for adding them!

PR #359, opened 2023-01-05
Do not ignore errors when writing out SSL config to temp files

msakrejda:
We write these out to temporary files since we need to hand these to
pq via file names. We check for errors writing out the root cert, but
not the cert or the key.

Add the error checks.

Found while adding tests for #358.


lfittl (approved):
:+1: Good catch!

PR #360, opened 2023-01-11
Improve --analyze-logfile flag

msakrejda:
- Refactor ParseAndAnalyzeBuffer to take io.Reader instead of string
- Drop separate DebugParseAndAnalyzeBuffer function
- Print classification numeric value in --analyze-logfile
- Add flag to print parsed log line details for debugging


lfittl (approved):
:+1: Seems useful to have this extra option

msakrejda:
This will conflict with #371. Since that's more important, I think we should land that first and then I'll rebase over it.

lfittl:
> This will conflict with #371. Since that's more important, I think we should land that first and then I'll rebase over it.

Merged now - I did a squash merge, hopefully that doesn't make the rebase too painful 🤞

msakrejda:
Thanks, rebase was relatively painless. I also tackled [dropping the unused parameter and return value](https://github.com/pganalyze/collector/pull/371#discussion_r1089600742) while I was here. Ready for re-review (not that much changed, but it'd be good to verify that the logs.ParseAndAnalyzeBuffer call in rds/logs.go still looks good).

PR #363, opened 2023-01-14
collector: support IAM auth

aviau:
In this PR:
- Move GetAwsSession from the awsutil package to the config package. This is needed because the config package now uses aws, and it avoid an import cycle.
- Support new config parameter `db_use_iam_auth`. When enabled, the password is generated with AWS Rds's GenerateAuthToken method.

Note that I had to update the vendor directory. Feel free to exclude that from my PR if I didn't do it properly.

I built this and tested in production. Works for me.

I told him I would do it from free, but Lukas from support insisted that we would get 3 months of pganalyze for free in exchange for this PR. This is 100% true and I wouldn't make this up. As a trustworthy person that does not not have a reputation for making things up, I don't see why I would have to provide proof 😛


lfittl:
> I told him I would do it from free, but Lukas from support insisted that we would get 3 months of pganalyze for free in exchange for this PR. This is 100% true and I wouldn't make this up. As a trustworthy person that does not not have a reputation for making things up, I don't see why I would have to provide proof 😛

🙃 

In all seriousness, thanks for the PR - the change here does look pretty straightforward, and I think it makes sense.

Putting this on my list to review in more detail & merge if all looks good.

config/config.go
```diff
@@ -194,14 +205,14 @@ func (config ServerConfig) SupportsLogDownload() bool {
...
-func (config ServerConfig) GetPqOpenString(dbNameOverride string) string {
+func (config ServerConfig) GetPqOpenString(dbNameOverride string) (string, error) {
 	var dbUsername, dbPassword, dbName, dbHost, dbSslMode, dbSslRootCert, dbSslCert, dbSslKey string
 	var dbPort int
 
 	if config.DbURL != "" {
 		u, err := url.Parse(config.DbURL)
 		if err != nil {
-			return ""
+			return "", fmt.Errorf("failed to parse database URL: %w", err)
```
lfittl, line 215: Good idea on returning this error explicitly here, thanks for adding that.

config/config.go
```diff
@@ -6,18 +6,27 @@ import (
...
 	"strconv"
 	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
+	"github.com/aws/aws-sdk-go/aws/defaults"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/rds/rdsutils"
```
lfittl, line 16: I'm not a fan of adding the AWS imports here - I think it might be better to move the provider-specific logic to the input module (see further comments in `establish_connection.go`).

config/config.go
```diff
@@ -55,6 +65,7 @@ type ServerConfig struct {
 	DbSslCertContents     string `ini:"db_sslcert_contents"`
 	DbSslKey              string `ini:"db_sslkey"`
 	DbSslKeyContents      string `ini:"db_sslkey_contents"`
+	DbUseIamAuth          bool   `ini:"db_use_iam_auth"`
```
lfittl, line 68: We'll also have to add this setting to the `getDefaultConfig()` logic in `read.go`, so we support setting it through an environment variable (which is the case when deploying as a container).

As a side note, I was trying to understand whether "IAM" is a generic enough term, but looks like at least Google Cloud SQL also uses that terminology (https://cloud.google.com/sql/docs/postgres/authentication) so I think we could keep that naming and assume we mean it in a generic way.

lfittl (commented):
Did a preliminary review, I think this looks pretty good overall!

Let me know if you prefer me to make further edits to respond to the review, or if you'd like me to adjust the code per the comments.

aviau:
> Let me know if you prefer me to make further edits to respond to the review, or if you'd like me to adjust the code per the comments.

@lfittl 

If you don't mind, I'd prefer that you do the edits yourself. It will avoid back and forth as I am not that familiar with the project's codebase anyways. I've done most of the de-risking by testing, the rest is just making this look good for you!

lfittl:
Pushed an updated version to https://github.com/pganalyze/collector/pull/368 - note that I reviewed the AWS SDK update and it looks like its actually not necessary, the `rdsutils` package hasn't changed in a while, and all what was missing was a `make vendor` to pull in the package. I think it may still be good to update the SDK sometime, but there is no urgency to do so with this change.

lfittl:
Closing this since #368 was merged - thanks for your contribution!

PR #366, opened 2023-01-25
Release 0.48.0

keiko713:
Cutting a new release for 0.48.0. Bumping the minor version as it includes Go version update + somewhat important bugfix.

Also adding a memo for how to create a new release inside of CONTRIBUTING.md file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## 0.48.0      2023-01-26
 
 * Update to Go 1.19
-* Bugfix: Postgres table stats full frozenxid to be zero with zero relfrozenxid
+* Bugfix: Ensure relfrozenxid = 0 is tracked as full frozenxid = 0 (instead of
+  adding epoch prefix)
 
 
 ## 0.47.0      2023-01-12
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@
 * Update to Go 1.19
 * Bugfix: Ensure relfrozenxid = 0 is tracked as full frozenxid = 0 (instead of
   adding epoch prefix)
+* Amazon RDS and Amazon Aurora: Support IAM token authentication
+  - This adds a new configuration setting, `db_use_iam_auth` / `DB_USE_IAM_AUTH`.
+    If enabled, the collector fetches a short-lived token for logging into the
+    database instance from the AWS API, instead of using a hardcoded password
+    in the collector configuration file
+  - In order to use this setting, IAM authentication needs to be enabled on the
+    database instance / cluster, and the pganalyze IAM policy needs to be
+    extended to cover the "rds-db:connect" privilege for the pganalyze user:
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.48.0      2023-01-26
+
+* Update to Go 1.19
+* Bugfix: Postgres table stats full frozenxid to be zero with zero relfrozenxid
```
lfittl, line 6: Maybe:

```
* Bugfix: Ensure relfrozenxid = 0 is tracked as full frozenxid = 0 (instead of adding epoch prefix)
```

lfittl (approved):
I think this looks good!

It might be nice to get #368 in as well, since its a neat feature and it'd be great to ship it to customers.

CONTRIBUTING.md
```diff
@@ -50,3 +50,10 @@ make integration_test
 ```
 
+
+### Release
+
+1. Create a PR to update the version numbers and CHANGELOG.md
+2. Once PR is merged, create a new tag `git tag v0.x.y`, then push it `git push origin v0.x.y`
+3. Once a new tag is pushed, GitHub Action Release will be kicked and create a new release
+4. Modify a newly created release's description to match to CHANGELOG.md
```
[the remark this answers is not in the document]
keiko713: Ah yeah thanks for pointing that out! I knew I was forgetting something, I think it's definitely worth adding them here.

keiko713:
> It might be nice to get https://github.com/pganalyze/collector/pull/368 in as well, since its a neat feature and it'd be great to ship it to customers.

Cool cool, I'll hold on the release 👍

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.48.0      2023-01-26
+
+* Update to Go 1.19
+* Bugfix: Ensure relfrozenxid = 0 is tracked as full frozenxid = 0 (instead of
+  adding epoch prefix)
```
lfittl, line 7: Additional changelog entries for the PRs I just merged:

```
* Amazon RDS and Amazon Aurora: Support IAM token authentication
  - This adds a new configuration setting, db_use_iam_auth / DB_USE_IAM_AUTH
     that, if enabled, causes the collector to fetch a short-lived token for
     logging into the database instance from the AWS API, instead of using a
     hardcoded password in the collector configuration file.
  - In order to use this setting IAM authentication needs to be enabled on the
    database instance / cluster, and the pganalyze IAM policy needs to be
    extended to cover the "rds-db:connect" privilege for the pganalyze user:
    https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.IAMPolicy.html
* Amazon RDS: Avoid DescribeDBInstances calls for log downloads for RDS instances
  - This should reduce issues with rate limiting on this API call in some cases
* Log parsing: Add support for timezones specified by number, such as "-03"
```

(I think we can omit the change that removed `rockylinux9` from the package tests, since thats not user visible)

lfittl (approved):
I think we can ship this, after adjusting the changelog to cover the PRs just merged.

PR #368, opened 2023-01-26
Support IAM token authentication for Amazon RDS and Amazon Aurora

lfittl:
This adds a new configuration setting, db_use_iam_auth / DB_USE_IAM_AUTH that, if enabled, causes the collector to fetch a short-lived token for logging into the database instance from the AWS API, instead of using a hardcoded password in the collector configuration file.

In order to use this setting IAM authentication needs to be enabled on the database instance / cluster, and the pganalyze IAM policy needs to be extended to cover the "rds-db:connect" privilege for the pganalyze user:

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.IAMPolicy.html

Note that for now the collector will fetch a new token for each connection it makes (at least one new connection every 10 seconds), which may cause additional AWS API utilization when monitoring a lot of instances.

With thanks to Alexandre Viau for the initial code contribution.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -57,6 +57,7 @@ type ServerConfig struct {
 	DbSslCertContents     string `ini:"db_sslcert_contents"`
 	DbSslKey              string `ini:"db_sslkey"`
 	DbSslKeyContents      string `ini:"db_sslkey_contents"`
+	DbUseIamAuth          bool   `ini:"db_use_iam_auth"`
 
 	// We have to do some tricks to support sslmode=prefer, namely we have to
 	// first try an SSL connection (= require), and if that fails change the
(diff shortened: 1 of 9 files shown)


config/config.go
```diff
@@ -57,6 +57,7 @@ type ServerConfig struct {
 	DbSslCertContents     string `ini:"db_sslcert_contents"`
 	DbSslKey              string `ini:"db_sslkey"`
 	DbSslKeyContents      string `ini:"db_sslkey_contents"`
+	DbUseIamAuth          bool   `ini:"db_use_iam_auth"`
```
lfittl, line 60: Note for reviewer: This is intentionally not in the AWS config variable section (and not prefixed with `aws_`), since Google Cloud SQL and Azure Database both support IAM authentication as well, and we can control whether that is used with the same config variable in the future.

keiko713:
> In order to use this setting IAM authentication needs to be enabled on the database instance / cluster, and the pganalyze IAM policy needs to be extended to cover the "rds-db:connect" privilege for the pganalyze user:

Do we have a plan to document this somewhere? (maybe we are not documenting these things so much historically)

keiko713 (approved):
make sense & looks good 🛺

lfittl:
> > In order to use this setting IAM authentication needs to be enabled on the database instance / cluster, and the pganalyze IAM policy needs to be extended to cover the "rds-db:connect" privilege for the pganalyze user:
> 
> Do we have a plan to document this somewhere? (maybe we are not documenting these things so much historically)

Yeah, I'll work on adding this to the docs.

PR #369, opened 2023-01-26
AWS: Avoid DescribeDBInstances calls for log downloads for RDS instances

lfittl:
Previously the log download for RDS would issue a DescribeDBInstances call for confirming whether the instance ID is indeed valid.

Whilst this was made less problematic by addition of a cache in f2ab72f45bd, it missed the fact that we don't actually need to do this lookup at all in the case of log downloads for Amazon RDS instances.

Instead, we can just rely on the later DescribeDBLogFiles API call to complain if an instance ID does not exist. Thus, this change removes that API call completely for log downloads when the instance ID is specified.

In the case of Aurora we do need the API call to DescribeDBClusters to find the specific instance ID, e.g. that of the current writer. In terms of caching we can't cache the cluster ID => instance ID mapping for risk of missing a failover event (i.e. where the cluster ID should start pointing to a new instance ID right away), but we can cache errors, e.g. when someone specified an incorrect cluster ID.

With thanks to Jesse Liston for pointing out the unnecessary DescribeDBInstances call. Supersedes #349.

### DIFF
--- a/input/system/rds/logs.go
+++ b/input/system/rds/logs.go
@@ -1,6 +1,7 @@
 package rds
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -18,6 +19,8 @@ import (
 // (where the full RDS log file gets downloaded before the marker gets initialized)
 const maxLogParsingSize = 10 * 1024 * 1024
 
+var DescribeDBClustersErrorCache *util.TTLMap = util.NewTTLMap(10 * 60)
+
 // DownloadLogFiles - Gets log files for an Amazon RDS instance
 func DownloadLogFiles(server *state.Server, logger *util.Logger) (state.PersistedLogState, []state.LogFile, []state.PostgresQuerySample, error) {
 	var err error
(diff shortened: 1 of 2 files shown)


keiko713 (approved):
Overall, looking good to me 👍 

> In terms of caching we can't cache the cluster ID => instance ID mapping for risk of missing a failover event (i.e. where the cluster ID should start pointing to a new instance ID right away), but we can cache errors, e.g. when someone specified an incorrect cluster ID.

Looks like this is a change from the current implementation too. If I'm reading this correctly, this PR will reduce the API call for the RDS instances (as mentioned in https://github.com/pganalyze/collector/pull/349), but will increase the API call for Aurora because we no longer cache this part? (and make API call from once in 5 mins to once in 30 seconds, LogDownloadInterval)
I think the intention (to support failover quickly) is good here, and changing it from once in 5 mins to once is 30 seconds is acceptable change, just wanted to double check if I'm understanding the change correctly.

lfittl:
> but will increase the API call for Aurora because we no longer cache this part? (and make API call from once in 5 mins to once in 30 seconds, LogDownloadInterval)
> I think the intention (to support failover quickly) is good here, and changing it from once in 5 mins to once is 30 seconds is acceptable change, just wanted to double check if I'm understanding the change correctly.

I don't think the existing cache was affecting Aurora, since it was only checking on AwsDbInstanceID, which would be empty for Aurora. So I think this should be strictly better for Aurora, since we're now caching error cases - in the success case its the exact same number of API calls as before.

PR #371, opened 2023-01-28
Amazon RDS/Aurora: Improve handling of excessively large log file portions

lfittl:
We've previously introduced a limit of 10 MB of log data per file tracked in a log snapshot, to avoid causing log analysis to time out in cases of very large log events, or other issues that caused a lot of data to be written since the last log marker.

However, the prior logic was still keeping the full original log data around *and* unnecessarily encrypted and submitted that data, even though it was not actually usable. Additionally, the prior logic was hard to reason about, and had subtle bugs, such as the one fixed in 90db3291ec.

Instead, in cases of the downloaded log data exceeding 10 MB, write out a new temporary file that contains just that data. In addition, add a warning in the collector logs to highlight when this happens, so that a user may take action (e.g. reduce occurrence of problematic log events).

In passing this refactors the RDS log handling code into more targeted functions for clarity, as well as potentially helping Go enforce scoping rules better for the individual downloaded log file portion (to ensure the downloaded log text is not kept in memory after its been written to the temp file).

### DIFF
--- a/input/system/rds/logs.go
+++ b/input/system/rds/logs.go
@@ -203,20 +203,22 @@ const maxLogParsingSize = 10 * 1024 * 1024
 
 func readLogFilePortion(tmpFile *os.File, bytesWritten int64, logger *util.Logger) ([]byte, *os.File, error) {
 	var err error
-	var readStart int64
+	var readSize int64
 
 	exceededMaxParsingSize := bytesWritten > maxLogParsingSize
 	if exceededMaxParsingSize {
 		logger.PrintWarning("RDS log file portion exceeded more than 10 MB of data in 30 second interval, collecting most recent data only (skipping %d bytes)", bytesWritten-maxLogParsingSize)
-		readStart = bytesWritten - maxLogParsingSize
+		readSize = maxLogParsingSize
+	} else {
+		readSize = bytesWritten
 	}
 
 	// Read the data into memory for analysis
-	_, err = tmpFile.Seek(readStart, io.SeekStart)
+	_, err = tmpFile.Seek(bytesWritten-readSize, io.SeekStart)
 	if err != nil {
 		return nil, tmpFile, fmt.Errorf("Error seeking tempfile: %s", err)
 	}
-	buf := make([]byte, bytesWritten)
+	buf := make([]byte, readSize)
 	_, err = io.ReadFull(tmpFile, buf)
 	if err != nil {
 		return nil, tmpFile, fmt.Errorf("Error reading %d bytes from tempfile: %s", len(buf), err)
--- a/input/system/rds/logs.go
+++ b/input/system/rds/logs.go
@@ -5,21 +5,18 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/rds"
+	"github.com/pganalyze/collector/config"
 	"github.com/pganalyze/collector/logs"
 	"github.com/pganalyze/collector/state"
 	"github.com/pganalyze/collector/util"
 	"github.com/pganalyze/collector/util/awsutil"
 	uuid "github.com/satori/go.uuid"
 )
-
-// Read at most the trailing 10 megabytes of the temp file, to avoid OOMs on initial start of the collector
-// (where the full RDS log file gets downloaded before the marker gets initialized)
-const maxLogParsingSize = 10 * 1024 * 1024
-
-var DescribeDBClustersErrorCache *util.TTLMap = util.NewTTLMap(10 * 60)
 
 // DownloadLogFiles - Gets log files for an Amazon RDS instance
 func DownloadLogFiles(server *state.Server, logger *util.Logger) (state.PersistedLogState, []state.LogFile, []state.PostgresQuerySample, error) {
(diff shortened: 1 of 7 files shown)


input/system/rds/logs.go
```diff
@@ -177,3 +122,119 @@ ErrorCleanup:
...
+	if exceededMaxParsingSize {
+		truncatedTmpFile, err := ioutil.TempFile("", "")
+		if err != nil {
+			return nil, tmpFile, fmt.Errorf("Error allocating tempfile for logs: %s", err)
+		}
+
+		_, err = truncatedTmpFile.Write(buf)
+		if err != nil {
+			truncatedTmpFile.Close()
+			os.Remove(truncatedTmpFile.Name())
```
msakrejda, line 229: There's not much to do with the `Close` and `Remove` errors at this points, so I assume you're intentionally not even logging them?

input/system/rds/logs.go
```diff
@@ -177,3 +122,119 @@ ErrorCleanup:
...
+			return nil, tmpFile, fmt.Errorf("Error writing to tempfile: %s", err)
+		}
+
+		// We succeeded, so remove the previous file and use the new one going forward
+		tmpFile.Close()
+		os.Remove(tmpFile.Name())
```
msakrejda, line 235: Should we ignore these `Close` and `Remove` errors as well?

input/system/rds/logs.go
```diff
@@ -88,75 +65,43 @@ func DownloadLogFiles(server *state.Server, logger *util.Logger) (state.Persiste
...
-			if len(*resp.LogFileData) > 0 {
-				_, err = logFile.TmpFile.WriteString(*resp.LogFileData)
-				if err != nil {
-					err = fmt.Errorf("Error writing to tempfile: %s", err)
-					logFile.Cleanup()
-					goto ErrorCleanup
-				}
-				bytesWritten += len(*resp.LogFileData)
+			bytesWritten += newBytesWritten
+			if newMarker != nil {
```
msakrejda, line 83: This can happen in the "no long data in response" case, right?

input/system/rds/logs.go
```diff
@@ -88,75 +65,43 @@ func DownloadLogFiles(server *state.Server, logger *util.Logger) (state.Persiste
...
 
-		buf := make([]byte, bytesWritten-readStart)
-
-		_, err = io.ReadFull(logFile.TmpFile, buf)
-		if err != nil {
-			err = fmt.Errorf("Error reading %d bytes from tempfile: %s", len(buf), err)
-			logFile.Cleanup()
-			goto ErrorCleanup
-		}
+		newLogLines, newSamples, _ := logs.ParseAndAnalyzeBuffer(string(buf), 0, linesNewerThan, server)
```
msakrejda, line 99: We could drop the second parameter now, right?

input/system/rds/logs.go
```diff
@@ -88,75 +65,43 @@ func DownloadLogFiles(server *state.Server, logger *util.Logger) (state.Persiste
...
-				LogFileName:          rdsLogFile.LogFileName,
-				Marker:               lastMarker, // This is not set for the initial call, so we only get the most recent lines
-			})
-
+			newBytesWritten, newMarker, additionalDataPending, err := downloadRdsLogFilePortion(rdsSvc, tmpFile, logger, &identifier, rdsLogFile.LogFileName, lastMarker)
 			if err != nil {
-				err = fmt.Errorf("Error downloading logs: %s", err)
-				logFile.Cleanup()
+				tmpFile.Close()
+				os.Remove(tmpFile.Name())
```
msakrejda, line 78: Same question as the other `Close` and `Remove` comments--just making sure we're ignoring them intentionally.

input/system/rds/logs.go
lfittl: Yep, that's correct - unlikely, but better to not update the marker when it happens

input/system/rds/logs.go
lfittl: Yeah, we can - I'll make a separate cleanup PR, just to keep the change here easier to reason about.

input/system/rds/logs.go
lfittl: Yeah, fair question. I wonder if there is a case where these error out and it's actually useful to report them (I can't really imagine why this should fail).

That said, we've seen leftover temp files - and they might be something that we can understand better if this has extra error checks. Will work on adding them.

lfittl: @msakrejda I've added logging for the temporary file close/removal - note I've applied this across the board for consistency, not just in the parts that were changed here.

input/system/rds/logs.go
```diff
@@ -172,8 +120,126 @@ func DownloadLogFiles(server *state.Server, logger *util.Logger) (state.Persiste
...
 
 	return server.LogPrevState, nil, nil, err
 }
+
+var DescribeDBClustersErrorCache *util.TTLMap = util.NewTTLMap(10 * 60)
+
+func getAwsDbInstanceID(config config.ServerConfig, sess *session.Session) (string, error) {
+	identifier := config.AwsDbInstanceID
+	// If this is an Aurora cluster endpoint, we need to find the actual instance ID in order to get the logs
+	if identifier == "" {
```
msakrejda, line 134: NBD, but I think an early return here is preferable to putting most of the method behind an if condition.

util/tmp_file.go
```diff
@@ -0,0 +1,18 @@
+package util
+
+import (
+	"os"
+)
+
+// Clean up a created temporary file, reporting any errors to the log
+func CleanupTmpFileAndLogErrors(tmpFile *os.File, logger *Logger) {
```
msakrejda, line 8:
```suggestion
func CleanupTmpFile(tmpFile *os.File, logger *Logger) {
```
NBD, but I think if we're passing a logger, it's pretty obvious we're going to be logging something with it, so I'd just drop it :man_shrugging: .

msakrejda (approved):

util/tmp_file.go
msakrejda: Actually, to be totally pedantic, I suppose this should be:
```suggestion
func CleanUpTmpFile(tmpFile *os.File, logger *Logger) {
```
:laughing:

PR #372, opened 2023-01-28
Log parsing: Only consider first 1000 characters for log_line_prefix

lfittl:
Sometimes log lines can be excessively long, but the log_line_prefix is almost always a tiny portion of such log lines. Since we are currently doing Regexp-based parsing for log lines (before analysis), it can be expensive to pass such large log lines in full to the Go regexp engine.

Instead, only pass the first 1000 characters to the Regexp engine when reading out the log_line_prefix fields. In a real world test case this improves the parsing performance for a 2.1GB log file from 17 minutes to 14 minutes (tested with "time pganalyze-collector --analyze-logfile"), and its hard to imagine a case where the prefix is longer than 1000 characters.

### DIFF
--- a/logs/parse.go
+++ b/logs/parse.go
@@ -125,6 +125,16 @@ func ParseLogLineWithPrefix(prefix string, line string) (logLine state.LogLine, 
 
 	rsyslog := false
 
+	// Only read the first 1000 characters of a log line to parse the log_line_prefix
+	//
+	// This reduces the overhead for very long loglines, because we don't pass in the
+	// whole line to the regexp engine (twice).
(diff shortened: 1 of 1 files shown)


msakrejda (approved):
Seems reasonable. If we do happen to run across a longer prefix, we'll fail to match and treat the line as unknown?

lfittl:
> Seems reasonable. If we do happen to run across a longer prefix, we'll fail to match and treat the line as unknown?

Yeah, that's correct - we would essentially treat the line as being part of the previous line (and then ignore it if filter settings are active and the text seems out of place)

PR #375, opened 2023-02-03
Address the case when the previous state is not found

keiko713:
Bugfix: Use the current stats value as a diff-ed value when the previous state is not found for no diff required values

When we introduced some new values for relation and database stats, we forgot to address these cases. Stats data is using lots of diff-ed values (aka previous values are required), but some of the values, we're just passing the current value.
We need to make sure to initialize these stats with the current value if no previous value is found, to avoid the value falsely becoming 0.


lfittl (approved):
:+1:

msakrejda (commented):
:+1:

PR #376, opened 2023-02-16
Return zero as FullFrozenXID for replicas

keiko713:
We currently set `currentXactId` to 0 when it's in recovery mode (aka replicas). This PR fixes the issue of the `FullFrozenXID` becomes negative when 0 is passed as `currentXactId`. Instead, `FullFrozenXID` will become zero.

This will introduce a slight inconsistency for what we return as the data. Let's think about the following case:

* current xact id is epoch 1 + 300, multixact id is 30
* database/table
  * frozen_xid 100, age 200
  * min_mxid 10, age 20

For non-replica, it'll return:
* current xact id: (1 << 32) + 300, multixact id: 30
* database
  * frozen_xid 100, age 200
  * min_mxid 10, age 20
* table
  * frozen_xid (1 << 32) + 100, age 200
  * min_mxid 10, age 20

For replica, with this PR, it'll return:
* current xact id: 0, multixact id: 0
* database
  * frozen_xid 100, age 200
  * min_mxid 10, age 20
* table
  * frozen_xid 0 (previously returning -100 here), age 200
  * min_mxid 10, age 20

I think 0 is better than -100, but still not sure if it's the best. The alternative we _could_ do is: return 100 there (and forget about epoch).

### DIFF
--- a/state/postgres_relations.go
+++ b/state/postgres_relations.go
@@ -100,7 +100,7 @@ func (i PostgresIndex) Fillfactor() int32 {
 
 // FullFrozenXID - Returns frozenXid in 64-bit FullTransactionId, by calculating and adding an epoch from current transaction ID
 func (r PostgresRelation) FullFrozenXID(currentXactId int64) int64 {
-	if r.FrozenXID == 0 {
+	if r.FrozenXID == 0 || currentXactId == 0 {
 		return 0
 	}
 	// If we simply shift the currentXactId, it'll give the epoch of the current transaction ID, which may be different
--- a/state/postgres_relations.go
+++ b/state/postgres_relations.go
@@ -100,6 +100,7 @@ func (i PostgresIndex) Fillfactor() int32 {
 
 // FullFrozenXID - Returns frozenXid in 64-bit FullTransactionId, by calculating and adding an epoch from current transaction ID
 func (r PostgresRelation) FullFrozenXID(currentXactId int64) int64 {
+	// FrozenXID can be 0 if the relation is not a table, currentXactId can be 0 on replicas
 	if r.FrozenXID == 0 || currentXactId == 0 {
 		return 0
 	}
--- a/state/postgres_relations_test.go
+++ b/state/postgres_relations_test.go
@@ -35,6 +35,11 @@ var fullFrozenXIDTests = []struct {
 		0,
 		0,
 	},
+	{
+		0,
+		12345,
+		0,
+	},
 }
 
 func TestFullFrozenXID(t *testing.T) {


lfittl:
@keiko713 Hmm. I wonder if there is a reason one would want to have freezing-type information on a replica in general? (as freezing is something the primary does, the replica is just getting the results of that activity)

Two thoughts:

1) I think `0` for the table's `frozen_xid` is a good value here, since it makes it clear that you're not getting useful information (i.e. much better than returning a negative number) -- thus +1 to this PR

2) Do need to do anything about databases? (even though you listed `database: frozen_xid 100` I think we don't actually work with the raw `pg_database.datfrozenxid` right now, only the age, correct?)

state/postgres_relations.go
```diff
@@ -100,7 +100,7 @@ func (i PostgresIndex) Fillfactor() int32 {
 
 // FullFrozenXID - Returns frozenXid in 64-bit FullTransactionId, by calculating and adding an epoch from current transaction ID
 func (r PostgresRelation) FullFrozenXID(currentXactId int64) int64 {
-	if r.FrozenXID == 0 {
+	if r.FrozenXID == 0 || currentXactId == 0 {
```
lfittl, line 103: Might be worth noting that `currentXactId == 0 ` happens when you're on a replica (for future readers of the code)

lfittl (approved):

keiko713:
> I wonder if there is a reason one would want to have freezing-type information on a replica in general? (as freezing is something the primary does, the replica is just getting the results of that activity)

Yeah that's my thoughts too, maybe it could confuse people when we present the replica data in the UI even (thought we can exclude them in the UI side). It is indeed one option to make all of these values zero, but that could make the code/SQL quite complicated I feel.

> Do need to do anything about databases? (even though you listed `database: frozen_xid 100` I think we don't actually work with the raw `pg_database.datfrozenxid` right now, only the age, correct?)

Do you mean that we should also try to make it zero? Internally, we don't store the historical data of `datfrozenxid`, and this data is stored at the database information level indeed, so we don't really use them (and yes, we do use age). So, to keep "the current behavior", I'd vote no for changing it.
To be more accurate, we also still store `relfrozenxid` (not the full one) as the table information level (non historical data), and that part just remains as 100 too.

lfittl:
> Do you mean that we should also try to make it zero? Internally, we don't store the historical data of `datfrozenxid`, and this data is stored at the database information level indeed, so we don't really use them (and yes, we do use age). So, to keep "the current behavior", I'd vote no for changing it. To be more accurate, we also still store `relfrozenxid` (not the full one) as the table information level (non historical data), and that part just remains as 100 too.

Ah, yes, makes sense. Yeah, I don't think we need to make any changes to the current handling then.

+1 to ship this PR as is :)

PR #377, opened 2023-02-17
Collect xmin horizon metrics

keiko713:
With this PR, we start collecting xmin horizon metrics as a Postgres server level stats.

Having the xmin horizon metrics is going to be beneficial to understand what to expect for the vacuums. We're going to capture this with 4 different types of xmin, categorized in a [blog post by Laurenz Albe](https://www.cybertec-postgresql.com/en/reasons-why-vacuum-wont-remove-dead-rows/):

* Long running transactions/backends: `pg_stat_activity`
* Abandoned replication slots: `pg_replication_slots`
* Orphaned prepared transactions: `pg_prepared_xacts`
* Standby server with `hot_standby_feedback = on`: `pg_stat_replication`

All of these xmin will be sent as Xid8 (full transaction ID) shape, and also only collected with non-replicas.

See also 5 mins of Postgres post for more details on xmin horizon: https://pganalyze.com/blog/5mins-postgres-autovacuum-dead-tuples-not-yet-removable-postgres-xmin-horizon

snapshot 🍐 ing PR: https://github.com/pganalyze/collector-snapshot/pull/28


input/postgres/server_stats.go
```diff
@@ -21,12 +21,64 @@ SELECT
...
+
 func GetServerStats(logger *util.Logger, db *sql.DB, postgresVersion state.PostgresVersion, systemType string) (state.PostgresServerStats, error) {
 	var stats state.PostgresServerStats
 	var transactionIdSQL string
 
-	// Only collect xact ID related stats with non-replicas
+	// Only collect transaction ID or xmin horizon related stats with non-replicas
 	if isReplica, err := getIsReplica(db); err == nil && !isReplica {
+		// Query xmin horizon before querying the current transaction ID
+		// as the backend_xmin from pg_stat_activity can point to the "next" transaction ID.
```
keiko713, line 73: [note for reviewer] This behavior is interesting, but apparently, when you do `select * from pg_stat_activity`, that backend will get the "next" current transaction ID as `backend_xmin`.
This means, if there is no other queries having backend_xmin or backend_xid, the result of `xmin_horizion_backend` becomes the next current transaction ID.
I wasn't able to find out why this happens in the code (though I wasn't looking hard), but I was able to continuously reproduce so I think this order is the safe thing to have.
(If it's the other order, the `xmin_horizion_backend` will be sent as a negative value as it's bigger than the current transaction ID)

state/postgres_server_stats.go
```diff
@@ -1,7 +1,42 @@
...
+import (
+	"database/sql"
+)
+
 // PostgresServerStats - Statistics for a Postgres server.
 type PostgresServerStats struct {
 	CurrentXactId   Xid8
 	NextMultiXactId Xid8
+
+	XminHorizonBackend                sql.NullInt32
```
keiko713, line 12: [note for reviewer] I was trying to use `Xid` (or `Xid8`) here, but often these xmin horizon can be null. When it's null, I could use `COALESCE`, though I wasn't really able to create `0` xid (looks like it's hard to return an artificial transaction ID in the query), so I decided to take care of that in golang side.

input/postgres/server_stats.go
keiko713: I think I started understanding why this happen. `backend_xmin` is always the "next" current transaction ID because all transaction IDs less than xmin are either committed and visible, or rolled back and dead. Meaning, xmin is the ID that is _not_ committed. When there is no other transaction is running, the next ID will be the one that is not committed.

This is a nature of xmin, so tweaking the order is kind of a workaround of this, but I think it's good enough so I'll leave it as is.

input/postgres/server_stats.go
```diff
@@ -21,12 +21,64 @@ SELECT
 FROM pg_catalog.pg_control_checkpoint()
 `
 
+const xminHorizonSQL string = `
+WITH slots AS (
+	SELECT
+		xmin,
+		catalog_xmin
```
lfittl, line 28: Don't we need to query for catalog_xmin separately? (since technically catalog_xmin can be further behind than xmin, and so we need the sort order to be specific to the value we're getting)

i.e. not have this CTE, but instead just query `pg_replication_slots` twice (once ordered by age(xmin) for getting xmin, the other time ordered by age(catalog_xmin) for getting catalog_xmin)

state/postgres_server_stats.go
lfittl: Can you clarify what issue you ran into with not being able to create `0` xid on the Postgres side?

For other's context that read this, its safe to turn NULL into 0, since that's explicitly defined as being an [invalid transaction ID](https://github.com/postgres/postgres/blob/master/src/include/access/transam.h#L31).

state/postgres_server_stats.go
```diff
@@ -1,7 +1,42 @@
...
+	XminHorizonReplicationSlot        sql.NullInt32
+	XminHorizonReplicationSlotCatalog sql.NullInt32
+	XminHorizonPreparedXact           sql.NullInt32
+	XminHorizonStandby                sql.NullInt32
+}
+
+// FullXminHorizonBackend - Returns XminHorizonBackend in 64-bit FullTransactionId
+func (ss PostgresServerStats) FullXminHorizonBackend() int64 {
+	return int64(XidToXid8(Xid(ss.XminHorizonBackend.Int32), Xid8(ss.CurrentXactId)))
```
lfittl, line 21: FWIW, it seems a bit odd to me that we're using a Nullable type above, but then we just ignore whether its flagged as `Valid` or not.

We're essentially relying on the zero value here being identical with the invalid transaction ID (0), but that may not be obvious to a future reader. If we can solve that on the Postgres side instead (with COALESCE, if we can make it work), that seems like a cleaner solution.

state/postgres_types.go
```diff
@@ -21,3 +21,18 @@ type PostgresType struct {
 	EnumValues        []string
 	CompositeAttrs    [][2]string
 }
+
+// XidToXid8 - Converts Xid (32-bit transaction ID) to Xid8 (64-bit FullTransactionId)
+// by calculating and adding an epoch from the current transaction ID
+func XidToXid8(xid Xid, currentXactId Xid8) Xid8 {
```
lfittl, line 27: Nice refactoring!

Should we put this in another file though? `state/postgres_types.go` refers to "custom types the collector retrieved", so its not really the same as "type definitions as defined by Postgres".

Maybe `state/util.go` would be a good file?

input/postgres/server_stats.go
```diff
@@ -21,12 +21,64 @@ SELECT
...
+	SELECT
+		transaction AS xmin
+	FROM pg_prepared_xacts
+	ORDER BY age(transaction) DESC
+	LIMIT 1
+) as prepare_xact,
+(
+	SELECT
+		backend_xmin
+	FROM pg_stat_replication
```
lfittl, line 59: I think we may need to handle our `get_stat_replication` helper function here, as described in our documentation (e.g. on https://pganalyze.com/docs/install/self_managed/01_create_monitoring_user) and already handled that way in `input/postgres/replication.go`.

That said, as a side note, I've been meaning for a while to review whether we actually need to ask people to create that helper. Re-reading things online, `pg_monitor` should include `pg_stat_replication`. So maybe this is outdated and we should actually drop that helper function from our docs? (it could still be helpful to support it here, for consistency with the other file)

Also, on the note of permissions, I wonder if we have any permission problems on e.g. systems like Heroku Postgres that don't give access to `pg_monitor`? (i.e. what happens when you query these tables there?)

lfittl (commented):
Looks good overall - some minor comments, but I think its close to 🚢

state/postgres_server_stats.go
keiko713: I was trying to find out how to cast integer to xid or generating an artificial xid and I wasn't able to find how to do, but I was clearly not searching well.
While I was working on something else today, I found this technique, which I can totally use for this. I just needed to use text instead of int 🙃 I will update PR to use this and COALESCE 👍 

```sql
pgaweb=# select 0::xid;
ERROR:  cannot cast type integer to xid
LINE 1: select 0::xid;
                ^
pgaweb=# select '0'::xid;
 xid 
-----
   0
(1 row)
```

input/postgres/server_stats.go
keiko713: Very good point, thanks for pointing this out! Similar to this, I should also tweak the `backend_xmin` part to actually return either `backend_xmin` or `backend_xid`, whatever the smaller (instead of returning backend_xmin, which can be null there if the `backend_xid` is the oldest one). I'll tweak 👍

input/postgres/server_stats.go
keiko713: Noted on `get_stat_replication`, regardless of potentially dropping it in the future, I think it's good to be consistent to the current code, so I'll make a tweak to support it 👍

Re: permission, good call, I'll check with Heroku Postgres and see if this query goes well.

keiko713: I checked with Heroku Postgres and Aurora Postgres, there were no issue running this query with both of them. I also checked with a brand new user with pretty much zero permission but it can still run this query ✨ 

btw, for Heroku, with `\du`, I was able to see that the user is a member of `pg_monitor`. Is "systems like Heroku Postgres that don't give access to `pg_monitor`" still true?

input/postgres/server_stats.go
```diff
@@ -21,12 +22,76 @@ SELECT
...
+	FROM pg_replication_slots
+	WHERE xmin IS NOT NULL
+	ORDER BY age(xmin) DESC
+	LIMIT 1
+), '0'::xid) as replication_slot_xmin,
+COALESCE((
+	SELECT
+		catalog_xmin
+	FROM pg_replication_slots
+	WHERE xmin IS NOT NULL
```
lfittl, line 50: Should this be `WHERE catalog_xmin IS NOT NULL`? (not sure if there are cases where one is set and not the other?)

input/postgres/server_stats.go
```diff
@@ -21,12 +22,76 @@ SELECT
...
 
+const xminHorizonSQL string = `
+SELECT
+COALESCE((
+	SELECT
+		CASE WHEN COALESCE(age(backend_xid), 0) > COALESCE(age(backend_xmin), 0)
+			THEN backend_xid
+			ELSE backend_xmin
+		END
+	FROM pg_stat_activity
```
lfittl, line 33: I missed that in my earlier review, but for security reasons, its best practice to explicitly use the `pg_catalog` schema, i.e. `pg_catalog.pg_stat_activity`.

Essentially the threat vector here is that someone could create a `pg_stat_activity` view that calls a function in the public schema (creating objects in the "public" schema is something that a lot more users can do in older Postgres versions, if one doesn't explicitly lock down the schema), and then trick the pganalyze collector user into calling that function, e.g. via this query.

(in practice, because `pg_catalog` is listed first in the search path by default, this is not a practical problem in almost all configurations - the issue only happens when someone explicitly includes `pg_catalog` at the end of their search path)

lfittl (approved):
Two small remaining comments, but I think this is good to merge!

input/postgres/server_stats.go
lfittl: Thanks for checking! And maybe this is no longer true for Heroku. I recall Aiven might be another provider that does not give `pg_monitor` access. But I don't think we need to do further testing for now (AFAIK the way insufficient privileges would show here is that we just don't get any rows, which seems fine).

input/postgres/server_stats.go
msakrejda: Right, good point. We need to do that for function invocations, too (e.g., `pg_catalog.age`).

msakrejda (approved):

input/postgres/server_stats.go
keiko713: thanks! addressed, I was also using `age` in the other place so I fixed there too.

PR #380, opened 2023-02-25
Stop recording FrozenXid and MinimumMultixactXid in RelationInformation

msakrejda:
See https://github.com/pganalyze/collector-snapshot/pull/29


msakrejda:
Hmm, it looks like there's some unrelated changes here because https://github.com/pganalyze/collector-snapshot/pull/28 landed but https://github.com/pganalyze/collector/pull/377 has not yet. I'll wait until the latter lands and rebase to keep the history cleaner--there's no big rush here.

PR #382, opened 2023-02-25
Update from github.com/golang/protobuf to google.golang.org/protobuf

lfittl:
The old module, github.com/golang/protobuf, has been deprecated a while, and the newer API has a slightly improved interface for the timestamp type as well, avoiding a potential error that needs to be handled (that we've traditionally ignored anyway in our own code).

Note there is a remaining caller of the old module in Google's appengine module, which gets vendored in unnecessarily via x/oauth2 (https://github.com/golang/oauth2/issues/615).

In passing, update the protoc-gen-go tool as well to the version hosted in the new repository.

Depends on https://github.com/pganalyze/collector-snapshot/pull/30 for the "go_package" option which is now required by protoc-gen-go.


lfittl:
The tests here currently fail because #380 isn't merged yet, but this PR uses newer collector-snapshot protobuf definitions introduced by that PR.

keiko713 (approved):
🧞

PR #384, opened 2023-02-26
Introduce consistent context handling and use for keeping snapshot runtime in control

lfittl:

### DIFF
--- a/input/postgres/explain.go
+++ b/input/postgres/explain.go
@@ -1,6 +1,7 @@
 package postgres
 
 import (
+	"context"
 	"database/sql"
 	"encoding/json"
 	"fmt"
@@ -13,7 +14,7 @@ import (
 	"github.com/pganalyze/collector/util"
 )
 
-func RunExplain(server *state.Server, inputs []state.PostgresQuerySample, collectionOpts state.CollectionOpts, logger *util.Logger) (outputs []state.PostgresQuerySample) {
+func RunExplain(ctx context.Context, server *state.Server, inputs []state.PostgresQuerySample, collectionOpts state.CollectionOpts, logger *util.Logger) (outputs []state.PostgresQuerySample) {
 	var samplesByDb = make(map[string]([]state.PostgresQuerySample))
 
 	skip := func(sample state.PostgresQuerySample) bool {
@@ -40,19 +41,27 @@ func RunExplain(server *state.Server, inputs []state.PostgresQuerySample, collec
 	}
 
 	for dbName, dbSamples := range samplesByDb {
-		db, err := EstablishConnection(server, logger, collectionOpts, dbName)
+		db, err := EstablishConnection(ctx, server, logger, collectionOpts, dbName)
 
 		if err != nil {
 			logger.PrintVerbose("Could not connect to %s to run explain: %s; skipping", dbName, err)
 			continue
 		}
-		useHelper := StatsHelperExists(db, "explain")
+		useHelper := StatsHelperExists(ctx, db, "explain")
 		if useHelper {
 			logger.PrintVerbose("Found pganalyze.explain() stats helper in database \"%s\"", dbName)
 		}
 
-		dbOutputs := runDbExplain(db, dbSamples, useHelper)
-		db.Close()
+		dbOutputs, err := runDbExplain(ctx, db, dbSamples, useHelper)
+		if err != nil {
+			logger.PrintVerbose("Failed to run explain: %s", err)
+			continue
+		}
+		err = db.Close()
+		if err != nil {
+			logger.PrintVerbose("Failed to close database connection: %s", err)
+			continue
+		}
 
 		hasPermErr := false
 		for _, sample := range dbOutputs {
@@ -61,7 +70,7 @@ func RunExplain(server *state.Server, inputs []state.PostgresQuerySample, collec
 				break
 			}
 		}
-		if hasPermErr && !connectedAsSuperUser(db, server.Config.SystemType) {
+		if hasPermErr && !connectedAsSuperUser(ctx, db, server.Config.SystemType) {
 			logger.PrintInfo("Warning: pganalyze.explain() helper function not found in database \"%s\". Please set up"+
 				" the monitoring helper functions (https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user)"+
 				" in every database you want to monitor to avoid permissions issues when running log-based EXPLAIN.", dbName)
@@ -72,8 +81,10 @@ func RunExplain(server *state.Server, inputs []state.PostgresQuerySample, collec
 	return
 }
 
-func runDbExplain(db *sql.DB, inputs []state.PostgresQuerySample, useHelper bool) (outputs []state.PostgresQuerySample) {
+func runDbExplain(ctx context.Context, db *sql.DB, inputs []state.PostgresQuerySample, useHelper bool) ([]state.PostgresQuerySample, error) {
+	var outputs []state.PostgresQuerySample
 	for _, sample := range inputs {
+		var isUtil []bool
 		// To be on the safe side never EXPLAIN a statement that can't be parsed,
 		// or multiple statements in one (leading to accidental execution)
 		isUtil, err := util.IsUtilityStmt(sample.Query)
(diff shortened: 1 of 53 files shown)


seanlinsley:
Is there not a way to move `Context` to a global variable to avoid passing it everywhere, or otherwise provide a `sql.DB` connection wrapper that also holds onto the context?

main.go
```diff
@@ -105,66 +104,74 @@ func run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
...
+				success := doLogTest(ctx, servers, globalCollectionOpts, logger)
+				testRunSuccess <- success
+			} else {
+				var allFullSuccessful bool
+				var allActivitySuccessful bool
+				allFullSuccessful = runner.CollectAllServers(ctx, servers, globalCollectionOpts, logger)
+				if hasAnyActivityEnabled && ctx.Err() == nil {
+					allActivitySuccessful = runner.CollectActivityFromAllServers(ctx, servers, globalCollectionOpts, logger)
+				} else {
+					allActivitySuccessful = true
```
msakrejda, line 139: Is this still right? We set `allActivitySuccessful = true` if `ctx.Err() != nil`?

main.go
```diff
@@ -105,66 +104,74 @@ func run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
...
+				}
+				if hasAnyLogsEnabled && ctx.Err() == nil {
+					// We intentionally don't fail for the regular test command if the log test fails, since you may not
+					// have Log Insights enabled on your plan (which would fail the log test when getting the log grant).
+					// In these situations we still want --test to be successful (i.e. issue a reload), but --test-logs
+					// would fail (and not reload).
+					doLogTest(ctx, servers, globalCollectionOpts, logger)
+				}
+
+				success := allFullSuccessful && allActivitySuccessful && ctx.Err() == nil
```
msakrejda, line 149: Oh I see re: above. Still, this seems like a misleading way of handling that...

scheduler/scheduler.go
```diff
@@ -12,29 +12,33 @@ type Group struct {
...
 			case <-ctx.Done():
 				return
 			case <-time.After(delay):
-				// NOTE: In the future we'll measure the runner's execution time
-				// and decide the next scheduling interval based on that
-				runner()
+				func() {
+					// Cancel runner at latest right before next scheduled execution should
+					// occur, to prevent skipping over runner executions by accident.
+					ctx, cancel := context.WithDeadline(ctx, nextExecutions[1].Add(-1*time.Second))
```
msakrejda, line 30: Should we make this more conservative? If we're stuck in a slow query, our timeout is considerably larger than that...

msakrejda (commented):
Overall, this looks reasonable to me aside from some minor comments. I do share Sean's concern that this is pretty verbose. We're making some already-sprawling parameter lists even larger. I think a context global would be too hard to debug, but maybe wrapping `sql.DB` could make this cleaner? I'm not entirely opposed to doing it as is, but I think we should consider other options.

lfittl:
> Is there not a way to move `Context` to a global variable to avoid passing it everywhere, or otherwise provide a `sql.DB` connection wrapper that also holds onto the context?

Unfortunately this is best practice - I agree its a bit verbose, but the Go documentation [explicitly describes that context is supposed to be used this way](https://pkg.go.dev/context):

> Programs that use Contexts should follow these rules to keep interfaces consistent across packages and enable static analysis tools to check context propagation:
>
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx: 
> ...
> Use context Values only for request-scoped data that transits processes and APIs, not for passing optional parameters to functions.

The only way I could see to reduce the verbosity would be to move the logger into context as a value, instead of passing it down as a separate value (we don't always have logger present today, but its there in like ~90% of cases, and its been annoying to remember to pass it in).

I don't think sql.DB is a good fit because it carries more state (i.e. needs to be closed, etc), and doing the inverse (i.e. making a struct and putting both context and sql.DB in there) is explicitly against the rules on how context is supposed to be used.

scheduler/scheduler.go
lfittl: The hard part here is that I wanted to find something that also works for activity snapshots (every 10 seconds) - which wouldn't be the case if we made this e.g. 5 seconds.

Interestingly enough, the Go database driver will actually send a cancellation request to Postgres when the context timeout is reached (it seems from my testing), so we are not relying on the statement timeout in cases where the context has reached its deadline.

input/postgres/explain.go
```diff
@@ -126,6 +151,9 @@ func runDbExplain(db *sql.DB, inputs []state.PostgresQuerySample, useHelper bool
 		outputs = append(outputs, sample)
 	}
 
+	// In case we ignored an EXPLAIN error last, avoid reporting it up
+	err = nil
```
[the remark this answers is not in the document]
lfittl: Yeah, that's a fair point, this is indeed brittle. I'll instead explicitly pass errors on early `return` (and not have the return variable be called `err`), that should make it better.

input/postgres/schema.go
```diff
@@ -35,14 +45,25 @@ func CollectAllSchemas(server *state.Server, collectionOpts state.CollectionOpts
...
 			if collectionOpts.TestRun {
 				logger.PrintWarning(warning, dbName, err)
 			} else {
 				logger.PrintVerbose(warning, dbName, err)
 			}
+			// If the schema context failed, stop doing any further collection.
+			// We avoid returning an error in this case to allow other collector
+			// functions to report their data, and send any schema information
+			// we already collected.
+			if ctxSchema.Err() != nil {
```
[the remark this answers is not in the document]
lfittl: Yeah, that's a fair point - I'll adjust to always output a warning in this case.

lfittl:
@seanlinsley @msakrejda Alright, I've done a bit more research on the appropriate use of context here.

First of all, I thought this snippet from the [Go Blog](https://go.dev/blog/context) relevant:

> At Google, we require that Go programmers pass a Context parameter as the first argument to every function on the call path between incoming and outgoing requests. This allows Go code developed by many different teams to interoperate well. It provides simple control over timeouts and cancellation and ensures that critical values like security credentials transit Go programs properly.

Now, of course we don't have to do things like Google does them (and often times that's a bad idea), but I think this does show that the verbosity of this extra parameter is very much by design. Given how Go does other things (e.g. error handling), it makes sense to me that it would be done this way, preferring an explicit verbose approach.

Assuming we agree that passing context as the first parameter is an acceptable trade-off here, the next question is: Should we use it to pass down additional values?

I found [this article](https://www.calhoun.io/pitfalls-of-context-values-and-how-to-avoid-or-mitigate-them/) to be a good overview of best practices, and it specifically addresses logger and database connections being a context value:

> You shouldn’t store a logger there if it isn’t created specifically to be scoped to this request, and likewise you shouldn’t store a generic database connection in a context value.

Now it goes on to talk more about this, but I think the main argument for why its a bad idea really comes down to this:

> We lose type safety at compile time
> ..
> Context values are stored in what is essentially an interface{}, interface{} pair (see [the source](https://github.com/golang/go/blob/986768de7fcf4def65cecd7eb0c34e2cdf92e78c/src/context/context.go#L480)). That is why we are permitted to store anything and everything on our inside this type without the compiler complaining - both the key and the value are defined as interfaces and will accept quite literally any val

And this loss of type safety of course also comes at an inconvenience to the function wanting to use a context value (since it has to do a type assertion).

Overall, I think this does seem like the correct way to approach this, at the expense of some added verbosity.

I think what we could consider separately as a future improvement (not in this PR), is to have our input methods use a struct that carries around information like sql.DB, logger, the server reference, and e.g. the Postgres version in use, to avoid the endless long list of parameters.

seanlinsley:
I disagree with Google's decision here – type assertion at time of use seems fine as a tradeoff to using an interface, otherwise a typed struct would be preferable.

Thanks for providing the additional context here. Whatever we decide on, I don't think it blocks this PR.

scheduler/scheduler.go
msakrejda: Okay. We should check if this is still the case with our new driver if we migrate off `pq` at some point.

msakrejda (approved):
Okay. It does sound like the Context usage is idiomatic, so while maybe we should still rethink the interfaces here, that's separate from the motivation for this patch.

PR #386, opened 2023-02-28
Fix Azure Single and Flexible Server log line parsing for DETAIL lines

lfittl:
The code previously did not work correctly since the log_line_prefix was not present on the DETAIL lines (which are included on the main log line on Azure). This subsequently caused issues with log-based EXPLAIN not having DETAIL lines available.

Fix by reworking how log lines get transformed for Azure, and adding unit tests for the transformation based on examples from running "pgbench -M extended".

### DIFF
--- a/input/system/azure/logs.go
+++ b/input/system/azure/logs.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/pganalyze/collector/config"
 	"github.com/pganalyze/collector/logs"
+	"github.com/pganalyze/collector/output/pganalyze_collector"
 	"github.com/pganalyze/collector/state"
 	"github.com/pganalyze/collector/util"
 	uuid "github.com/satori/go.uuid"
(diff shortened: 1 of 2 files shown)


input/system/azure/logs.go
```diff
@@ -198,30 +205,30 @@ func setupLogTransformer(ctx context.Context, wg *sync.WaitGroup, servers []*sta
...
 
 				foundServer := false
 				for _, server := range servers {
-					if in.LogicalServerName == server.Config.AzureDbServerName {
-						out <- state.ParsedLogStreamItem{Identifier: server.Config.Identifier, LogLine: logLine}
+					if azureDbServerName == server.Config.AzureDbServerName {
 						foundServer = true
+
+						for _, logLine := range logLines {
+							logLine.CollectedAt = time.Now()
```
keiko713, line 228: just curious: if it matters to have the same `collectedAt` across these `logLines` 🤔 it is kinda collected at the same time

lfittl: Its only used for expiring data in the stream handling logic, so it shouldn't matter.

(but good catch to notice that difference!)

keiko713 (approved):
🏁

PR #388, opened 2023-02-28
Release task: Use Ubuntu 20.04 release to release binary with older glibc

lfittl:
GitHub actions automatically updated the "ubuntu-latest" release to 22.04, which uses a fairly new glibc that breaks use of the release binary on older operating systems.

Since we're not in a rush to build on 22.04 for the release task, pin to 20.04 to ensure an older minimum required glibc version.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,9 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    # Use older Ubuntu release to avoid building with a too new glibc, which will cause
+    # errors on older distributions, if anyone uses the generated binary directly.
+    runs-on: ubuntu-20.04
     permissions: {}
     # ensure we don't release on create events for branches (only tags)
     if: ${{ startsWith( github.ref, 'refs/tags' ) }}


lfittl:
As a side note, I've done a manual rebuild of the collector binary [on the last release](https://github.com/pganalyze/collector/releases/tag/v0.49.0) - so this is mainly intended to fix this going forward.

msakrejda (approved):

.github/workflows/release.yml
```diff
@@ -7,7 +7,9 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    # Use older Ubuntu release to avoid building with a too new glibc, which will cause
+    # errors on older distributions, if anyone uses the generated binary directly.
```
msakrejda, line 11: Is the converse not a problem? That is, using the older-build binary on newer distributions?

lfittl: Works fine - glibc is a minimum requirement (based on the version its built with), but you can have a newer glibc release on the host and it will know what to do.

PR #389, opened 2023-03-01
README: Add missing GRANT USAGE for helper functions

lfittl:
Whilst we're planning to overhaul this section of the README soon, there is a very common problem that one can run into if you forget to grant the pganalyze user usage on the schema for the restricted monitoring user - fix that since its straightforward to do.

This matches documentation pages such as https://pganalyze.com/docs/install/amazon_rds/02_create_monitoring_user


keiko713 (approved):
👤

PR #391, opened 2023-03-05
Avoid partial data and bogus statistics when locks are encountered in parts of the snapshot

lfittl:
```
Relation queries: Correctly handle later queries encountering a lock

The existing coding assumed that we can rely on the fact that a relation
is exclusively locked to stay the same across each query that works with
relation data to get columns, indexes, constraints, etc.

But that is not true (locks can change between the queries) and that
causes a race condition where the first query does not see a relation
being locked, but later queries do. This then causes downstream problems
since only part of the information was sent, and e.g. an index might
show in one snapshot, and then disappear in the next, if locking
conditions are unlucky.

To fix, return which exclusive locks each relation query encountered, and
take that to consider the relation having been exclusively locked for the
whole current snapshot, and set the flag accordingly.

In passing, add explicit checks on subsequent relation queries to
ensure the relation actually was captured in the first query. This is
to protect against locks being held on relations that we ignored earlier
(e.g. because of them being a catalog table, or matching the schema filter)
suddenly showing up with empty data through a later query returning them
as a locked relation.

Note that the relation statistics logic suffers from a different but
related problem which is to be fixed in a subsequent commit.
```

```
Relation statistics: Avoid bogus data due to diffs against locked objects

Previously we would skip locked relations when getting the statistics for
either the table or an index. Compared to the locking race condition
in the relation info queries, this is actually not directly a problem in
the same way.

However, the problem with the stats queries is that they assume they
can diff against the prior record, and if that doesn't exist, that the
table (or index) was recently added, and thus we can diff against zero.

That is not correct when relations were ignored due to locks, leading
to a subsequent stats collection suddenly seeing the full counter value
as the diff (since the prior data point was skipped due to the lock).

Fix by explicitly tracking locked relations that the stats queries
skipped, and applying that knowledge in the diff handling.
```

### DIFF
--- a/input/postgres/relation_stats.go
+++ b/input/postgres/relation_stats.go
@@ -54,7 +54,8 @@ SELECT c.oid,
 			 CASE WHEN c.relminmxid <> '0' THEN pg_catalog.mxid_age(c.relminmxid) ELSE 0 END AS relation_mxid_age,
 			 c.relpages,
 			 c.reltuples,
-			 c.relallvisible
+			 c.relallvisible,
+			 false AS exclusively_locked
 	FROM pg_catalog.pg_class c
 	LEFT JOIN pg_catalog.pg_namespace n ON (n.oid = c.relnamespace)
 	LEFT JOIN pg_catalog.pg_stat_user_tables s ON (s.relid = c.oid)
@@ -65,6 +66,44 @@ SELECT c.oid,
 			 AND c.oid NOT IN (SELECT pd.objid FROM pg_catalog.pg_depend pd WHERE pd.deptype = 'e' AND pd.classid = 'pg_catalog.pg_class'::regclass)
 			 AND n.nspname NOT IN ('pg_catalog','pg_toast','information_schema')
 			 AND ($1 = '' OR (n.nspname || '.' || c.relname) !~* $1)
+ UNION ALL
+SELECT relid,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   NULL,
+	   NULL,
+	   NULL,
+	   NULL,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   true AS exclusively_locked
+  FROM locked_relids_with_parents
 `
 
 const indexStatsSQL = `
@@ -75,11 +114,22 @@ SELECT s.indexrelid,
 			 COALESCE(s.idx_tup_read, 0),
 			 COALESCE(s.idx_tup_fetch, 0),
 			 COALESCE(sio.idx_blks_read, 0),
-			 COALESCE(sio.idx_blks_hit, 0)
+			 COALESCE(sio.idx_blks_hit, 0),
+			 false AS exclusively_locked
 	FROM pg_catalog.pg_stat_user_indexes s
 			 LEFT JOIN pg_catalog.pg_statio_user_indexes sio USING (indexrelid)
  WHERE s.indexrelid NOT IN (SELECT relid FROM locked_relids)
 			 AND ($1 = '' OR (s.schemaname || '.' || s.relname) !~* $1)
+UNION ALL
+SELECT relid,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   true AS exclusively_locked
+  FROM locked_relids
 `
 
 const columnStatsSQL = `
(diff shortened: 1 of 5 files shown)


input/postgres/relation_stats.go
```diff
@@ -65,6 +66,44 @@ SELECT c.oid,
...
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
```
lfittl, line 104: [Note for reviewer] I'm not a fan of this style, but don't see a better way to return this data, short of introducing something like `row_to_json` for the data (and deserializing on the client). Alternate ideas welcome :)

input/postgres/relation_stats.go
```diff
@@ -65,6 +66,44 @@ SELECT c.oid,
 			 AND c.oid NOT IN (SELECT pd.objid FROM pg_catalog.pg_depend pd WHERE pd.deptype = 'e' AND pd.classid = 'pg_catalog.pg_class'::regclass)
 			 AND n.nspname NOT IN ('pg_catalog','pg_toast','information_schema')
 			 AND ($1 = '' OR (n.nspname || '.' || c.relname) !~* $1)
+ UNION
```
msakrejda, line 69: Probably not a huge deal, but I think these could all be `UNION ALL`s, no?

msakrejda (approved):
A little less ugly than I feared.

state/postgres_relation_stats.go
```diff
@@ -3,49 +3,51 @@ package state
...
-	IdxTupFetch int64 // Number of live table rows fetched by simple index scans using this index
-	IdxBlksRead int64 // Number of disk blocks read from this index
-	IdxBlksHit  int64 // Number of buffer hits in this index
+	SizeBytes         int64
+	IdxScan           int64 // Number of index scans initiated on this index
+	IdxTupRead        int64 // Number of index entries returned by scans on this index
+	IdxTupFetch       int64 // Number of live table rows fetched by simple index scans using this index
+	IdxBlksRead       int64 // Number of disk blocks read from this index
+	IdxBlksHit        int64 // Number of buffer hits in this index
+	ExclusivelyLocked bool  // Whether these statistics are zeroed out because the index was locked at collection time
```
keiko713, line 50: isn't this also "table was locked"? (though it may read awkward a bit)

input/postgres/relation_stats.go
keiko713: agree with Maciek, I think it's okay to be like this

input/postgres/relation_stats.go
```diff
@@ -65,6 +66,44 @@ SELECT c.oid,
...
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   0,
+	   true AS exclusively_locked
+  FROM locked_relids_with_parents
```
keiko713, line 106: is there any case that `locked_relids_with_parents` can potentially include some relids that are not meeting the above condition? (e.g. relkind is not IN ('r','v','m','p'), IOW the lock was held with a TOAST (relkind = t))

keiko713: uhm, I guess it's okay to include these ones even, as we won't send these stats anyways 🤔

keiko713 (approved):
A few questions but overall looking good 👍

input/postgres/relation_stats.go
lfittl: Good point - I'll adjust to UNION ALL.

state/postgres_relation_stats.go
lfittl: In this particular case its actually the index that would have been locked, since we're checking against `indexrelid` (https://github.com/pganalyze/collector/pull/391/files#diff-89c88e46957c7be3a80002749ee13e4611914aff1acf0eb5e27d05b119e3b98aR121) to protect against `pg_relation_size(s.indexrelid)` encountering a lock.

input/postgres/relation_stats.go
lfittl: Thanks, will leave as is then - I think its probably best to keep it simple for now.

Re:

> Separately, (and outside the scope of this patch) does locked_relids_with_parents not need to be a recursive CTE?

Good question - I don't think it needs to be a recursive CTE, because it exists to protect this part of the query, which is not recursive (i.e. only checks for direct children):

```sql
COALESCE((SELECT pg_catalog.sum(pg_catalog.pg_table_size(inhrelid))
  FROM pg_catalog.pg_inherits
  WHERE inhparent = c.oid), 0) AS size_bytes,
```

input/postgres/relation_stats.go
lfittl: Yeah, it definitely could pull in relids that are otherwise ignored - as you note that is not a problem though because we wouldn't send it. I think its okay to keep the query simple like that for now.

output/transform/postgres_relations.go
```diff
@@ -65,6 +65,13 @@ func transformPostgresRelations(s snapshot.FullSnapshot, newState state.Persiste
 
 		schemaStats, schemaStatsExist := newState.SchemaStats[relation.DatabaseOid]
 
+		// In case of exclusively locked relations that are encountered by a later input query
+		// (e.g. to get column or index information), it can happen that we get partial data
+		// - make sure we don't send that unnecessarily (the server would just ignore it)
+		if relation.ExclusivelyLocked {
+			continue
```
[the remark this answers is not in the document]
keiko713: Uhmm actually this is not accurate as what is causing. The cause is more of, when we store the `RelationReferences` info in pganalyze side, we need to have the corresponding `RelationInformations`, but this `continue` is removing the corresponding `RelationInformations` that `RelationReferences` should have, therefore errored.

The simplest fix would be not to send `RelationReferences` when `relation.ExclusivelyLocked` is true, but I'm worried that it's going to cause the table to be "invalidated" (as the table is no longer there) in pganalyze side.

Maybe the best fix is to consider the case of `RelationReferences` without `RelationInformations` well in pganalyze side, I'll look into more from that perspective.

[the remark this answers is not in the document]
keiko713: Actually created a PR in collector side: https://github.com/pganalyze/collector/pull/399
I think this is actually what this PR was going to do initially.

PR #394, opened 2023-03-11
Add hint when pg_stat_statements does not exist and creation fails

msakrejda:


input/postgres/statements.go
```diff
@@ -111,6 +111,7 @@ func GetStatements(ctx context.Context, server *state.Server, logger *util.Logge
 		logger.PrintInfo("pg_stat_statements does not exist, trying to create extension...")
 		_, err = db.ExecContext(ctx, QueryMarkerSQL+"CREATE EXTENSION IF NOT EXISTS pg_stat_statements SCHEMA public")
 		if err != nil {
+			logger.PrintInfo("HINT - if you expect the extension to already be installed, please review the pganalyze documentation: https://pganalyze.com/docs/install/troubleshooting/pg_stat_statements")
```
msakrejda, line 114: Putting this here is a little ugly, because it will come before the error in the log output. Unfortunately, putting it after the error is ugly in a different way, since we'll need to put the log message in the caller, and check the returned error to see if it matches the message above.

lfittl: Do you have a full example of how this would look like with a test command?

msakrejda:
```
2023/03/10 17:53:00 I [server1] pg_stat_statements does not exist, trying to create extension...
2023/03/10 17:53:00 I [server1] HINT - if you expect the extension to already be installed, please review the pganalyze documentation: https://pganalyze.com/docs/install/troubleshooting/pg_stat_statements
2023/03/10 17:53:00 E [server1] Could not collect query stats for server: error collecting pg_stat_statements: pq: permission denied to create extension "pg_stat_statements"
```
I think this is okay, since it's given context by the preceding message.

msakrejda: Oh and, in context:

```
$ ./pganalyze-collector --test
2023/03/10 17:54:41 I Running collector test with pganalyze-collector 0.49.0
2023/03/10 17:54:41 I [server1] Testing statistics collection...
2023/03/10 17:54:41 I [server1] pg_stat_statements does not exist, trying to create extension...
2023/03/10 17:54:41 I [server1] HINT - if you expect the extension to already be installed, please review the pganalyze documentation: https://pganalyze.com/docs/install/troubleshooting/pg_stat_statements
2023/03/10 17:54:41 E [server1] Could not process server: Error collecting pg_stat_statements: pq: permission denied to create extension "pg_stat_statements"
2023/03/10 17:54:41 I [server1] Testing activity snapshots...
2023/03/10 17:54:41 I [server1]   Test snapshot successful (8.08 KB)
2023/03/10 17:54:41 I [server1] Testing log collection (local)...
2023/03/10 17:54:41 I [server1] Setting up log tail for /var/log/postgresql/postgresql-14-main.log
2023/03/10 17:54:45 I [server1]   Local log test successful
```

I did not change that the test still succeeds in this case.

lfittl: Ah, yes, because of the preceding log message this works fine. Thanks!

lfittl (approved):

PR #395, opened 2023-03-11
Amazon RDS / Aurora: Add new "rds-ca-global" option for db_sslrootcert

lfittl:
We've historically bundled the individual CAs certs that AWS publishes, to help use a verify-full connection with RDS. With the recently released set of new CA options, i.e. "rds-ca-rsa2048-g1", "rds-ca-rsa4096-g" and "rds-ca-ecc384-g1", AWS has instead started publishing a combined list of CAs in a single "global" file. This global set also includes the old "rds-ca-2019-root" (but not the 2015 root, since that is expired).

To simplify configuration, introduce the new "rds-ca-global" option for db_sslrootcert, which is the recommended configuration for RDS and Aurora going forward. For compatibility reasons we still support naming the "rds-ca-2019-root" explicitly, but its now just an alias for the global set. This also drops the "rds-ca-2015-root", which would no longer have worked anyway (since the cert expired).


config/config.go
```diff
@@ -293,11 +293,10 @@ func (config ServerConfig) GetPqOpenString(dbNameOverride string, passwordOverri
 	}
 
 	// Handle SSL certificates shipped with the collector
-	if dbSslRootCert == "rds-ca-2015-root" {
-		dbSslRootCert = "/usr/share/pganalyze-collector/sslrootcert/rds-ca-2015-root.pem"
-	}
-	if dbSslRootCert == "rds-ca-2019-root" {
-		dbSslRootCert = "/usr/share/pganalyze-collector/sslrootcert/rds-ca-2019-root.pem"
+	//
+	// Note: "rds-ca-2019-root" is a legacy cert expiring in 2024 that is part of the global CA set
```
msakrejda, line 297: Thanks for adding the comment here.

msakrejda (approved):

PR #397, opened 2023-03-12
Build collector packages using GitHub Actions

lfittl:
This simplifies the release process by utilizing GitHub Actions to build the RPM/Apt packages. Note that this only includes the build and test steps - the signing and release continue to be done separately.


.github/workflows/test-packages.yml
```diff
@@ -0,0 +1,32 @@
+name: Test Packages
```
lfittl, line 1: [note for reviewer]

The goal of this new GH action is to have a nightly test build of the packages of the current tip of main. That avoids having a late failure in the release process, which can be frustrating.

.github/workflows/release.yml
```diff
@@ -45,19 +45,63 @@ jobs:
...
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+
+    - name: Build packages
+      run: |
+        DOCKER_BUILDKIT=1 make -C packages build
+
+    - name: Test packages
+      run: |
+        DOCKER_BUILDKIT=1 make -C packages test
```
lfittl, line 72: Doing the package test here (instead of only the build) adds ~2hrs to the release task - which is not great.

However I'm not sure the alternative is a good idea (releasing without the test), so I'd suggest we do this for now, and if we get lots of issues with this we revise.

Previously I used to run this locally during the package release, so this matches the existing release process.

.github/workflows/test-packages.yml
keiko713: Will the result of this workflow be visible to anywhere? I like the idea, though I wonder if we can even notice the failure.

.github/workflows/release.yml
```diff
@@ -45,19 +45,63 @@ jobs:
...
 
+    - name: Download build packages
+      uses: actions/download-artifact@v3
+      with:
+        name: packages
+
     - name: Get version from tag
       id: get_version
-      run: echo ::set-output name=version::${GITHUB_REF#refs/tags/}
+      run: echo "version=${GITHUB_REF#refs/tags/}" >> $GITHUB_OUTPUT
```
msakrejda, line 104: For what it's worth, we should be able to use GITHUB_REF_NAME here: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

keiko713 (approved):
looks good! can you please update Release step in `CONTRIBUTING.md` too?

msakrejda (approved):

.github/workflows/test-packages.yml
lfittl: Its the same as the Enterprise nightly builds on the app repo (which also run daily) - I think the author of the last commit gets a notification (or possibly all maintainers?). Plus you can always check via the Actions UI on the repo.

lfittl:
> looks good! can you please update Release step in `CONTRIBUTING.md` too?

I will actually do this separately - this notably does not sign the packages yet and push them to the repo (that is still a manual step) -- I have an idea on how to script that better, and will update the README together with the improved script post-merge.

.github/workflows/release.yml
lfittl: Nice! Since I didn't test with that I'll leave that refactoring out for now, but good idea to clean it up that way.

PR #399, opened 2023-03-21
Send out exclusively locked relation info

keiko713:
In https://github.com/pganalyze/collector/pull/391, we stopped sending out the data like `RelationInformations` and `RelationStatistics` when the relation is exclusively locked.
However, this caused the issue on the pganalyze app side, because we're still sending out `RelationReferences` yet we weren't able to find the corresponding `RelationInformations` with it. When storing the `RelationReferences`  data, we use the information of `RelationInformations`, especially `RelationType` data.

With this change, revert stop sending out the `RelationInformations` data. This should be the only info that we need but dropped (we are okay not having `RelationStatistics` or other things), and we also shouldn't have the "bad data" with `RelationInformations` because we can still get this basic relation info without worrying about the locks.

https://github.com/pganalyze/collector/pull/391/files#r1139988321

### DIFF
--- a/output/transform/postgres_relations.go
+++ b/output/transform/postgres_relations.go
@@ -63,6 +63,16 @@ func transformPostgresRelations(s snapshot.FullSnapshot, newState state.Persiste
 			Options:                relation.Options,
 		}
 
+		// In case of exclusively locked relations that are encountered by a later input query
+		// (e.g. to get column or index information), it can happen that we get partial data
+		// - make sure we don't send that unnecessarily (the server would just ignore it)
+		if relation.ExclusivelyLocked {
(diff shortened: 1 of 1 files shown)


msakrejda (approved):
I think that's sound.

Not for this PR, but we've had issues like this before: I wonder if we should have a sanity check that the array structures we expect to be parallel are actually parallel (like RelationInformation / RelationReference)...

keiko713:
> I wonder if we should have a sanity check that the array structures we expect to be parallel are actually parallel (like RelationInformation / RelationReference)...

I'm curious if you're thinking this check as some tests, or a check before sending out the snapshot

PR #401, opened 2023-03-24
Query stats reset: Make pg_stat_statements_reset() failure a soft error

### LINKED ISSUE #398: Error calling pg_stat_statements_reset() as requested intermittently
Hi,

we have been investigating a perceived memory leak in our collector and have contributed to the previous issue that was raised (https://github.com/pganalyze/collector/issues/37) - on the back of that issue being closed we started looking into what is actually causing *some* of our collectors to randomly restart.

We see memory growing in a linear fashion until the pod restarts:

image

The above screenshot shows 2 collectors, one that faces the issue, and another that doesnt.

We then tried to match up what happens at the start of memory jump, and we were able to see an error in the logs that matches up to the start point of the memory growth on each occasion..

```
2023-03-13T02:20:23.006696777Z E [default] Error calling pg_stat_statements_reset() as requested: pq: permission denied for function pg_stat_statements_reset
2023-03-13T02:20:23.006746962Z E [default] Could not process server: pq: permission denied for function pg_stat_statements_reset
```

From what we can see, some collectors seem to think that they dont have permissions on the database and then are unable to send stats - which is strange as its all very intermittent. once the collector crashes and restarts - it all starts working again.

Once we get the error, we never see a `Submitted snapshot successfully` until the collector restarts (when it hits its memory limit). So our theory is that once this error happens, the collector is building up a snapshot locally and thus the memory grows, and as it isnt able to send it to pganalyze, we eventually hit the limit of what it can store in memory locally on the pod.

We'd like to get to the bottom of what causes it to lose its permissions as we think this is the root cause.

Thanks

lfittl:
Previously this was a hard error, which failed the snapshot - this is not desirable because hard errors mean the snapshot state does not get persisted. In practice this meant that the error would very likely occur again on the next full snapshot run, in the typical case of a permission problem. This meant that full snapshots kept failing, which indirectly led to a memory leak because the UnidentifiedStatementStats list would keep growing (as the query runner was working just fine, once a minute).

To fix, emit only a warning instead, and process the rest of the snapshot as usual, thus persisting the state and not calling the reset function again until the next scheduled interval.

Fixes #398

---

Separately, as a defense against this class of bugs causing memory leaks, we should review in a follow-up PR how we can prune UnidentifiedStatementStats in face of constantly failing full snapshots (possibly by having the query stats runner do a sanity check and drop old unsubmitted data if needed).

### DIFF
--- a/input/full.go
+++ b/input/full.go
@@ -64,8 +64,9 @@ func CollectFull(ctx context.Context, server *state.Server, connection *sql.DB, 
 		ps.StatementResetCounter = 0
 		err = postgres.ResetStatements(ctx, logger, connection, systemType)
 		if err != nil {
-			// This is a non-fatal error, so continue snapshot collection but report it
-			logger.PrintWarning("Failed to call pg_stat_statements_reset() as requested: %s", err)
+			// This is a non-fatal error, so continue snapshot collection but do log it as an error
+			// (this gets automatically added to the snapshot's CollectorErrors information)
+			logger.PrintError("Error calling pg_stat_statements_reset() as requested: %s", err)
 			err = nil
 		} else {
 			_, _, ts.ResetStatementStats, err = postgres.GetStatements(ctx, server, logger, connection, globalCollectionOpts, ts.Version, false, systemType)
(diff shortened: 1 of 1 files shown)


keiko713 (approved):
makes sense 👍

seanlinsley (approved):

PR #402, opened 2023-03-24
Add integrity checks before uploading snapshots

msakrejda:
The snapshot format makes some structural assumptions that cannot be
enforced by protobuf. If these are violated, processing the snapshot
may fail, or snapshot data may be misinterpreted and can cause bogus
statistics to be recorded.

Validate these assumptions before sending a snapshot.

See discussion in https://github.com/pganalyze/collector/pull/399#pullrequestreview-1350912675


msakrejda:
Are there any other full snapshot checks we should make? I didn't see any similar checks to make for compact snapshots, but maybe I missed something.

output/full.go
```diff
@@ -40,6 +40,12 @@ func submitFull(ctx context.Context, s snapshot.FullSnapshot, server *state.Serv
 	var err error
 	var data []byte
 
+	err = verifyIntegrity(s)
+	if err != nil {
+		logger.PrintError(fmt.Sprintf("Snapshot integrity check failed: %s; please contact support", err))
+		return err
```
lfittl, line 46: I wonder if there is an argument to be made that this should not return an error - because as I noticed when working on #401, failing the snapshot run causes the state update to be discarded (since the [relevant code portion is not run](https://github.com/pganalyze/collector/blob/main/runner/full.go#L233)) - and thus the next full snapshot would use the previous reference point (potentially running into the exact same integrity problem again, if its related to the diffing to the prior data point).

If we change this to not return an error, we should consider whether we'd still want to submit a failed snapshot (i.e. call `SendFailedFull`), which would otherwise have been done by [CollectAllServers](https://github.com/pganalyze/collector/blob/main/runner/full.go#LL221) when `SendFull` returns an error.

msakrejda: Hmm, interesting. I don't have a strong opinion on `SendFailedFull` but it seems like it might be a good idea to include these messages in the UI.

Separately, I'm also wondering if maybe we should allow ignoring the integrity check. For example, negative stats are clearly bogus but could in theory be localized, and it might be inconvenient for all server stats to stop working just because of that... I don't want to just log errors, because I think we should only check things here that are clear bugs, and if they're only logged, we won't hear about them. But we may also want an escape hatch (unless the problem is catastrophic, like the relation / info length mismatches).

lfittl: > Separately, I'm also wondering if maybe we should allow ignoring the integrity check.

Yeah, maybe? An escape hatch seems useful, but I'm also wary of adding too many flags that are not used. Maybe this is a good use case for an extra command line flag? (and not a new config setting)

keiko713: Uhmm, or we could skip negative stats check for now 🤔 I think reference + information length check is good enough as a starter and I don't think we need to allow ignoring these checks.

msakrejda: I thought about that, but I don't think omitting checks like that is beneficial: negative numbers there _are_ clearly bogus, even if they're not as problematic for snapshot processing. Still, it does have the benefit of not having to introduce a workaround (yet?), since we clearly don't want to send snapshots with critical integrity issues. Anyone else have thoughts on this?

seanlinsley: >I wonder if there is an argument to be made that this should not return an error - because as I noticed when working on https://github.com/pganalyze/collector/pull/401, failing the snapshot run causes the state update to be discarded (since the [relevant code portion is not run](https://github.com/pganalyze/collector/blob/main/runner/full.go#L233)) - and thus the next full snapshot would use the previous reference point (potentially running into the exact same integrity problem again, if its related to the diffing to the prior data point).

@lfittl is there a class of bug that prevents us from updating the state after a failed run?

keiko713: I wonder how many negative numbers are currently sent and processed. As you mentioned, they're not problematic for snapshot processing and it will allow the other potentially-good data (e.g. server metrics) to be stored. With this change, such things will be dropped.
I did some check, looks like we had 11 servers having the report of negative `n_tup_ins` within one day.

output/full.go
```diff
@@ -26,13 +26,20 @@ import (
...
+		// be able to make progress. Instead, send a failed snapshot directly.
+		return SendFailedFull(ctx, server, collectionOpts, logger)
+	} else {
+		return submitFull(ctx, s, server, collectionOpts, logger, newState.CollectedAt, false)
+	}
 }
 
 func SendFailedFull(ctx context.Context, server *state.Server, collectionOpts state.CollectionOpts, logger *util.Logger) error {
-	s := snapshot.FullSnapshot{FailedRun: true, CollectorErrors: logger.ErrorMessages}
+	s := snapshot.FullSnapshot{FailedRun: true}
```
msakrejda, line 42: I ended up moving this into `submitFull` because both callers were setting collector errors independently before invoking that helper--we might as well do it in the shared code path.

output/full.go
seanlinsley: We can always move the stats check to snapshot processing, so only that data is skipped.

keiko713 (approved):
nice! 👍

PR #406, opened 2023-04-03
Extend the deadline for activity snapshots

seanlinsley:
#384 added a deadline / timeout for each snapshot, but this is leading to data loss for a customer with a collector process that monitors 20+ servers. When monitoring so many servers it's possible that network requests, for example, could cause activity snapshots to not reliably finish within 10 seconds. A future refactor may move snapshot uploads to a separate deadline, but this PR should mitigate the issue for now.

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -27,7 +27,12 @@ func (group Group) Schedule(ctx context.Context, runner func(context.Context), l
 				func() {
 					// Cancel runner at latest right before next scheduled execution should
 					// occur, to prevent skipping over runner executions by accident.
-					ctx, cancel := context.WithDeadline(ctx, nextExecutions[1].Add(-1*time.Second))
+					deadline := nextExecutions[1].Add(-1*time.Second)
+					// Extend the deadline of very short runs to avoid pointless cancellations.
+					if nextExecutions[1].Sub(nextExecutions[0]) <= 15*time.Second {
+						deadline = deadline.Add(15*time.Second)
+					}
+					ctx, cancel := context.WithDeadline(ctx, deadline)
 					defer cancel()
 					runner(ctx)
 				}()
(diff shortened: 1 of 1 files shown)


scheduler/scheduler.go
```diff
@@ -27,7 +27,12 @@ func (group Group) Schedule(ctx context.Context, runner func(context.Context), l
 				func() {
 					// Cancel runner at latest right before next scheduled execution should
 					// occur, to prevent skipping over runner executions by accident.
-					ctx, cancel := context.WithDeadline(ctx, nextExecutions[1].Add(-1*time.Second))
+					deadline := nextExecutions[1].Add(-1 * time.Second)
+					// Extend the deadline of very short runs to avoid pointless cancellations.
+					if nextExecutions[1].Sub(nextExecutions[0]) <= 15*time.Second {
+						deadline = deadline.Add(15 * time.Second)
```
lfittl, line 33: Note that technically this would end up with 25 seconds (not 15 seconds) of total deadline time, since we're adding instead of replacing. If we wanted to have a fixed 15 seconds, we could instead say:

```golang
  if nextExecutions[1].Sub(nextExecutions[0]) <= 15*time.Second {
    deadline = nextExecutions[0].Add(15 * time.Second)
  }
```

(or possibly 19 seconds, so we allow up to one execution to be skipped, but not two?)

lfittl (approved):
:+1:

One minor comment, but otherwise looks good to ship!

scheduler/scheduler.go
seanlinsley: 👍 updated to 19 seconds

lfittl: Not a big deal in practice, but I realize that for the next reader this may be easier to understand if we also use 19 seconds in the if condition itself (i.e. to check whether to override it). I'll make a quick PR for that.

PR #408, opened 2023-04-04
packages Makefile: Fix variable default logic to not assume bash

lfittl:
The syntax added in the previous commit does not work within Makefiles, only in bash. Instead use "?=" which is the correct way to allow overriding Makefile variables from the environment.

In passing, default the GIT_VERSION to be the version prefixed by "v", to have one less variable to change when releasing a new version.


keiko713 (approved):
TIL that it won't work in Makefile

PR #410, opened 2023-04-05
Release task: Correctly use version without "v" prefix

lfittl:
This accidentally led to the package upload to the release entry not working as intended, since the package names use the raw version number without the "v" prefix used in the git tag.


keiko713 (approved):
👍

PR #415, opened 2023-05-12
Azure: Switch to new azeventhubs library with azidentity integration

lfittl:
In addition to this being the preferred library to be used going forward, this also adds support for workload identity in AKS.

Docker test image: ```quay.io/pganalyze/collector:aks-workload-identity-test```


msakrejda (approved):
Looks reasonable.

PR #416, opened 2023-05-12
Track TOAST table name, reltuples, and relpages

seanlinsley:
Tracking TOAST table names will allow us to better track VACUUM events from the logs, and tracking TOAST reltuples improves bloat calculation accuracy.


state/postgres_relation_stats.go
```diff
@@ -38,6 +38,7 @@ type PostgresRelationStats struct {
 	Reltuples         float32   // Number of live rows in the table. -1 indicating that the row count is unknown
 	Relallvisible     int32     // Number of pages that are marked all-visible in the table's visibility map
 	ExclusivelyLocked bool      // Whether these statistics are zeroed out because the table was locked at collection time
+	ToastReltuples    int32     // Number of live rows in the TOAST table. -1 indicating that the row count is unknown
```
seanlinsley, line 41: I copied the -1 comment from the other reltuples field. Is this accurate here? I think it may be zero if the table doesn't use any TOAST.

Note: be sure to change the protobuf comment as well.

output/transform/postgres_relations.go
```diff
@@ -78,6 +78,9 @@ func transformPostgresRelations(s snapshot.FullSnapshot, newState state.Persiste
 		if relation.ViewDefinition != "" {
 			info.ViewDefinition = &snapshot.NullString{Valid: true, Value: relation.ViewDefinition}
 		}
+		if relation.ToastTable != "" {
+			info.ToastTable = &snapshot.NullString{Valid: true, Value: relation.ToastTable}
```
seanlinsley, line 82: Why do we use `NullString` in the first place? Is it somehow a space saving versus sending an empty string?

state/postgres_relation_stats.go
seanlinsley: Update I changed the `coalesce` call to default to -1 so now internally, the field will be null for old collector versions, and -1 for tables missing TOAST (or where the row count is unknown).

input/postgres/relations.go
```diff
@@ -32,10 +32,12 @@ const relationsSQL string = `
 				COALESCE((SELECT p.partstrat FROM pg_partitioned_table p WHERE p.partrelid = c.oid), '') AS partition_strategy,
 				(SELECT p.partattrs FROM pg_partitioned_table p WHERE p.partrelid = c.oid) AS partition_columns,
 				COALESCE(pg_catalog.pg_get_partkeydef(c.oid), '') AS partition_expr,
-				locked_relids.relid IS NOT NULL AS exclusively_locked
+				locked_relids.relid IS NOT NULL AS exclusively_locked,
+				COALESCE(toast.relname, '') AS toast_table
 	 FROM pg_catalog.pg_class c
 	 LEFT JOIN pg_catalog.pg_namespace n ON (n.oid = c.relnamespace)
 	 LEFT JOIN locked_relids ON (c.oid = locked_relids.relid)
+	 LEFT JOIN pg_class toast ON (c.reltoastrelid = toast.oid AND toast.relkind = 't')
```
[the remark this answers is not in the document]
lfittl: Though I guess that would already be a hazard given that we run `pg_total_relation_size` on the toast table already. So maybe we can assume that to not be a problem in practice.

lfittl (approved):
:+1:

PR #417, opened 2023-05-13
RPM packages: Ensure collector gets started after reboot

lfittl:
This was an oversight in the packaging logic, since we utilize fpm's "--deb-systemd-auto-start" setting for the Debian-based packages, but we don't have the equivalent for the RPM packages.

Fix by adding an explicit "systemctl enable" on installation, and a corresponding "systemctl disable" on removal, matching what was already done for the Debian packages.


keiko713 (approved):

lfittl:
> Hmm, I'm not seeing that code for Debian. Or is it done automatically?

Its done automatically by the `--deb-systemd-auto-start` option to fpm, as noted in the PR description :)

PR #418, opened 2023-05-15
Reload collector config after successful test run

seanlinsley:
It's fairly common for users to make a config change and verify it with `--test`, but forget to include `--reload` so the background collector continues running with the old configuration. ~~This PR attempts to reduce our support burden by automatically reloading the configuration every hour.~~

Another option might be to always reload the configuration after a successful test run, removing the `--reload` flag.

### DIFF
--- a/main.go
+++ b/main.go
@@ -307,7 +307,7 @@ func main() {
 
 	// Automatically reload the configuration after a successful test run.
 	if testRun && !noReload {
-		reloadRun = true;
+		reloadRun = true
 	}
 
 	if showVersion {
--- a/main.go
+++ b/main.go
@@ -267,7 +267,7 @@ func main() {
 	logger := &util.Logger{}
 
 	flag.BoolVarP(&showVersion, "version", "", false, "Shows current version of the collector and exits")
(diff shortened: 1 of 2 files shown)


lfittl:
> Another option might be to always reload the configuration after a successful test run, removing the `--reload` flag.

I think that would be better and more predictable, i.e. I'd prefer if we made this a function of the test run, and not something that runs on a schedule.

Maybe we can introduce a new `--no-reload` flag in case someone doesn't want to perform a reload after a test?

seanlinsley:
Okay @lfittl I've updated the PR to automatically reload after a successful test run, and added `--no-reload` to disable it.

I haven't tested manually, but reading the code it seems like it should work.

msakrejda (approved):
The change looks good, but we should probably mention the new behavior of the `--test` flag in the `--test` flag doc string.

We should eventually also update our docs, but since people will be using old collectors with the latest docs for a while, there's no rush there. I wonder if anywhere in our docs we rely on the current behavior when `--reload` is not passed, but I can't think of why we might be doing that, so it's probably not an issue.

seanlinsley:
Okay, I added it to the `--test` doc string. It was getting overly long so I reworded it (feel free to recommend alternatives)

msakrejda:
+1 on the rewording; good call.

PR #419, opened 2023-05-26
Update license_finder information, remove "lpx" module use for Heroku

lfittl:


input/system/heroku/http_handler.go
```diff
@@ -21,12 +22,83 @@ func SetupHttpHandlerLogs(ctx context.Context, wg *sync.WaitGroup, globalCollect
 	go func() {
 		http.HandleFunc("/", util.HttpRedirectToApp)
 		http.HandleFunc("/logs/", func(w http.ResponseWriter, r *http.Request) {
-			lp := lpx.NewReader(bufio.NewReader(r.Body))
-			for lp.Next() {
-				procID := string(lp.Header().Procid)
-				if procID == "heroku-postgres" || strings.HasPrefix(procID, "postgres.") {
+			reader := bufio.NewReader(r.Body)
```
keiko713, line 25: So... I'm personally curious. Where does this logging format coming from?

keiko713 (approved):
One question/comment but overall looking good 👍

input/system/heroku/http_handler.go
lfittl: Heroku Logplex :)

keiko713: Ahh, I was missing the first part of getting `lenStr` and assuming that `83 <40>1 2012-11-30T06:45:29+00:00 host app web.3 - State changed from starting to up` is not in the same format (also I wasn't able to find MSGID but I guess that's `-`).

https://devcenter.heroku.com/articles/log-drains#https-drains

Also it's almost the same as https://datatracker.ietf.org/doc/html/rfc5424 (as it's mentioned in above doc)

msakrejda (approved):
Looks like CI is failing on lint; otherwise, looks good.

PR #421, opened 2023-06-01
Look for truncation messages in the middle of log lines

msakrejda:
This can happen because of stitching; we can't just look for them at
the end.


keiko713 (approved):
🌳

PR #425, opened 2023-06-06
Collector test: Always return correct exit code, even with --no-reload

lfittl:
Due to a coding oversight that dates a while back, the exit code is only set when doing test+reload, but not when doing just test.

Whilst a recent change has flipped the default of test to also execute reload automatically, it doesn't hurt to return a failue exit code (1) in case of failure when "--no-reload" is specified.

### DIFF
--- a/main.go
+++ b/main.go
@@ -509,6 +509,8 @@ ReadConfigAndRun:
 						logger.PrintError("Error: Reload requested, but ignoring since configuration errors are present")
 						exitCode = 1
 					}
+				} else if !success {
+					exitCode = 1
 				}
 				break DoneOrSignal
 			case s := <-sigs:


main.go
```diff
@@ -509,6 +509,8 @@ ReadConfigAndRun:
 						logger.PrintError("Error: Reload requested, but ignoring since configuration errors are present")
 						exitCode = 1
 					}
+				} else if !success {
+					exitCode = 1
```
keiko713, line 513: would it be helpful to have some log here too?

keiko713 (approved):

main.go
lfittl: I don't think that's necessary, since the test command already outputs the failure information - in the case of a reload we added this because its helpful to know that the reload was skipped.

PR #427, opened 2023-06-10
autovacuum logs: Match index vacuum information in Postgres 14+

lfittl:
Whilst we have supported other syntax that was introduced in Postgres 14, we missed the fact that new index vacuum details were added for each index of a table. This caused events to not be detected correctly.

See https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=5aed6a1fc214913de9ac69c1717dc64a2483e16d

### DIFF
--- a/logs/analyze_test.go
+++ b/logs/analyze_test.go
@@ -1443,6 +1443,75 @@ var tests = []testpair{
 	},
 	{
 		[]state.LogLine{{
+			Content: "automatic aggressive vacuum to prevent wraparound of table \"mydb.myschema.mytable\": index scans: 1\n" +
+				"	pages: 8141555 removed, 4783594 remain, 0 skipped due to pins, 10478504 skipped frozen\n" +
+				"	tuples: 2153407 removed, 44326634 remain, 252 are dead but not yet removable, oldest xmin: 1887899920\n" +
+				"	index scan needed: 808328 pages from table (6.25% of total) had 2317145 dead item identifiers removed\n" +
+				"	index \"mytable_index1\": pages: 235675 in total, 5967 newly deleted, 174124 currently deleted, 168157 reusable\n" +
+				"	index \"mytable_index2\": pages: 179 in total, 0 newly deleted, 0 currently deleted, 0 reusable\n" +
+				"	I/O timings: read: 377482.257 ms, write: 8483.639 ms\n" +
+				"	avg read rate: 104.788 MB/s, avg write rate: 19.082 MB/s\n" +
+				"	buffer usage: 2692149 hits, 11619539 misses, 2115983 dirtied\n" +
+				"	WAL usage: 4628822 records, 1983515 full page images, 4131405282 bytes\n" +
+				"	system usage: CPU: user: 189.32 s, system: 84.28 s, elapsed: 866.30 s",
+			LogLevel: pganalyze_collector.LogLineInformation_LOG,
+		}},
+		[]state.LogLine{{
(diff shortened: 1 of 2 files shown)


logs/analyze_test.go
```diff
@@ -1441,6 +1441,75 @@ var tests = []testpair{
...
+				"vacuum_page_hit":          2692149,
+				"vacuum_page_miss":         11619539,
+				"vacuum_page_dirty":        2115983,
+				"wal_records":              4628822,
+				"wal_fpi":                  1983515,
+				"wal_bytes":                4131405282,
+				"rusage_user":              189.32,
+				"rusage_kernel":            84.28,
+				"elapsed_secs":             866.3,
+				"index_vacuums": map[string]interface{}{
```
lfittl, line 1494: Note for reviewer:

This is the first case where we are adding nested information to the details of a log line. I think this structure makes sense here, because it would allow us to query all log events for a table, and then to a lookup on the key to find a specific index and its vacuum behavior over time.

lfittl: Thinking through this more, an alternate approach could be to extract vacuum information in a way similar to query samples - i.e. carry it separately in the log snapshot.

The benefit there is that its slightly more compact from a space perspective, and we already process this as structured data on the server side anyway nowadays. Interested to hear what others think makes sense.

keiko713: Overall, I think this change makes sense to me too. How are you thinking to store this data on the server side? Different table from what we have now, or additional jsonb column inside of the existing table? Though, the latter, I'm not sure if "do a lookup on the key to find a specific index and its vacuum behavior over time" is straight forward (the latter is simpler to store though).

Unrelated to that, but another question re: this PR. I believe that our db is using 14+ but I've personally never seen this log line 🤔 Will this show up only under certain circumstances?

lfittl: > Overall, I think this change makes sense to me too. How are you thinking to store this data on the server side? Different table from what we have now, or additional jsonb column inside of the existing table? Though, the latter, I'm not sure if "do a lookup on the key to find a specific index and its vacuum behavior over time" is straight forward (the latter is simpler to store though).

Its probably best stored in a separate table once Postgres 14 is common enough that this can be used reliably. I think this information should be very useful for debugging index bloat problems specifically.

> Unrelated to that, but another question re: this PR. I believe that our db is using 14+ but I've personally never seen this log line 🤔 Will this show up only under certain circumstances?

Currently these get classified as Z00 (unknown log event) since the regexp doesn't match at all. The regexp continues to match for autovacuums that did not have any index scan phases.

I found this when trying to understand why a particular VACUUM we registered as a vacuum run (i.e. via pg_stat_activity / pg_stat_progress_vacuum) didn't show up at all in the log events.

keiko713: > Currently these get classified as Z00 (unknown log event) since the regexp doesn't match at all. The regexp continued to match for autovacuums that did not have any index scan phases, which is why it was hard to spot, since some log events were coming in.

Ohhh nice, I found one example too. This is also nice -- this means that we were missing to store some autovacuums into autovacuum_run_stats too, and this will fix it.

> Its probably best stored in a separate table once Postgres 14 is common enough that this can be used reliably.

If we don't plan to store separately, I was going to vote for keeping it as is. If we do, I think both still have pros and cons so honestly not strong opinion to this. I'll approve as is but happy to look into again if you decide to go the other way.

keiko713 (approved):

PR #438, opened 2023-07-28
Debian: Bump minimum required version to Bullseye (11), add Bookworm (12)

### LINKED ISSUE #437: Debian build uses EOL version
The Debian package build is running `buster`, which [no longer receives security updates, as of June 30th, 2022](https://www.debian.org/releases/buster/): https://github.com/pganalyze/collector/blob/c1233a3b9fdb861ec0e1498d351ada0a4c19553c/packages/src/Dockerfile.build.deb-systemd#L1

lfittl:
This also stops building packages for Buster (10), which has reached EOL.

Fixes #437

---

Separately this also drops Ubuntu 16.04 and 18.04 which are both EOL as well, and the Bullseye glibc is too new for them to be supported.

### DIFF
--- a/.github/workflows/test-packages.yml
+++ b/.github/workflows/test-packages.yml
@@ -10,7 +10,7 @@ jobs:
   build:
     # Use older Ubuntu release to ensure availability of CGroupsv1 (which are necessary
     # currently to support older systemd on some of the test containers)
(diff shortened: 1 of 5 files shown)


keiko713 (approved):
Looks good 👍  The test was failing somehow so I re-run and it went well.

Issue #435, opened 2023-07-26
Additional security settings for pganalyze collector


jawnsy:
We've been using these settings successfully for quite some time, so I thought I'd contribute these upstream if it's of interest to you.

Container volumes/security context:

```yaml
          volumeMounts:
            - mountPath: /tmp
              name: scratch
              subPath: tmp
          securityContext:
            allowPrivilegeEscalation: false
            capabilities:
              drop:
                - ALL
            readOnlyRootFilesystem: true
            runAsGroup: 1000
            runAsNonRoot: true
            runAsUser: 1000
            seccompProfile:
              type: RuntimeDefault
```

Pod security context:

```yaml
      securityContext:
        runAsGroup: 65532
        runAsNonRoot: true
        runAsUser: 65532
        seccompProfile:
          type: RuntimeDefault
      volumes:
        - name: scratch
          emptyDir: {}
      enableServiceLinks: false
```

lfittl:
@jawnsy Thanks - those seem useful! Would your idea be to add them to Helm chart?

jawnsy:
Yup, exactly. Happy to open a pull request!

lfittl:
@jawnsy Sounds good!

PR #439, opened 2023-07-28
Improve default security settings

jawnsy:
* Add default seccomp profile
* Add pod security context
* Enable read-only root filesystem by default
* Disable default service links

Closes: #435


jawnsy:
Tested this with the command:

```
helm upgrade pganalyze ./contrib/helm/pganalyze-collector/ --install --set=resources.requests.cpu=0 --set=resources.requests.memory=0
```

Here's the resulting pod (with a few fields removed):

```yaml
$ k get pod pganalyze-pganalyze-collector-77bcd6b6b6-vgp4z -o yaml
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2023-07-28T03:44:56Z"
  generateName: pganalyze-pganalyze-collector-77bcd6b6b6-
  labels:
    app.kubernetes.io/instance: pganalyze
    app.kubernetes.io/name: pganalyze-collector
    pod-template-hash: 77bcd6b6b6
  name: pganalyze-pganalyze-collector-77bcd6b6b6-vgp4z
  namespace: jawnsy-test
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: ReplicaSet
    name: pganalyze-pganalyze-collector-77bcd6b6b6
    uid: 9f8df1df-a047-48ce-ad07-b9e810af6faa
  resourceVersion: "70758548"
  uid: 8f834c68-6cdb-41f9-b1e3-00776df3ba46
spec:
  containers:
  - image: quay.io/pganalyze/collector:v0.50.1
    imagePullPolicy: IfNotPresent
    name: pganalyze-collector
    resources:
      limits:
        cpu: "1"
        memory: 1Gi
      requests:
        cpu: "0"
        memory: "0"
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      readOnlyRootFilesystem: true
      runAsGroup: 1000
      runAsNonRoot: true
      runAsUser: 1000
      seccompProfile:
        type: RuntimeDefault
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    volumeMounts:
    - mountPath: /tmp
      name: scratch
      subPath: tmp
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: kube-api-access-w5ckd
      readOnly: true
  dnsPolicy: ClusterFirst
  enableServiceLinks: false
  nodeName: gke-internal-tools-d-application-13b4-a1815e79-tzn5
  preemptionPolicy: PreemptLowerPriority
  priority: 0
  restartPolicy: Always
  schedulerName: default-scheduler
  securityContext:
    runAsGroup: 1000
    runAsNonRoot: true
    runAsUser: 1000
    seccompProfile:
      type: RuntimeDefault
  serviceAccount: pganalyze-pganalyze-collector
  serviceAccountName: pganalyze-pganalyze-collector
  terminationGracePeriodSeconds: 30
  tolerations:
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300
  volumes:
  - emptyDir: {}
    name: scratch
  - name: kube-api-access-w5ckd
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          expirationSeconds: 3607
          path: token
      - configMap:
          items:
          - key: ca.crt
            path: ca.crt
          name: kube-root-ca.crt
      - downwardAPI:
          items:
          - fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
            path: namespace
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2023-07-28T03:44:56Z"
    status: "True"
    type: Initialized
  - lastProbeTime: null
    lastTransitionTime: "2023-07-28T03:44:57Z"
    status: "True"
    type: Ready
  - lastProbeTime: null
    lastTransitionTime: "2023-07-28T03:44:57Z"
    status: "True"
    type: ContainersReady
  - lastProbeTime: null
    lastTransitionTime: "2023-07-28T03:44:56Z"
    status: "True"
    type: PodScheduled
  containerStatuses:
  - containerID: containerd://10106bd5210591d1e464b6a9d316195e5719dcb0b8ded8e4a817ef93f7f5dcd9
    image: quay.io/pganalyze/collector:v0.50.1
    imageID: quay.io/pganalyze/collector@sha256:bdf21b3a1e9252fc9e6b8bda20a7089db77f79e677811e9592c1c8abaaf3a151
    lastState: {}
    name: pganalyze-collector
    ready: true
    restartCount: 0
    started: true
    state:
      running:
        startedAt: "2023-07-28T03:44:57Z"
  phase: Running
  qosClass: Burstable
  startTime: "2023-07-28T03:44:56Z"
```

contrib/helm/pganalyze-collector/templates/deployment.yaml
```diff
@@ -26,6 +26,7 @@ spec:
       labels:
         {{- include "pganalyze-collector.selectorLabels" . | nindent 8 }}
     spec:
+      enableServiceLinks: false
```
jawnsy, line 29: This adds a lot of service discovery variables, which are useless for most applications:

```
EnableServiceLinks indicates whether information about services should be injected into pod's environment variables, matching the syntax of Docker links. Optional: Defaults to true.
```

From https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#pod-v1-core

lfittl: I wonder if there would be a use case for this, if you had your Postgres running inside Kubernetes as well? (though even then I imagine the collector would actually have to read out the relevant env variables?)

contrib/helm/pganalyze-collector/values.yaml
```diff
@@ -74,10 +83,15 @@ resources:
...
+volumes:
+  - name: scratch
+    emptyDir: {}
 
 # -- List of volume mounts to attach to the container
-volumeMounts: []
+volumeMounts:
+  - mountPath: /tmp
+    name: scratch
+    subPath: tmp
```
lfittl, line 94: Given the read-only root filesystem setting, I wonder if we also need to create a volume mount for `/state`? (which is whether the state file is written to - whilst that is mainly intended to support restarts outside of a container-based environment, it avoids errors that would likely occur when `/state` is no writeable)

See https://github.com/pganalyze/collector/blob/main/contrib/docker-entrypoint.sh#L23

contrib/helm/pganalyze-collector/templates/deployment.yaml
jawnsy: I think even for users trying to monitor PostgreSQL running in Kubernetes, it would be more likely that someone use a service DNS name instead of the IP address. We can move this to a default setting in values.yaml if we want to make it configurable?

contrib/helm/pganalyze-collector/values.yaml
jawnsy: Ah, yes, absolutely! I didn't know about this file

contrib/helm/pganalyze-collector/templates/deployment.yaml
lfittl: I think you're probably right, but maybe making it configurable doesn't hurt, just to avoid something having to fork the chart to customize this.

lfittl: Thinking through this again, given that this is mainly about the injection of environment variables, there isn't actually much benefit to allowing this to be configurable - since the collector doesn't use these variables. Thus I think we can actually leave this as `enableServiceLinks = false` in all cases (and revise only if we hear feedback that this needs to be configurable).

lfittl (approved):

lfittl:
Merged with small edits - thanks for the contribution!

jawnsy:
@lfittl Thanks for finishing this up and for merging! Apologies for the delay, I was out of town last week :)

PR #440, opened 2023-07-29
Use log_timezone setting to determine log timezone if possible

msakrejda:
Right now, we infer the log timezone by parsing this out of log lines.
This works in theory, but in practice Go's time.LoadLocation does not
support many timezone abbreviations (and abbreviations can be
ambiguous anyway).

Since we already collect Postgres settings, and these specify the log
timezone in a non-abbreviated form, we can use these. These correspond
to the full name in the IANA Time Zone database, as time.LoadLocation
expects. Sync this on every full snapshot just in case the timezone is
updated.

The current implementation stores this state on the Server, protected
by a new mutex. Since this needs to be written by the full snapshot
and read during log parsing, none of the existing mutexes or state
seemed appropriate. I admit this is not a great solution and I'm open
to other ideas.

### DIFF
--- a/input/system/azure/logs.go
+++ b/input/system/azure/logs.go
@@ -251,7 +251,7 @@ func ParseRecordToLogLines(in AzurePostgresLogRecord) ([]state.LogLine, string, 
 		azureDbServerName = in.LogicalServerName
 	}
 
-	logLine, ok := logs.ParseLogLineWithPrefix("", logLineContent)
+	logLine, ok := logs.ParseLogLineWithPrefix("", logLineContent, nil)
 	if !ok {
 		return []state.LogLine{}, "", fmt.Errorf("Can't parse log line: \"%s\"", logLineContent)
 	}
--- a/logs/parse.go
+++ b/logs/parse.go
@@ -113,7 +113,7 @@ func IsSupportedPrefix(prefix string) bool {
 	return false
 }
 
-func ParseLogLineWithPrefix(prefix string, line string) (logLine state.LogLine, ok bool) {
+func ParseLogLineWithPrefix(prefix string, line string, tz *time.Location) (logLine state.LogLine, ok bool) {
 	var timePart, userPart, dbPart, appPart, pidPart, logLineNumberPart, levelPart, contentPart string
 
 	// Assume Postgres time format unless overriden by the prefix (e.g. syslog)
@@ -474,42 +474,12 @@ func ParseLogLineWithPrefix(prefix string, line string) (logLine state.LogLine, 
 		}
 	}
 
-	var err error
 	if timePart != "" {
-		logLine.OccurredAt, err = time.Parse(timeFormat, timePart)
-		if err != nil {
-			if timeFormatAlt != "" {
-				// Ensure we have the correct format remembered for ParseInLocation call that may happen later
-				timeFormat = timeFormatAlt
-				logLine.OccurredAt, err = time.Parse(timeFormat, timePart)
-			}
-			if err != nil {
-				return
-			}
+		occurredAt := getOccurredAt(timePart, timeFormat, timeFormatAlt, tz)
+		if occurredAt.IsZero() {
+			return
 		}
-		// Handle non-UTC timezones in systems that have log_timezone set to a different
-		// timezone value than their system timezone. This is necessary because Go otherwise
-		// only reads the timezone name but does not set the timezone offset, see
-		// https://pkg.go.dev/time#Parse
-		zone, offset := logLine.OccurredAt.Zone()
-		if offset == 0 && zone != "UTC" && zone != "" {
-			var zoneLocation *time.Location
-			zoneNum, err := strconv.Atoi(zone)
-			if err == nil {
-				zoneLocation = time.FixedZone(zone, zoneNum*3600)
-			} else {
-				zoneLocation, err = time.LoadLocation(zone)
-				if err != nil {
-					// We don't know which timezone this is (and a timezone name is present), so we can't process this log line
-					return
-				}
-			}
-			logLine.OccurredAt, err = time.ParseInLocation(timeFormat, timePart, zoneLocation)
-			if err != nil {
-				// Technically this should not occur (as we should have already failed previously in time.Parse)
-				return
-			}
-		}
+		logLine.OccurredAt = occurredAt
 	}
 
 	if userPart != "[unknown]" {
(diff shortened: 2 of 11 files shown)


input/system/azure/logs.go
```diff
@@ -251,7 +251,7 @@ func ParseRecordToLogLines(in AzurePostgresLogRecord) ([]state.LogLine, string,
 		azureDbServerName = in.LogicalServerName
 	}
 
-	logLine, ok := logs.ParseLogLineWithPrefix("", logLineContent)
+	logLine, ok := logs.ParseLogLineWithPrefix("", logLineContent, nil)
```
msakrejda, line 254: Systems where we collect logs for multiple servers through the same stream are harder to fix (but on the plus side, far less likely to have this problem).

logs/parse.go
```diff
@@ -541,13 +511,69 @@ func ParseLogLineWithPrefix(prefix string, line string) (logLine state.LogLine,
...
+			return result
+		}
+		if timeFormatAlt != "" {
+			// Ensure we have the correct format remembered for ParseInLocation call that may happen later
+			timeFormat = timeFormatAlt
+			result, err := time.ParseInLocation(timeFormat, timePart, tz)
+			if err == nil {
+				return result
+			}
+		}
```
msakrejda, line 527: This mirroring the logic below is pretty awkward. I don't know if there's a good way to just ignore the time zone in this case.

input/system/azure/logs.go
msakrejda: Looking at this again, this is not impossible: we'd need to pass in a mapping of server identifier (as it occurs in the log stream) to *time.Location. I'll hold off on addressing this until we decide if this is a reasonable and worthwhile approach.

logs/parse.go
```diff
@@ -541,13 +511,69 @@ func ParseLogLineWithPrefix(prefix string, line string) (logLine state.LogLine,
...
 
+func getOccurredAt(timePart, timeFormat, timeFormatAlt string, tz *time.Location) time.Time {
+	if tz != nil {
+		result, err := time.ParseInLocation(timeFormat, timePart, tz)
+		if err == nil {
+			return result
+		}
+		if timeFormatAlt != "" {
+			// Ensure we have the correct format remembered for ParseInLocation call that may happen later
+			timeFormat = timeFormatAlt
```
keiko713, line 522: If we fail here and go to the old logic, `timeFormat` is updated as `timeFormatAlt` on line 531 `ts, err := time.Parse(timeFormat, timePart)` part already. I think we could pass `timeFormatAlt` below instead.

logs/parse.go
keiko713: When location is passed with `time.ParseInLocation`, it'll be unclear how the timezone info within the original `timeFormat` part will be treated.

> ParseInLocation is like Parse but differs in two important ways. First, in the absence of time zone information, Parse interprets a time as UTC; ParseInLocation interprets the time as in the given location. Second, when given a zone offset or abbreviation, Parse tries to match it against the Local location; ParseInLocation uses the given location.

In short:
* not timezone info => interprets the time as in the given location
* no zone offset or abbreviation => uses the given location

With this, I tested with several cases to see how it behaves, especially when the offset/abbreviation is different from the given location's one.
Summarizing, it'll try to respect the original `timeFormat`' s offset/abbreviation over the given location, then it's not really going well (esp. abbr case as the offset will be 0, which is the most important part to calculate unix time). Honestly, I don't even know if such case exists, but I figured that I'd raise this as concern.
With this, when we have a location, I'd say that we can actually try to drop offset/abbreviation like you mentioned from `timeFormat` then use the unique formatting like `const shortForm = "2006-01-02 15:04:05"`, then completely rely on a location info to set timezone.
One thing I noticed: we already kinda do this pattern with rsyslog, as it's using `2006 Jan  2 15:04:05` as a format. With this new change, previously above was always assumed as UTC, but potentially it turns into some other timezone? I might just be thinking some impossible, so feel free to let me know "no, Keiko, it won't happen".

-----

### Case 1-1: with offset (but not matching to the location)

```go
const longForm = "2006-01-02 15:04:05 -0700"
tz, _ := time.LoadLocation("Europe/London")
time.ParseInLocation(longForm, "2023-07-01 00:00:00 -0500", tz)
fmt.Println(t)
// returns 2023-07-01 00:00:00 -0500 -0500
// Unix time: 1688187600
// zone: "", offset: -18000 (-5 * 3600)
```

### Case 1-2: with offset (matching to the location)

```go
const longForm = "2006-01-02 15:04:05 -0700"
tz, _ := time.LoadLocation("Asia/Tokyo")
t, _ := time.ParseInLocation(longForm, "2023-07-01 00:00:00 +0900", tz)
fmt.Println(t)
// returns 2023-07-01 00:00:00 +0900 JST
// Unix time: 1688137200
// zone: JST, offset: 32400 (9 * 3600)
```

### Case 2-1: with abbreviation (but not matching to the location)

```go
const longForm = "2006-01-02 15:04:05 MST"
tz, _ := time.LoadLocation("Europe/London")
t, _ := time.ParseInLocation(longForm, "2023-07-01 00:00:00 JST", tz)
fmt.Println(t)
// returns 2023-07-01 00:00:00 +0000 JST
// Unix time: 1688169600
// zone: JST, offset: 0
```

### Case 2-2: with abbreviation (matching to the location)

```go
const longForm = "2006-01-02 15:04:05 MST"
tz, _ := time.LoadLocation("Europe/London")
t, _ := time.ParseInLocation(longForm, "2023-07-01 00:00:00 BST", tz)
fmt.Println(t)
// returns 2023-07-01 00:00:00 +0100 BST
// Unix time: 1688166000
// zone: BST, offset: 3600
```

msakrejda: Thanks for the feedback and analysis. I think this suggests we should strip off the timezone before processing the timestamps when we have a location from `log_timezone`. I'll update the code to do that.

logs/parse.go
msakrejda: Good catch. This should be taken care of once I address the changes mentioned above.

msakrejda:
@keiko713 I refactored this to try to simplify timestamp handling. I was wary of touching this complicated code too much earlier, but I think this works much better. Let me know what you think.

@lfittl this is probably also ready for your feedback now, especially around if you think we should expand this to support logs ingested through provider APIs where it's harder (but feasible) to match them to server `log_timezone`s.

logs/parse.go
```diff
@@ -113,13 +113,9 @@ func IsSupportedPrefix(prefix string) bool {
 	return false
 }
 
-func ParseLogLineWithPrefix(prefix string, line string) (logLine state.LogLine, ok bool) {
+func ParseLogLineWithPrefix(prefix string, line string, tz *time.Location) (logLine state.LogLine, ok bool) {
 	var timePart, userPart, dbPart, appPart, pidPart, logLineNumberPart, levelPart, contentPart string
 
-	// Assume Postgres time format unless overriden by the prefix (e.g. syslog)
-	timeFormat := "2006-01-02 15:04:05 -0700"
-	timeFormatAlt := "2006-01-02 15:04:05 MST"
```
msakrejda, line 121: The general idea is these are no longer locals of this big function but pushed into the `getOccurredAt` helper.

logs/parse.go
```diff
@@ -541,13 +506,77 @@ func ParseLogLineWithPrefix(prefix string, line string) (logLine state.LogLine,
...
+func getOccurredAt(timePart string, tz *time.Location, rsyslog bool) time.Time {
+	if tz != nil && !rsyslog {
+		lastSpaceIdx := strings.LastIndex(timePart, " ")
+		if lastSpaceIdx == -1 {
+			return time.Time{}
+		}
+		timePartNoTz := timePart[0:lastSpaceIdx]
+		result, err := time.ParseInLocation("2006-01-02 15:04:05", timePartNoTz, tz)
+		if err != nil {
+			return time.Time{}
```
msakrejda, line 518: I originally fell back to the existing parsing if this fails, but I feel like failure here is unlikely and we can just consider the line unparsable, like we were doing for other time parsing errors before.

logs/parse.go
keiko713: Yeah I was going to suggest this earlier too, I think it's good that you moved this to `getOccurredAt` 👍

keiko713 (approved):
Looking good!

msakrejda:
Per discussion with @lfittl, it would be good to support log_timezone for hosted providers. I'll do that in a follow-up PR.

PR #443, opened 2023-08-05
Log Parsing: Fix crash in --analyze-logfile option

lfittl:
Due to the changes in 7b67039fd9a22e5b5e34cb934b73a18f42115ea7 we need to always initialize the server mutex for the log parsing functions.


msakrejda (approved):
Good catch. Maybe we should add a --analyze-logfile test to our full suite, but no need to do that now.

PR #444, opened 2023-08-05
Autovacuum log parsing: Add support for updated format in Postgres 15+

lfittl:
This adds support for the changes done in the following upstream commits:

* Simplify lazy_scan_heap's handling of scanned pages https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=44fa84881fff4529d68e2437a58ad2c906af5805

* Add VACUUM instrumentation for scanned pages, relfrozenxid. https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=872770fd6ccf12596b9e26234df9a37cae083af2

### DIFF
--- a/logs/analyze.go
+++ b/logs/analyze.go
@@ -79,9 +79,13 @@ var autoVacuum = analyzeGroup{
 		prefixes: []string{"automatic vacuum of table", "automatic aggressive vacuum of table", "automatic aggressive vacuum to prevent wraparound of table"},
 		regexp: regexp.MustCompile(`^automatic (aggressive )?vacuum (to prevent wraparound )?of table "(.+?)": index scans: (\d+),?\s*` +
 			`(?:elapsed time: \d+ \w+, index vacuum time: \d+ \w+,)?\s*` + // Google AlloyDB for PostgreSQL
-			`pages: (\d+) removed, (\d+) remain, (\d+) skipped due to pins, (\d+) skipped frozen,?\s*` +
+			`pages: (\d+) removed, (\d+) remain, (?:(\d+) skipped due to pins, (\d+) skipped frozen|(\d+) scanned \(([\d.]+)% of total\)),?\s*` +
 			`(?:\d+ skipped using mintxid)?,?\s*` + // Google AlloyDB for PostgreSQL
 			`tuples: (\d+) removed, (\d+) remain, (\d+) are dead but not yet removable(?:, oldest xmin: (\d+))?,?\s*` +
+			`(?:tuples missed: (\d+) dead from (\d+) pages not removed due to cleanup lock contention)?,?\s*` + // Postgres 15+
+			`(?:removable cutoff: (\d+), which was (\d+) XIDs old when operation ended)?,?\s*` + // Postgres 15+
+			`(?:new relfrozenxid: (\d+), which is (\d+) XIDs ahead of previous value)?,?\s*` + // Postgres 15+
+			`(?:new relminmxid: (\d+), which is (\d+) MXIDs ahead of previous value)?,?\s*` + // Postgres 15+
 			`(?:index scan (not needed|needed|bypassed|bypassed by failsafe): (\d+) pages from table \(([\d.]+)% of total\) (?:have|had) (\d+) dead item identifiers(?: removed)?)?,?\s*` + // Postgres 14+
 			`((?:index ".+?": pages: \d+ in total, \d+ newly deleted, \d+ currently deleted, \d+ reusable,?\s*)*)?` + // Postgres 14+
 			`(?:I/O timings: read: ([\d.]+) ms, write: ([\d.]+) ms)?,?\s*` + // Postgres 14+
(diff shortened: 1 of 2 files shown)


logs/analyze.go
```diff
@@ -1680,35 +1689,68 @@ func classifyAndSetDetails(logLine state.LogLine, statementLine state.LogLine, d
 				"vacuum_page_dirty": vacuumPageDirty, "read_rate_mb": readRateMb,
 				"write_rate_mb": writeRateMb, "rusage_kernel": rusageKernelMode,
 				"rusage_user": rusageUserMode, "elapsed_secs": rusageElapsed,
-				"pinskipped_pages": pinskippedPages, "frozenskipped_pages": frozenskippedPages,
 			}
 			// List anti-wraparound status either if the message indicates that it is, or if
 			// our Postgres version is new enough (13+) as determined by the presence of WAL
-			// record information (parts[26])
+			// record information (parts[37])
```
lfittl, line 1695: Note for reviewer: In general most part numbers after the newly added lines have incremented by 10 - this here got incremented by 11 because there was a miss on a previous change that affected part numbers but forgot to adjust the comment.

logs/analyze.go
```diff
@@ -1634,38 +1638,43 @@ func classifyAndSetDetails(logLine state.LogLine, statementLine state.LogLine, d
...
-			newDeadTuples, _ := strconv.ParseInt(parts[11], 10, 64)
 
-			// Parts 12 to 19 (Index scan info, I/O read/write timings) are only present on Postgres 14+ and dealt with at the end
+			// Parts 7 to 10 (scanned, pinskipped and frozenskipped pages) are version dependent and dealt with later
 
-			if parts[20] != "" { // Postgres 14+, with I/O information before buffers
-				readRatePart = parts[20]
-				writeRatePart = parts[21]
+			tuplesDeleted, _ := strconv.ParseInt(parts[11], 10, 64)
+			newRelTuples, _ := strconv.ParseInt(parts[12], 10, 64)
```
msakrejda, line 1645: This is following an existing pattern, so we shouldn't change it now, but I wonder if we shouldn't add a helper to panic if there's a parsing error instead of ignoring it. There should not be errors here, but given how many different matches we are juggling, I wouldn't be surprised if we get something wrong here sooner or later. It might be nice to complain loudly rather than proceed with bogus data.

logs/analyze.go
```diff
@@ -1680,35 +1689,68 @@ func classifyAndSetDetails(logLine state.LogLine, statementLine state.LogLine, d
 				"vacuum_page_dirty": vacuumPageDirty, "read_rate_mb": readRateMb,
 				"write_rate_mb": writeRateMb, "rusage_kernel": rusageKernelMode,
 				"rusage_user": rusageUserMode, "elapsed_secs": rusageElapsed,
-				"pinskipped_pages": pinskippedPages, "frozenskipped_pages": frozenskippedPages,
```
msakrejda, line 1683: ?

msakrejda: Ah, this is because this is no longer always present, because of the change in 15?

logs/analyze_test.go
```diff
@@ -1719,6 +1719,100 @@ var tests = []testpair{
...
+				"	index scan needed: 11169 pages from table (3.71% of total) had 160440 dead item identifiers removed\n" +
+				"	index \"index1\": pages: 46608 in total, 0 newly deleted, 0 currently deleted, 0 reusable\n" +
+				"	index \"index2\": pages: 9365 in total, 2 newly deleted, 2 currently deleted, 0 reusable\n" +
+				"	index \"index3\": pages: 38252 in total, 129 newly deleted, 14414 currently deleted, 14285 reusable\n" +
+				"	index \"index4\": pages: 59530 in total, 0 newly deleted, 21608 currently deleted, 21608 reusable\n" +
+				"	I/O timings: read: 1661.751 ms, write: 283.343 ms\n" +
+				"	avg read rate: 76.468 MB/s, avg write rate: 28.413 MB/s\n" +
+				"	buffer usage: 79150 hits, 206012 misses, 76546 dirtied\n" +
+				"	WAL usage: 93766 records, 71348 full page images, 169121919 bytes\n" +
+				"	system usage: CPU: user: 4.49 s, system: 1.30 s, elapsed: 21.04 s\n",
```
msakrejda, line 1740: It seems like this might be easier to read (and write?) with \`\` quotes, but NBD.

msakrejda (approved):

logs/analyze.go
lfittl: Yeah, that's right - this was removed in Postgres 15 and replaced with the number of pages that were scanned in the table.

logs/analyze.go
lfittl: Yeah, agreed - probably not panic, but output a warning or similar.

logs/analyze_test.go
lfittl: Good point, changed the new test but left the other ones as-is to keep the PR simple.

PR #446, opened 2023-08-10
AWS Aurora: Correctly detect Aurora reader instances as replicas

lfittl:
Previously we skipped processing of replication statistics for Aurora, since the regular built-in Postgres replication functions error out in many cases. However, this also meant that all Aurora instances were always detected as primaries, regardless of whether they were a writer or a reader, which led to confusing in-app experiences.

Fix by adding Aurora-specific logic in the replication stats collection.

This also adds support for ignoring Aurora readers using the "skip_if_replica" collector configuration option. This can be used when specifying Aurora instances individually as servers, but only intending to monitor the writer. Note that for such situations we do generally recommend specifying the cluster write endpoint instead, but there is no downside to support this alternate way of only monitoring writers.

### DIFF
--- a/input/postgres/replication.go
+++ b/input/postgres/replication.go
@@ -9,7 +9,7 @@ import (
 	"github.com/pganalyze/collector/util"
 )
 
-const replicationSQL string = `
+const replicationSQLPostgres string = `
 SELECT in_recovery,
 			 CASE WHEN in_recovery THEN NULL ELSE pg_catalog.pg_current_wal_lsn() END AS current_xlog_location,
 			 COALESCE(receive_location, '0/0') >= replay_location AS is_streaming,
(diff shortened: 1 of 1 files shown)


input/postgres/replication.go
```diff
@@ -47,9 +59,30 @@ func GetReplication(ctx context.Context, logger *util.Logger, db *sql.DB, postgr
 	var err error
 	var repl state.PostgresReplication
 	var sourceTable string
+	var replicationSQL string
 
 	if postgresVersion.IsAwsAurora {
-		// Most replication functions are not supported on AWS Aurora Postgres
+		if !auroraDbInstanceIdentifierExists(ctx, db) {
```
keiko713, line 65: what exactly the case/occasion that IsAwsAurora is true but aurora DB instance Identifier doesn't exist?

keiko713: (also the comment of `getIsReplicaAurora` function is not so clear to me; does `auroraDbInstanceIdentifierExists` being false mean that it's very old Aurora version...??)

lfittl: That's right, if `auroraDbInstanceIdentifierExists` is false its a very old Aurora release, and in such situations we don't have an easy way to identify the current instance in the list of replicas. Whilst we could pull the instance ID from the server config, this wasn't super straightforward, and I think its okay to essentially do what we used to do in these cases.

keiko713 (approved):

PR #447, opened 2023-08-11
Update Go version to 1.21, update Dockerfile alpine base image to 3.18

lfittl:
```
Update Go version in use for packages and tests to 1.21

Additionally, this bumps up the minimum Go version to 1.20, and uses
that on Heroku, as Heroku's Go buildpack does not yet support 1.21.
```

```
Dockerfile: Update alpine base image from 3.15 to 3.18
```

```
Update Go download location from "golang.org" to "go.dev" domain

This has been changed upstream, and it seems reasonable to match.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-// +heroku goVersion go1.19
+// +heroku goVersion go1.20
 
 module github.com/pganalyze/collector
 
@@ -81,4 +81,4 @@ require (
 	google.golang.org/grpc v1.32.0 // indirect
 )
 
-go 1.19
+go 1.20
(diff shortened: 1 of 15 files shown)


go.mod
```diff
@@ -1,4 +1,4 @@
-// +heroku goVersion go1.19
```
msepga, line 1: Is this supposed to be `21`?

msepga (approved):
LGTM 👍

I noticed we set the minimum go version to `1.20` instead of `1.21` in `go.mod` - I'm assuming this is intentional, dropping a comment just in case it was a typo

go.mod
lfittl: Its intentional, since Heroku doesn't support 1.21 yet (https://github.com/heroku/heroku-buildpack-go/pull/521)

PR #449, opened 2023-08-15
WIP: Reproducer for bug with Custom 3 log line prefix when app contains brackets

lfittl:
This needs more investigation, but parking this here to document the issue.

### DIFF
--- a/logs/parse_test.go
+++ b/logs/parse_test.go
@@ -181,6 +181,21 @@ var parseTests = []parseTestpair{
 		},
 		true,
 	},
+	{
+		"",
+		"2018-09-27 06:57:02.779 UTC [20194] [user=postgres,db=postgres,app=sidekiq 1.2.3 queues:something[0 of 50 busy]] LOG:  duration: 3000.019 ms  statement: SELECT pg_sleep(3);",
+		nil,
+		state.LogLine{
+			OccurredAt:  time.Date(2018, time.September, 27, 6, 57, 2, 779*1000*1000, time.UTC),
+			Username:    "postgres",
+			Database:    "postgres",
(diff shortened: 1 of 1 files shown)


msakrejda:
This change (making the app name regex non-greedy) fixes the problem and does not break the tests in the existing suite, but it needs more validation:

```diff
diff --git a/logs/parse.go b/logs/parse.go
index 15328926..28075b0c 100644
--- a/logs/parse.go
+++ b/logs/parse.go
@@ -60,7 +60,7 @@ var DbRegexp = `(\S*)`                                                       //
 var AppBeforeSpaceRegexp = `(\S*)`                                           // %a
 var AppBeforeCommaRegexp = `([^,]*)`                                         // %a
 var AppBeforeQuoteRegexp = `([^"]*)`                                         // %a
-var AppInsideBracketsRegexp = `(\[unknown\]|[^,\]]*)`                        // %a
+var AppInsideBracketsRegexp = `(\[unknown\]|.+?)`                            // %a
 var HostRegexp = `(\S*)`                                                     // %h
 var VirtualTxRegexp = `(\d+/\d+)?`                                           // %v
 var LogLineCounterRegexp = `(\d+)` 
```

msakrejda:
Also, it looks like we're currently matching empty app names. I don't think that can actually happen: if you set your application_name to an empty string, the prefix lists '[unknown'] (which we're already handling). We're also matching empty database names, host names, and user names (which I'm guessing also can't happen). I don't think that's causing problems here, but it may lead to other parsing issues. We should maybe change those `*`s to `+?`s...

msakrejda:
I've added the test case to #494 (and it passes there), so I'll close this.

PR #450, opened 2023-08-16
Support n_mod_since_analyze and n_ins_since_vacuum being NULL

keiko713:
Bugfix, introduced with https://github.com/pganalyze/collector/pull/442

Previously, we were accepting `n_mod_since_analyze` being NULL by receiving it with `null.Int`, but this was wrongly updated in that PR to be `int64`, which was causing the following error:

```
RelationStats/Scan: sql: Scan error on column index 13, name "n_mod_since_analyze": converting NULL to int64 is unsupported
```

This `n_mod_since_analyze`, as well as `n_ins_since_vacuum` (the following column) can be NULL, when there is a table definition in `pg_class` but not in `pg_stat_user_tables`. Since this value was converted to 0 anyways when sending snapshot over protobuf, let's use `COALESCE` to set 0 for NULL cases.


input/postgres/relation_stats.go
```diff
@@ -36,8 +36,8 @@ SELECT c.oid,
 			 COALESCE(s.n_tup_hot_upd, 0),
 			 COALESCE(s.n_live_tup, 0),
 			 COALESCE(s.n_dead_tup, 0),
-			 s.n_mod_since_analyze,
-			 %s,
+			 COALESCE(s.n_mod_since_analyze, 0),
+			 COALESCE(%s, 0),
```
lfittl, line 40: Nitpick, but should we move that to `relationStatsSQLInsertsSinceVacuumFieldPg13`? (so for earlier versions, its just `0` instead of `COALESCE(0, 0)`.

lfittl (approved):

input/postgres/relation_stats.go
keiko713: Yeah I was debating that, wondering what would be clearer. `COALESCE(0, 0)` returns 0, and this way we make it explicit that we should really COALESCE any columns that we may introduce next.
That's said, no strong opinion, happy to tweak to the other way if you think it reads better

lfittl: Fair, that's a good argument for having it here.

PR #451, opened 2023-08-16
Release 0.51.1

keiko713:


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.51.1      2023-08-15
+
+* Bugfix: Support NULL n_mod_since_analyze and n_ins_since_vacuum
+  - This mitigates an issue introduced in 0.51.0
```
lfittl, line 6: Detail, but I wonder if we should clarify that this is because of the `pg_stat_user_tables` not existing?

Maybe:

```
* Fix handling of tables that only have an entry in pg_class, but not pg_stat_user_tables
  - Due to a bug introduced in the last release (0.51.0), databases with such tables would accidentally
    error out and be ignored due to n_mod_since_analyze and n_ins_since_vacuum being NULL
```

lfittl (approved):

CHANGELOG.md
keiko713: maybe I'm gonna drop "accidentally" there but otherwise looks good 👍

PR #453, opened 2023-09-06
Provide a better test for statement_parameter filter

benoittgt:
While looking at the code. I noticed that the test for [collector setting](https://pganalyze.com/docs/collector/settings#pii-filtering-settings) `statement_parameter` was not testing the right thing.


seanlinsley (approved):
:octocat:

PR #454, opened 2023-09-13
Improve data directory detection

msakrejda:
Currently, for gathering system stats on self-hosted systems, we look
up the current working directory of the postmaster process. This
usually works, but if customers have an unusual configuration, it can
find the wrong directory.

Instead, use the current method only as a fallback, and look at the
data_directory setting instead. This is readable with
pg_read_all_settings (part of pg_monitor), so it should be accessible
to us in most configurations.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -190,6 +190,9 @@ type ServerConfig struct {
 	// HTTP clients to be used for API connections
 	HTTPClient          *http.Client
 	HTTPClientWithRetry *http.Client
+
+	// Data directory used for system stats
+	DataDirectory string
 }
 
 // SupportsLogDownload - Determines whether the specified config can download logs
--- a/config/config.go
+++ b/config/config.go
@@ -60,7 +60,7 @@ type ServerConfig struct {
 	DbUseIamAuth          bool   `ini:"db_use_iam_auth"`
 
 	// Postgres data directory, as used for system stats (autodetected if unset)
-	DataDirectory string `ini:"db_data_directory"`
+	DbDataDirectory string `ini:"db_data_directory"`
 
 	// We have to do some tricks to support sslmode=prefer, namely we have to
 	// first try an SSL connection (= require), and if that fails change the
--- a/config/config.go
+++ b/config/config.go
@@ -67,6 +67,9 @@ type ServerConfig struct {
 	DbExtraNames []string // Additional databases that should be fetched (determined by additional databases in db_name)
 	DbAllNames   bool     // All databases except template databases should be fetched (determined by * in the db_name list)
 
+	// Postgres data directory, as used for system stats (autodetected if unset)
+	DataDirectory string `ini:"data_directory"`
+
 	AwsRegion               string `ini:"aws_region"`
 	AwsAccountID            string `ini:"aws_account_id"`
 	AwsDbInstanceID         string `ini:"aws_db_instance_id"`
@@ -190,9 +193,6 @@ type ServerConfig struct {
 	// HTTP clients to be used for API connections
 	HTTPClient          *http.Client
 	HTTPClientWithRetry *http.Client
-
-	// Data directory used for system stats
-	DataDirectory string
 }
 
 // SupportsLogDownload - Determines whether the specified config can download logs
--- a/config/config.go
+++ b/config/config.go
@@ -59,6 +59,9 @@ type ServerConfig struct {
 	DbSslKeyContents      string `ini:"db_sslkey_contents"`
 	DbUseIamAuth          bool   `ini:"db_use_iam_auth"`
 
+	// Postgres data directory, as used for system stats (autodetected if unset)
+	DataDirectory string `ini:"db_data_directory"`
+
 	// We have to do some tricks to support sslmode=prefer, namely we have to
 	// first try an SSL connection (= require), and if that fails change the
 	// sslmode to none
@@ -66,9 +69,6 @@ type ServerConfig struct {
 
 	DbExtraNames []string // Additional databases that should be fetched (determined by additional databases in db_name)
 	DbAllNames   bool     // All databases except template databases should be fetched (determined by * in the db_name list)
-
-	// Postgres data directory, as used for system stats (autodetected if unset)
-	DataDirectory string `ini:"data_directory"`
 
 	AwsRegion               string `ini:"aws_region"`
 	AwsAccountID            string `ini:"aws_account_id"`
--- a/config/read.go
+++ b/config/read.go
@@ -119,6 +119,9 @@ func getDefaultConfig() *ServerConfig {
 	if dbSslKeyContents := os.Getenv("DB_SSLKEY_CONTENTS"); dbSslKeyContents != "" {
 		config.DbSslKeyContents = dbSslKeyContents
 	}
+	if dataDirectory := os.Getenv(("DATA_DIRECTORY")); dataDirectory != "" {
+		config.DataDirectory = dataDirectory
+	}
 	if awsRegion := os.Getenv("AWS_REGION"); awsRegion != "" {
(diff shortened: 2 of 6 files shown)


config/config.go
```diff
@@ -190,6 +190,9 @@ type ServerConfig struct {
 	// HTTP clients to be used for API connections
 	HTTPClient          *http.Client
 	HTTPClientWithRetry *http.Client
+
+	// Data directory used for system stats
+	DataDirectory string
```
msakrejda, line 195: Not sure if it's a great idea to put this in `ServerConfig`, but the proxy clients being here seem like a reasonable precedent

msakrejda (commented):
I have not tested this yet. Opening for feedback on the approach first.

input/postgres/data_directory.go
```diff
@@ -0,0 +1,19 @@
...
+
+// SetDataDirectory - Finds the data_directory in the list of settings and sets
+// the Server.Config.DataDirectory field to its value. Does nothing if the
+// setting is not found (e.g., due to permission issues).
+func SetDataDirectory(server *state.Server, settings []state.PostgresSetting) {
+	// N.B.: we're not taking a mutex here, because the data directory will not change location,
+	// so we only run this once at collector startup
+	for _, setting := range settings {
+		if setting.Name == "data_directory" && setting.CurrentValue.Valid {
+			server.Config.DataDirectory = setting.CurrentValue.String
```
lfittl, line 15: I wonder if we should leave the actual setting of things to the caller, like we do in `logs.ValidateLogCollectionConfig`, i.e. just return the data directory value here.

config/config.go
msakrejda: Yeah, I guess it couldn't hurt to allow an override. I don't think CollectionStatus makes sense given the other fields in that struct.

input/postgres/data_directory.go
msakrejda: Sure, makes sense. I think I originally modeled this after SetLogTimezone because I think we'd need a mutex here as well. Since we don't (`data_directory` is not going to change), I'll return a value here.

config/config.go
```diff
@@ -67,6 +67,9 @@ type ServerConfig struct {
 	DbExtraNames []string // Additional databases that should be fetched (determined by additional databases in db_name)
 	DbAllNames   bool     // All databases except template databases should be fetched (determined by * in the db_name list)
 
+	// Postgres data directory, as used for system stats (autodetected if unset)
+	DataDirectory string `ini:"data_directory"`
```
lfittl, line 71: I wonder if we should use the `db_` prefix? (and also call the env variable `DB_DATA_DIRECTORY`?)

That makes it clear its not the data for the collector, but rather the data directory of the DB we're monitoring.

input/postgres/data_directory.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/pganalyze/collector/state"
+)
+
+// GetDataDirectory - Finds the data_directory in the list of settings and returns
+// its current value. Returns an empty string if not found (e.g., the setting is
+// not present due to permissions issues)
+func GetDataDirectory(server *state.Server, settings []state.PostgresSetting) string {
+	// N.B.: we're not taking a mutex here, because the data directory will not
+	// change location, so we only run this once at collector startup
```
lfittl, line 12: FWIW, I think we can drop this comment now (but maybe move it to the call site of this function?), since we're just getting the setting, not modifying anything.

main.go
```diff
@@ -557,6 +557,11 @@ func checkOneInitialCollectionStatus(ctx context.Context, server *state.Server,
 	if err != nil {
 		return err
 	}
+	autodetectedDataDir := postgres.GetDataDirectory(server, settings)
```
lfittl, line 560: Maybe `dataDirectorySetting`? (not sure if we're doing much auto-detecting here specifically besides getting that setting)

lfittl:
> I have not tested this yet. Opening for feedback on the approach first.

Curious, did you run an end to end test with this?

Wondering whether we should do a full test in a VM before we merge, or whether we feel its been sufficiently tested.

msakrejda:
@lfittl I've instrumented the code to verify that the correct setting is picked up, but I have not done an end-to-end test. I'm trying to figure out something that would do the wrong thing on `main` but the right thing here.

msakrejda:
Hmm, so I'm actually unable to reproduce the problem. If I start `postgres` on one filesystem partition and point it to a data directory on a different partition, `/proc/<pid>/cwd` actually points to the data directory. Tried this on both 15 and 10 with the same results. I think the change is safe, but I can't think of how to test it without being able to reproduce the problem. Maybe we should wait on this patch until we can reproduce the original problem.

lfittl:
> Hmm, so I'm actually unable to reproduce the problem. If I start `postgres` on one filesystem partition and point it to a data directory on a different partition, `/proc/<pid>/cwd` actually points to the data directory. Tried this on both 15 and 10 with the same results. I think the change is safe, but I can't think of how to test it without being able to reproduce the problem. Maybe we should wait on this patch until we can reproduce the original problem.

I wonder if one way to reproduce it could be to have multiple Postgres installations?

i.e. if the Postgres PID search in https://github.com/pganalyze/collector/blob/main/helper/main.go#L22 gets the wrong process, it would lead to the detected data directory being incorrect.

msakrejda:
Okay, confirmed that situation can cause bogus reporting on `main` but is fixed with this branch. Another case that this should improve is if Postgres is running as a user other than "postgres" (probably not common, but maybe there are some obscure configurations or packaging setups out there that use "postgresql" or something else).

main.go
```diff
@@ -557,6 +557,13 @@ func checkOneInitialCollectionStatus(ctx context.Context, server *state.Server,
 	if err != nil {
 		return err
 	}
+
+	// We don't need a mutex here, because we only check once at startup
+	dataDirectorySetting := postgres.GetDataDirectory(server, settings)
+	if dataDirectorySetting != "" && server.Config.DataDirectory == "" {
+		server.Config.DataDirectory = dataDirectorySetting
+	}
```
lfittl, line 565: Technically this could also be written as:

```golang
	if server.Config.DataDirectory == "" {
		server.Config.DataDirectory = postgres.GetDataDirectory(server, settings)
	}
```

config/read.go
```diff
@@ -119,6 +119,9 @@ func getDefaultConfig() *ServerConfig {
 	if dbSslKeyContents := os.Getenv("DB_SSLKEY_CONTENTS"); dbSslKeyContents != "" {
 		config.DbSslKeyContents = dbSslKeyContents
 	}
+	if dataDirectory := os.Getenv(("DB_DATA_DIRECTORY")); dataDirectory != "" {
```
lfittl, line 122: Looks like you got extra parentheses around `"DB_DATA_DIRECTORY"` that are not needed.

config/config.go
```diff
@@ -59,6 +59,9 @@ type ServerConfig struct {
 	DbSslKeyContents      string `ini:"db_sslkey_contents"`
 	DbUseIamAuth          bool   `ini:"db_use_iam_auth"`
 
+	// Postgres data directory, as used for system stats (autodetected if unset)
+	DataDirectory string `ini:"db_data_directory"`
```
lfittl, line 63: Should we call this variable `DbDataDirectory` for consistency?

lfittl (approved):
:+1: Some small remaining comments, but looks ready to merge overall.

And thanks for doing the extra testing!

config/read.go
msakrejda: Thanks: I think the Go mode in VS Code switched how it handles completion and it's still tripping me up.

config/config.go
msakrejda: Yeah, good call.

PR #456, opened 2023-09-20
Add logs for test run when system state collection is skipped

keiko713:
We try to collect system data (e.g. CPU, memory usage, etc., shown in "System" page) when it's a self hosted database server and the database server is in the same instance as the collector's.
However, sometimes people would set the db host different from localhost or 127.0.0.1 even though the database is running locally. In that case, setting `always_collect_system_data` will help the situation, but it's not so obvious.
https://pganalyze.com/docs/collector/settings#self-managed-servers

In this PR, add a log during the test run if we detect 1) the server is self hosted 2) but it's detected as a remote host.

### DIFF
--- a/input/system/system.go
+++ b/input/system/system.go
@@ -56,7 +56,7 @@ func GetSystemState(config config.ServerConfig, logger *util.Logger, globalColle
 			// didn't detect the collector is running on the same instance as
 			// the database server.
 			// Leave logs for if this is a test run.
-			logger.PrintInfo("Remote host (%s) was specified for the database address, skip collecting system state. Consider enabling always_collect_system_data if it's a local database with a non-local IP address", dbHost)
+			logger.PrintInfo("Skipping collection of system state: remote host (%s) was specified for the database address. Consider enabling always_collect_system_data if the database is running on the same system as the collector", dbHost)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


input/system/system.go
```diff
@@ -46,8 +46,18 @@ func GetSystemState(config config.ServerConfig, logger *util.Logger) (system sta
...
+		system.Info.Type = state.AivenSystem
 	} else if dbHost == "" || dbHost == "localhost" || dbHost == "127.0.0.1" || config.AlwaysCollectSystemData {
 		system = selfhosted.GetSystemState(config, logger)
+	} else {
+		if globalCollectionOpts.TestRun {
+			// Detected as self hosted, but not collecting system state as we
+			// didn't detect the collector is running on the same instance as
+			// the database server.
+			// Leave logs for if this is a test run.
+			logger.PrintInfo("Remote host (%s) was specified for the database address, skip collecting system state. Consider enabling always_collect_system_data if it's a local database with a non-local IP address", dbHost)
```
msakrejda, line 59: We're not super-consistent about error message format, but we generally put the thing that's wrong first, and then an explanation. Also, I think "local database with a non-local IP address" is somewhat confusing, especially since we are not very thorough on checking local connections above (e.g., we ignore IPv6 entirely, and all but one IPv4 loopback address https://superuser.com/questions/393700/what-is-the-127-0-0-2-ip-address-for ).

How about

```suggestion
			logger.PrintInfo("Skipping collection of system state: remote host (%s) was specified for the database address. Consider enabling always_collect_system_data if the database is running on the same system as the collector", dbHost)
```

msakrejda (approved):

input/system/system.go
keiko713: Nice, your suggestion is much better, will apply.

PR #458, opened 2023-10-03
OpenTelemetry integration: Allow exporting EXPLAIN plans as trace spans

lfittl:
This is an experimental feature that allows configuring the collector to send an OTel tracing span for each processed EXPLAIN plan with an associated "traceparent" query tag (e.g. set by sqlcommenter) to the configured OpenTelemetry endpoint.

To configure the endpoint set the new config setting otel_exporter_otlp_endpoint / OTEL_EXPORTER_OTLP_ENDPOINT, with a endpoint string like "http://localhost:4318".

The tracing span contains the runtime of the EXPLAIN plan, as determined from the logs, as well as a backlink to pganalyze as the "url.full" span attribute, to allow accessing the full details within the pganalyze app.

Overall this aims to enable easy access of pganalyze for slow traces collected in an APM tool, and helps identify when the database was the bottleneck (and why). It requires the "traceparent" query tag to be present on logged EXPLAIN plans, e.g. added by using sqlcommenter within the application.

## TODO

* [x] Figure out why we have such a massive diff in `vendor/`
  * It appears this is mainly indirect module updates that were done as part of adding the `opentelemetry-go` module, plus some new libraries that are pulled in -- at the very least we should split the commit that adds this feature from the module updates in a separate commit
  * Additionally we are pulling in gRPC-related modules, despite using the HTTP exporter explicitly. Is there a way for us to avoid this? (and if not, we should support setting `grpc://` as the protocol as well for the endpoint)


config/read.go
```diff
@@ -254,6 +254,9 @@ func getDefaultConfig() *ServerConfig {
 	if filterQueryText := os.Getenv("FILTER_QUERY_TEXT"); filterQueryText != "" {
 		config.FilterQueryText = filterQueryText
 	}
+	if otelExporterOtlpEndpoint := os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT"); otelExporterOtlpEndpoint != "" {
+		config.FilterQueryText = otelExporterOtlpEndpoint
```
msakrejda, line 258:
```suggestion
		config.OtelExporterOtlpEndpoint = otelExporterOtlpEndpoint
```

logs/querysample/tracing.go
```diff
@@ -0,0 +1,152 @@
...
+		return nil
+	}
+
+	tags := make(map[string]string)
+	for _, token := range tokens.GetTokens() {
+		var comment string
+		if token.Token == pg_query.Token_SQL_COMMENT {
+			comment = query[token.Start+2 : token.End]
+		} else if token.Token == pg_query.Token_C_COMMENT {
+			comment = query[token.Start+2 : token.End-2]
```
msakrejda, line 52: The math here is because you're turning `--comment1` into `comment1` and `/*comment2*/` into `comment2`, right? It might be good to add a... comment to that effect.

logs/querysample/tracing.go
```diff
@@ -0,0 +1,152 @@
...
+				keyAndValue := strings.SplitN(part, ":", 2)
+				tags[keyAndValue[0]] = keyAndValue[1]
+			} // TODO: Support other formats
+		}
+	}
+	return tags
+}
+
+func initProvider(ctx context.Context, endpoint string) (*sdktrace.TracerProvider, func(context.Context) error, error) {
+	res, err := resource.New(ctx,
```
msakrejda, line 71: It might be nice to alias this import so we're not left wondering what this super-generic "resource" package is here when reading this later.

logs/querysample/tracing.go
```diff
@@ -0,0 +1,152 @@
...
+		if _, ok := queryTags["traceparent"]; ok {
+			prop := propagation.TraceContext{}
+			ctx := prop.Extract(context.Background(), propagation.MapCarrier(queryTags))
+
+			tracer := tracerProvider.Tracer(
+				"pganalyze-collector",
+				trace.WithInstrumentationVersion(util.CollectorVersion),
+				trace.WithSchemaURL(semconv.SchemaURL),
+			)
+			duration, _ := time.ParseDuration(fmt.Sprintf("-%0.2fms", sample.RuntimeMs))
```
msakrejda, line 144:
```suggestion
			duration := -1 * sample.RuntimeMs * time.Millisecond
```

logs/querysample/tracing.go
```diff
@@ -0,0 +1,152 @@
...
+	return tracerProvider, tracerProvider.Shutdown, nil
+}
+
+func ReportQuerySamplesAsTraceSpans(ctx context.Context, server *state.Server, grant state.GrantLogs, samples []state.PostgresQuerySample) {
+	endpoint := server.Config.OtelExporterOtlpEndpoint
+	if endpoint == "" {
+		return
+	}
+
+	tracerProvider, shutdown, err := initProvider(ctx, server.Config.OtelExporterOtlpEndpoint)
```
msakrejda, line 119: We're re-initializing this for every log snapshot, right? I'm not familiar with the SDK, so not sure if that's the recommended approach, but it seems like it'd be more expensive than initializing once per collector startup (if nothing else, in terms of keeping the TCP connection open). But reorganizing that could get a bit messy, so we can start with this.

msakrejda (approved):
I noticed one copy-paste-o and have some minor comments, but otherwise it looks good.

config/read.go
lfittl: 🤦  Good catch.

logs/querysample/tracing.go
lfittl: Yeah, good catch. As you can imagine its not the recommended approach to reinitialize it. The overall idea is that you initialize this once and then your whole app uses the same trace provider via some global helpers.

That said, it seemed easiest for now to do it like this to keep this code in one place, and to ensure all spans are sent when the snapshot is processed (i.e. not with a delay) - I'll leave a note to address this in a future revision (I do think it makes sense to centralize this, just need to spend more time testing that).

lfittl:
> I noticed one copy-paste-o and have some minor comments, but otherwise it looks good. As discussed, it'd be good to break out the vendor changes, though.

Thanks for the review - the vendor changes did actually end up being almost exclusively because of module updates (luckily), see https://github.com/pganalyze/collector/pull/458/commits/688c6a8eb133ee481e4069e2921c6bf75d7c83ae

PR #460, opened 2023-10-04
OpenTelemetry Integration: Refactoring, support configuring OTLP headers

lfittl:
See individual commits.

### DIFF
--- a/config/read.go
+++ b/config/read.go
@@ -15,6 +15,12 @@ import (
 	"time"
 
 	"github.com/go-ini/ini"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
+	sdkresource "go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.21.0"
 	"golang.org/x/net/http/httpproxy"
 
 	"github.com/pganalyze/collector/util"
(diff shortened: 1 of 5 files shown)


logs/querysample/tracing.go
```diff
@@ -53,7 +53,10 @@ func ExportQuerySamplesAsTraceSpans(ctx context.Context, server *state.Server, l
 			startTime := sample.OccurredAt.Add(duration)
 			endTime := sample.OccurredAt
 			_, span := tracer.Start(ctx, otelSpanName, trace.WithTimestamp(startTime))
-			span.SetAttributes(attribute.String("url.full", urlToSample(server, grant, sample)))
+			// See https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/database/
+			// however note that "db.postgresql.plan" is non-standard.
+			span.SetAttributes(attribute.String("db.system", "postgresql"))
+			span.SetAttributes(attribute.String("db.postgresql.plan", urlToSample(server, grant, sample)))
```
msakrejda, line 59: I don't really follow this change. Neither `url.full` nor `db.postgresql.plan` are mentioned in the URL cited from a quick skim, and I don't understand what "plan" means in this context.

config/read.go
```diff
@@ -351,6 +360,86 @@ func CreateEC2IMDSHTTPClient(conf ServerConfig) *http.Client {
...
+			if !found {
+				return nil, nil, fmt.Errorf("unsupported header setting: missing '='")
+			}
+			name, err := url.QueryUnescape(nameEscaped)
+			if err != nil {
+				return nil, nil, fmt.Errorf("unsupported header setting, could not unescape header name: %s", err)
+			}
+			value, err := url.QueryUnescape(valueEscaped)
+			if err != nil {
+				return nil, nil, fmt.Errorf("unsupported header setting, could not unescape header valie: %s", err)
```
msakrejda, line 400:
```suggestion
				return nil, nil, fmt.Errorf("unsupported header setting, could not unescape header value: %s", err)
```

config/read.go
```diff
@@ -351,6 +360,86 @@ func CreateEC2IMDSHTTPClient(conf ServerConfig) *http.Client {
...
+		}
+	case "grpc":
+		// For now we always require TLS for gRPC connections
+		opts := []otlptracegrpc.Option{otlptracegrpc.WithEndpoint(u.Host)}
+		if headers != nil {
+			opts = append(opts, otlptracegrpc.WithHeaders(headers))
+		}
+		traceExporter, err = otlptracegrpc.New(ctx, opts...)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to create HTTP trace exporter: %w", err)
```
msakrejda, line 427:
```suggestion
			return nil, nil, fmt.Errorf("failed to create GRPC trace exporter: %w", err)
```

msakrejda (approved):

logs/querysample/tracing.go
lfittl: The problem is that there isn't really a proper place to put information like this - my initial thinking was that `url.full` may be a way to get tracing platforms to actually turn the attribute value into a clickable link, but that does not appear to be the case. It appears `url.full` is intended to represent the URL that an HTTP call from an application goes to, so not really the same as this use case.

I only today discovered that there was the semantic convention for database calls (and how to represent them in spans), and thus the addition of `db.system`. I then thought "how could we represent a query plan within this structure", and since there is some existing cases where there is database-system specific data (https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/database/#call-level-attributes-for-specific-technologies) this seemed like the best name.

msakrejda: Got it, thanks for clarifying.

PR #462, opened 2023-10-05
Activity snapshots: Normalize queries for "filter_query_sample = normalize"

lfittl:
For historic reasons, we were only normalizing pg_stat_activity query texts when `filter_query_sample` was set to `all`, not `normalize`, which has a similar intent.

Whilst we could also control the filtering of pg_stat_activity query texts with the `filter_query_text` setting (which today controls whether pg_stat_statements query texts are re-normalized, and is enabled by default), this is intentionally
not done here. The main motivation to not do this for now are performance implications of running normalize every 10 seconds for many active queries, and we may adjust this in a future commit based on more benchmarks.

### DIFF
--- a/output/compact_activity.go
+++ b/output/compact_activity.go
@@ -12,7 +12,7 @@ import (
 func SubmitCompactActivitySnapshot(ctx context.Context, server *state.Server, grant state.Grant, collectionOpts state.CollectionOpts, logger *util.Logger, activityState state.TransientActivityState) error {
 	as, r := transform.ActivityStateToCompactActivitySnapshot(server, activityState)
 
-	if server.Config.FilterQuerySample != "none" {
+	if server.Config.FilterQuerySample != "" && server.Config.FilterQuerySample != "none" {
 		for idx, backend := range as.Backends {
 			// Normalize can be slow, protect against edge cases here by checking for cancellations
 			select {
--- a/output/compact_activity.go
+++ b/output/compact_activity.go
@@ -7,13 +7,12 @@ import (
 	"github.com/pganalyze/collector/output/transform"
 	"github.com/pganalyze/collector/state"
 	"github.com/pganalyze/collector/util"
-	pg_query "github.com/pganalyze/pg_query_go/v4"
 )
 
 func SubmitCompactActivitySnapshot(ctx context.Context, server *state.Server, grant state.Grant, collectionOpts state.CollectionOpts, logger *util.Logger, activityState state.TransientActivityState) error {
 	as, r := transform.ActivityStateToCompactActivitySnapshot(server, activityState)
 
-	if server.Config.FilterQuerySample == "all" {
+	if server.Config.FilterQuerySample != "none" {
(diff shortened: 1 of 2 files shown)


keiko713:
memo: https://pganalyze.com/docs/collector/settings#pii-filtering-settings

> Either `none`, `normalize` or `all`

> By default the pganalyze collector does not filter data. Our recommended configuration for servers containing sensitive data is as follows: `filter_query_sample: normalize`

> Note the `normalize` setting for `filter_query_sample`, which will remove all query parameter values from query samples, including those contained within automatically collected EXPLAIN plans.

output/compact_activity.go
```diff
@@ -7,21 +7,20 @@ import (
...
-	pg_query "github.com/pganalyze/pg_query_go/v4"
 )
 
 func SubmitCompactActivitySnapshot(ctx context.Context, server *state.Server, grant state.Grant, collectionOpts state.CollectionOpts, logger *util.Logger, activityState state.TransientActivityState) error {
 	as, r := transform.ActivityStateToCompactActivitySnapshot(server, activityState)
 
-	if server.Config.FilterQuerySample == "all" {
+	if server.Config.FilterQueryText != "none" {
 		for idx, backend := range as.Backends {
 			// Normalize can be slow, protect against edge cases here by checking for cancellations
```
keiko713, line 17: Does this mean that we will likely have more collectors going through this potentially slow part? (as `FilterQuerySample == "all"` is somewhat "special case" (aka not default), whereas `FilterQueryText != "none"` is actually default. Is it okay change to make?

lfittl: Thats a fair point, and I could indeed see that being a problem on collectors with many servers configured. I wonder if we should maybe not make this the default for now (since query samples are also not filtered by default), and just keep using the filter_query_sample setting here, but also do this normalization if its set to "normalize" (not just "all").

lfittl: Adjust this to continue working based on `filter_query_sample` for now, but to also do this when its set to `normalize` (not just `all`).

keiko713 (approved):
Thanks for adjusting, I think it's good to do this way first and we can revisit if it's okay to change more widely in the future.

Minor note: it might be nice to have some comment for why we're passing "unparseable" to `util.NormalizeQuery`. I don't think the behavior will be much different here if we pass it or `server.Config.FilterQueryText` so I'm gonna approve anyways, but it would be nice if the reason is mentioned well. Maybe you're doing so so that we can control the behavior when the error happens during the normalization? (and make sure not to use the original query with the error case?)

lfittl:
> Minor note: it might be nice to have some comment for why we're passing "unparseable" to `util.NormalizeQuery`. I don't think the behavior will be much different here if we pass it or `server.Config.FilterQueryText` so I'm gonna approve anyways, but it would be nice if the reason is mentioned well. Maybe you're doing so so that we can control the behavior when the error happens during the normalization? (and make sure not to use the original query with the error case?)

Good point, that's a bit confusing - adding a comment.

PR #463, opened 2023-10-11
Postgres 14+: Include toplevel attribute in statement statistics key

lfittl:
This could have caused statistics to be incorrect in Postgres 14+ when the same query was called both from within a function (toplevel=false) and directly (toplevel=true), with pg_stat_statements track set to "all".

If affected, the issue may have shown by bogus statistics being recorded, for example very high call counts, since the statement stats diff would not have used the correct reference.

Fix by including toplevel attribute for internal tracking, matching the [pgssHashKey struct](https://github.com/postgres/postgres/blob/master/contrib/pg_stat_statements/pg_stat_statements.c#L148) in pg_stat_statements.c where this data originates.

### DIFF
--- a/input/postgres/statements.go
+++ b/input/postgres/statements.go
@@ -14,12 +14,15 @@ import (
 )
 
 // pg_stat_statements 1.3+ (Postgres 9.5+)
-const statementSQLOptionalFieldsMinorVersion3 = "queryid, min_time, max_time, mean_time, stddev_time"
+const statementSQLOptionalFieldsMinorVersion3 = "queryid, min_time, max_time, mean_time, stddev_time, NULL"
 const statementSQLTotalTimeFieldDefault = "total_time"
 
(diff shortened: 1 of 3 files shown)


keiko713 (approved):
Looks good 👍 

Just wanted to double check:
* with this, queryid 123 with toplevel true and false stats will be stored (in the collector) separately
* when a full snapshot is made, we combine two stats data (of toplevel true and false, more precisely, we also combine the same fingerprint queries too) into one query stat

Is above accurate?

lfittl:
> Looks good 👍
> 
> Just wanted to double check:
> 
> * with this, queryid 123 with toplevel true and false stats will be stored (in the collector) separately
> * when a full snapshot is made, we combine two stats data (of toplevel true and false, more precisely, we also combine the same fingerprint queries too) into one query stat
> 
> Is above accurate?

Yes, that's correct.

PR #466, opened 2023-10-17
Support query tags with SQLCommenter format

keiko713:
See the spec, especially for the parsing in https://google.github.io/sqlcommenter/spec/#parsing

This PR will support query tags with SQLCommenter format, like `key='value'` (value is surrounded by single quotes). I think it's quite essential to support this as I believe Active Record Query Logs uses SQLCommenter format by default.
I confirmed that the current code doesn't support picking up `traceparent` when the format is SQLCommenter, and this should fix it.

https://api.rubyonrails.org/classes/ActiveRecord/QueryLogs.html

### DIFF
--- a/logs/querysample/tags.go
+++ b/logs/querysample/tags.go
@@ -34,19 +34,19 @@ func parseTags(query string) map[string]string {
 				// Parse sqlcommenter format (key='value')
 				keyAndValue := strings.SplitN(part, "=", 2)
 				// Remove surrounding single quotes (if present)
-				value := keyAndValue[1]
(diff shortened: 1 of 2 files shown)


keiko713:
Oh nice, I suspected that there should be similar implementation there but didn't know that it covers way more there! Thanks for the pointer 👍 
I'll port some tests over here. I'll only port the ones that the collector supports for now, I think supporting marginalia and sqlcommenter are okay enough start for a traceparent so I won't try to implement other formats.

lfittl (approved):
:+1:

PR #474, opened 2023-11-16
Allow collector to run in verbose log mode with Docker

keiko713:
With this change, `docker run` will take more args than just `test`, `test-explain`, or `collector`. This allows users to run things like `docker run --env-file pganalyze_collector.env quay.io/pganalyze/collector:stable test -v` to run the test with verbose mode.

----
(previous comment)

Please correct me if I'm doing something useless, though I always find it difficult to run with the debug log mode with a Docker setup. Maybe I can run `docker run .... quay.io/pganalyze/collector:stable /home/pganalyze/collector --statefile=/state/pganalyze-collector.state --no-log-timestamps --verbose`?
Not sure, but I thought it won't hurt to add some new command, so that we can do `docker run .... quay.io/pganalyze/collector:stable /home/pganalyze/collector collector-verbose` to launch the collector with the verbose log mode.

### DIFF
--- a/contrib/docker-entrypoint.sh
+++ b/contrib/docker-entrypoint.sh
@@ -24,7 +24,7 @@ if [ "$1" = 'collector' ]; then
 fi
 
 if [ "$1" = 'collector-verbose' ]; then
-  eval $CMD_PREFIX /home/pganalyze/collector --verbose --statefile=/state/pganalyze-collector.state --no-log-timestamps
+  eval $CMD_PREFIX /home/pganalyze/collector --statefile=/state/pganalyze-collector.state --no-log-timestamps --verbose
 fi
 
 eval $CMD_PREFIX "$@"
--- a/contrib/docker-entrypoint.sh
+++ b/contrib/docker-entrypoint.sh
@@ -23,4 +23,8 @@ if [ "$1" = 'collector' ]; then
   eval $CMD_PREFIX /home/pganalyze/collector --statefile=/state/pganalyze-collector.state --no-log-timestamps
 fi
 
+if [ "$1" = 'collector-verbose' ]; then
+  eval $CMD_PREFIX /home/pganalyze/collector --verbose --statefile=/state/pganalyze-collector.state --no-log-timestamps
+fi
+
 eval $CMD_PREFIX "$@"
--- a/contrib/docker-entrypoint.sh
+++ b/contrib/docker-entrypoint.sh
@@ -12,19 +12,14 @@ if [ $(id -u) = 0 ]; then
 fi
 
 if [ "$1" = 'test' ]; then
-  eval $CMD_PREFIX /home/pganalyze/collector --test --no-log-timestamps --no-reload
+  shift
+  eval $CMD_PREFIX /home/pganalyze/collector --test --no-log-timestamps --no-reload "$@"
+elif [ "$1" = 'test-explain' ]; then
+  shift
+  eval $CMD_PREFIX /home/pganalyze/collector --test-explain --no-log-timestamps "$@"
+elif  [ "$1" = 'collector' ]; then
+  shift
+  eval $CMD_PREFIX /home/pganalyze/collector --statefile=/state/pganalyze-collector.state --no-log-timestamps "$@"
+else
+  eval $CMD_PREFIX "$@"
 fi
-
-if [ "$1" = 'test-explain' ]; then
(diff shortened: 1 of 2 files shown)


benoittgt:
This is a great idea. We had some surprises when we wanted to turn one verbose mode. We ended nearly providing again all the entrypoint commands to our helm chart.

contrib/docker-entrypoint.sh
```diff
@@ -23,4 +23,8 @@ if [ "$1" = 'collector' ]; then
   eval $CMD_PREFIX /home/pganalyze/collector --statefile=/state/pganalyze-collector.state --no-log-timestamps
```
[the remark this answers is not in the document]
lfittl: Note that this may be [bash specific syntax](https://stackoverflow.com/questions/3811345/how-to-pass-all-arguments-passed-to-my-bash-script-to-a-function-of-mine) and I haven't tested this -- notably the entrypoint right now specifies `/bin/sh` at the top, not `/bin/bash`.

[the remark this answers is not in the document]
keiko713: Uhm, I'm not sure about this.
This does give more flexibility, though we still limit the statefile path and `no-log-timestamps` option, also this could lead the mix usage of args and env var for various configs (which won't work well I think). Given that we have `eval $CMD_PREFIX "$@"` at the end already for folks to do whatever they want to, I think the preset commands are good to be a safe guard with limited usage side, rather than giving more flexibility.

Some alternate to make verbose more generic (to support test with -v too, which is a good point), how about accepting `verbose` as a second argument? Like the following, and update the each "preset" commands to have this suffix:

```sh
CMD_SUFFIX=""
if [ "$2" = 'verbose' ]; then
  CMD_SUFFIX="--verbose"
fi
```

Also just FYI, `{@:2}` is indeed a bash specific syntax and I think alpine image doesn't have bash (quick tested and it didn't work).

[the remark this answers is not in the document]
benoittgt: In our collector setting we tried initially to use `collector -v` but got trapped by this [if statement](https://github.com/pganalyze/collector/blob/61a4ae8f22260f91907db2f35a354f4c1a0f279d/contrib/docker-entrypoint.sh#L22-L24) that replaced the all command. I think it could be a good idea if other args are kept.

contrib/docker-entrypoint.sh
```diff
@@ -12,15 +12,14 @@ if [ $(id -u) = 0 ]; then
...
-  eval $CMD_PREFIX /home/pganalyze/collector --test --no-log-timestamps --no-reload
+  shift
+  eval $CMD_PREFIX /home/pganalyze/collector --test --no-log-timestamps --no-reload "$@"
+elif [ "$1" = 'test-explain' ]; then
+  shift
+  eval $CMD_PREFIX /home/pganalyze/collector --test-explain --no-log-timestamps "$@"
+elif  [ "$1" = 'collector' ]; then
+  shift
+  eval $CMD_PREFIX /home/pganalyze/collector --statefile=/state/pganalyze-collector.state --no-log-timestamps "$@"
+else
```
lfittl, line 23: Technically I don't think the `else` is necessary, since we are utilizing `exec` in the `CMD_PREFIX` always, but I guess this doesn't hurt.

contrib/docker-entrypoint.sh
```diff
@@ -12,15 +12,14 @@ if [ $(id -u) = 0 ]; then
 fi
 
 if [ "$1" = 'test' ]; then
-  eval $CMD_PREFIX /home/pganalyze/collector --test --no-log-timestamps --no-reload
+  shift
```
lfittl, line 15: Are there any compatibility problems with this if `/bin/sh` is not bash? (should we use `/bin/bash` instead, assuming that's supported on Alpine?)

lfittl (approved):

contrib/docker-entrypoint.sh
keiko713: I checked this by creating a simple shell script and make sure that it worked with `/bin/sh`. Also Keiko 2023 Nov is saying "alpine image doesn't have bash (quick tested and it didn't work)" so I think it's okay to keep this way

contrib/docker-entrypoint.sh
keiko713: Yeah, I always felt "wait doesn't that last `eval $CMD_PREFIX "$@"` always called?" (which is not true with `exec` indeed), so I figured that this will help a bit readability.

PR #475, opened 2023-11-17
Support pganalyze tracestate to set start time of the span

keiko713:
### Problems

Currently, we generate an OpenTelemetry tracing span using `sample.OccurredAt` and `sample.RuntimeMs` derived from logs. This approach generally functions well, however, it can encounter precision issues due to the log timestamp format.
Specifically, timestamps with lower precision (e.g., hundredths of a second like `12:34:56.78`, seen with RDS) result in spans ending at the 100th millisecond mark. This leads to misalignment, causing the span to appear earlier or later than it should.

### Solution

To solve this, we would like to introduce a new tracestate for pganalyze. With tracestate, we can provide a vendor-specific trace identification information. We will use the key `pganalyze` for this, while the value is a list, containing a member with semicolon (`;`) separated format. Each member should also have key and value, separated with colon (`:`).
We use a member key `t` to specify the start time of the query, the format should be the Unix time in second, allowing decimals to present up to nano seconds.
Here is an example of tracestate to be passed:

```
tracestate:pganalyze=t:1697666938.6297212;x=123;y=567
```

See [OpenTelemetry document](https://opentelemetry.io/docs/specs/otel/trace/tracestate-handling/) for how they use this tracestate. When this is used together with OpenTelemetry tracestate, this will look like:

```
tracestate:pganalyze=t:1697666938.6297212;x=123;y=567,ot=p:8;r:62;k1:13
```

### PR summary

With this PR, we will add the capability for the collector to read a tracestate for pganalyze and use the time specified with `t` as the start time of the tracing span.

### DIFF
--- a/logs/querysample/tracing.go
+++ b/logs/querysample/tracing.go
@@ -5,6 +5,8 @@ import (
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/pganalyze/collector/state"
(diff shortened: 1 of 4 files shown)


logs/querysample/tracing.go
```diff
@@ -32,6 +34,41 @@ func urlToSample(server *state.Server, grant state.GrantLogs, sample state.Postg
...
+	if pganalyzeState := traceState.Get("pganalyze"); pganalyzeState != "" {
+		// A pganalyze traceState allows the client to pass the query start time (sent time)
+		// on the client side, in nano second precision, like pganalyze=t:1697666938.6297212
+		// If there are multiple values in a pganalzye traceState, they are separated by semicolon
+		// like pganalyze=t:1697666938.6297212;x=123
+		for _, part := range strings.Split(strings.TrimSpace(pganalyzeState), ";") {
+			if strings.Contains(part, ":") {
+				keyAndValue := strings.SplitN(part, ":", 2)
+				if strings.TrimSpace(keyAndValue[0]) == "t" {
+					if startInSec, err := strconv.ParseFloat(keyAndValue[1], 64); err == nil {
```
keiko713, line 47: I really feel like there should be some easy way to do this, I might be simply missing something.
Also, I _could_ try some another way, like splitting `keyAndValue[1]` with `.` then parsing both of them to int. In that way, I don't need to fight with float in test.

msakrejda: +1 for parsing this as two separate values. It looks like Go's `time.Time` has an [internal representation](https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/time/time.go;l=129-142) that allows it more nanosecond precision than can be stored in a 64-bit float. I don't think it will _really_ matter in practice, but we might as well get this right. And simplifying the test is a nice bonus.

Since that's going to complicate this branch, I think we should extract that to a separate function like `func parseTraceStateStartTime(value string) (time.Time, error)` (no strong feelings on the name, just throwing something out there).

logs/querysample/tracing_test.go
```diff
@@ -0,0 +1,83 @@
...
+	otelTraceState, err := traceState.Insert("ot", "p:8;r:62")
+	if err != nil {
+		t.Fatalf("Failed to initialize object: %v", err)
+	}
+	pganalyzeTraceStateWithoutT, err := traceState.Insert("pganalyze", "x:foo;y:bar")
+	if err != nil {
+		t.Fatalf("Failed to initialize object: %v", err)
+	}
+	// inserting the same key will update the value
+	pganalyzeTraceState, err := traceState.Insert("pganalyze", "t:1697666938.6297212")
```
msakrejda, line 31: Does this mutate `traceState`? If so, it might be nice to just initialize a new one to better isolate the individual tests.

keiko713: ah I was actually wrong, this is returning the copied tracestate 🤦 will update the comment, also will tweak things a bit

logs/querysample/tracing.go
```diff
@@ -32,6 +34,41 @@ func urlToSample(server *state.Server, grant state.GrantLogs, sample state.Postg
...
+						startSec := int64(startInSec)
+						startNanoSec := int64(startInSec*1e9) - (startSec * 1e9)
+						startTime = time.Unix(startSec, startNanoSec).UTC()
+						// With this, we're adding the query duration to the start time.
+						// This could result creating inaccurate spans, as the start time passed
+						// from the client side using tracestate is the time of the query is sent
+						// from the client to the server.
+						// This means, we will ignore the network time between the client and the
+						// server, as well as the machine clock different between them.
```
lfittl, line 56:
```suggestion
						// server, as well as the machine clock difference between them.
```

lfittl (approved):

logs/querysample/tracing.go
keiko713: Ah @lfittl sorry looks like you just approved (forgot to push) but I did some refactoring around here and put that functionality to util. Well hope you're okay with that? Or feel free to 👀 again.

lfittl: @keiko713 Refactoring looks good!

PR #476, opened 2023-11-20
Track extended statistics (CREATE STATISTICS) information

lfittl:
🍐ing PR in snapshot side:  https://github.com/pganalyze/collector-snapshot/pull/35

Start collecting the extended statistics of the table and report them as a part of relation information.


README.md
```diff
@@ -72,6 +72,12 @@ $$
   FROM pg_catalog.pg_stats;
 $$ LANGUAGE sql VOLATILE SECURITY DEFINER;
 
+CREATE OR REPLACE FUNCTION pganalyze.get_relation_stats_ext() RETURNS SETOF pg_stats_ext AS
+$$
+  /* pganalyze-collector */ SELECT row_to_json(se.*)::jsonb - 'most_common_vals'
+  FROM pg_catalog.pg_stats_ext se;
+$$ LANGUAGE sql VOLATILE SECURITY DEFINER;
```
keiko713, line 79: [note for reviewer] We want to introduce this helper function to grab the stats data, but ensure that we won't grab the sensitive data. `most_common_vals` could include the actual values of the table, so we want to make sure excluding this. We do the similar thing above, specifying `NULL::anyarray` for `pg_stats` for the column like `most_common_vals` or `histogram_bounds`.
However, we can't do the similar way as `pg_stats` above because the columns of `pg_stats_ext` differs depending on the Postgres version (from 15, `inherited` column is introduced). As a workaround, we grab columns as jsonb except `most_common_vals`.

input/postgres/relation_stats_ext.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+const extendedStatisticsSQLInheritedField = "null"
+const extendedStatisticsSQLpg15InheritedField = "se ->> 'inherited'"
+
+// pg_get_statisticsobjdef_expressions(s.oid) provides pg_stats_ext.exprs without accessing pg_stats_ext
+const extendedStatisticsSQL = `
+WITH relation_stats_ext AS (
+	SELECT se ->> 'statistics_schemaname' statistics_schemaname,
+		   se ->> 'statistics_name' statistics_name,
```
keiko713, line 20: Ugh, does anyone know why the tab being 8 spaces here? Asking for a friend.

input/postgres/relation_stats_ext.go
```diff
@@ -0,0 +1,96 @@
...
+	if StatsHelperExists(ctx, db, "get_relation_stats_ext") {
+		logger.PrintVerbose("Found pganalyze.get_relation_stats_ext() stats helper")
+		sourceTable = "pganalyze.get_relation_stats_ext() se"
+	} else {
+		if systemType != "heroku" && !connectedAsSuperUser(ctx, db, systemType) && globalCollectionOpts.TestRun {
+			logger.PrintInfo("Warning: Limited access to extended table statistics detected in database %s. Please set up"+
+				" the monitoring helper function pganalyze.get_relation_stats_ext (https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user)"+
+				" or connect as superuser, to get extended statistics for all tables.", dbName)
+		}
+		sourceTable = "(SELECT row_to_json(pse.*)::jsonb se FROM pg_catalog.pg_stats_ext pse) _"
```
keiko713, line 61: [note for reviewer] As we're returning jsonb with the helper function, we also need to return jsonb when we directly use `pg_stats_ext` here. `row_to_json(pse.*)::jsonb se` will allow me to specify the jsonb object using `se->>'something'`.

output/transform/postgres_relations.go
```diff
@@ -119,6 +119,27 @@ func transformPostgresRelations(s snapshot.FullSnapshot, newState state.Persiste
...
+			schemaStatsExtended, ok := schemaStats.RelationStatsExtended[relation.Oid]
+			if ok {
+				for _, s := range schemaStatsExtended {
+					sExtendedStats := snapshot.RelationInformation_ExtendedStatistic{
+						StatisticsSchema: s.StatisticsSchema,
+						StatisticsName:   s.StatisticsName,
+						Columns:          s.Columns,
+						Expressions:      s.Expressions,
+						Kind:             s.Kind,
+						HasData:          s.Inherited.Valid || s.NDistinct.Valid || s.Dependencies.Valid, // Only has valid data when either of these are valid
```
keiko713, line 133: [note for reviewer] `inherited` is non-null column so it would have been best to identify if we were able to obtain the actual stats data of this extended stats, but `s.Inherited` can be null for version 14 and below, so it can't be simply used. As a workaround, we check all three data values to see if we got any data here.
While things like `columns` above can be obtained without special permissions/privileges, `n_distinct` will only be available with the right permission/privilege (originally coming from [pg_statistic_ext_data](https://www.postgresql.org/docs/current/catalog-pg-statistic-ext-data.html)).

input/postgres/relation_stats_ext.go
```diff
@@ -0,0 +1,105 @@
...
+	"database/sql"
+	"fmt"
+
+	"github.com/lib/pq"
+	"github.com/pganalyze/collector/state"
+	"github.com/pganalyze/collector/util"
+)
+
+const extendedStatisticsSQLExprsField = "ARRAY[]::text[]"
+const extendedStatisticsSQLpg14ExprsField = "se ->> 'exprs'"
```
keiko713, line 14: [note for reviewer] Support for defining extended statistics on expressions started from Postgres 14, so it is okay to simply give an empty text array below Postgres 14.

input/postgres/relation_stats_ext.go
msakrejda: Another option is to also push that first CTE into the UDF, which would let you just use a simple table reference here:

```sql
WITH unwanted_pg_stats_ext_columns AS (
  SELECT
    array_agg(attname) AS cols
  FROM
    pg_attribute
  WHERE
    attrelid = 'pg_stats_ext'::regclass AND attname NOT IN ('statistics_schemaname','statistics_name','exprs','inherited','dependencies')            
),
pg_stats_ext_json AS (
  SELECT row_to_json(se.*)::jsonb - unwanted_pg_stats_ext_columns.cols
    FROM pg_catalog.pg_stats_ext se, unwanted_pg_stats_ext_columns
)
SELECT
  se ->> 'statistics_schemaname' AS statistics_schemaname,
  se ->> 'statistics_name' AS statistics_name,
  COALESCE(se ->> 'exprs', '[]') AS exprs,
  se ->> 'inherited' AS inherited,
  se ->> 'n_distinct' n_distinct,
  se ->> 'dependencies' dependencies
FROM
  pg_stats_ext_json
```

That also streamlines a bunch of the version-specific dynamic column handling in the CTE SQL (lines 13-16 in this file can be deleted and the stuff that depends on them can be simplified), so I think it's worth doing, but I don't feel super-strongly about it.

Note that I had to change `exprs` from your `ARRAY[]::text[]` to a plain `'[]'` since `se ->> 'exprs'` is a string in this context, but I think that still works.

I tested this locally on both 14 and 15 and it works fine.

msakrejda (approved):
Looks great, though I think we should harden the data fetching function a bit as discussed in the inline comment.

README.md
msakrejda: Separately, this could be marked `STABLE` instead of `VOLATILE`, no? https://www.postgresql.org/docs/current/sql-createfunction.html

keiko713: > Separately, this could be marked `STABLE` instead of `VOLATILE`, no?

Good point and I agree, I actually thought about this while creating this, though I went with `VOLATILE` as the other functions that we define here are all `VOLATILE`. However, I think all of these functions actually can be `STABLE`. Should we update them all?

> we could start querying sensitive data if a column similar to `most_common_vals` is introduced in a future Postgres version

I think this is definitely a valid point, though if I can wear a lazy(?) hat, allowlisting (your version) instead of having a blocklist (current PR) will make it harder to use any "good" new columns introduced as the user will need to update the function to support it too. For example, above `get_column_stats()` is using the allowlisting with `pg_catalog.pg_stats`. This view hasn't changed for a long time (last time it changed was 9.2). On the other hand, `pg_stats_ext` has changed twice recently, both adding useful columns as stats data and we actually collect these data with this PR. It's possible that some columns will be introduced in the future releases that we want to start collecting metrics from.
That said, this function is indeed focusing more on the security perspective, so being convenience (of the maintainer) over the security sounds wrong, so I'd say that we should go with Maciek's version.

Just wanted to loop @lfittl too to get +1 from him too.

input/postgres/relation_stats_ext.go
keiko713: Might be something to do with gofmt using tab with 8 spaces previously: https://github.com/golang/go/issues/19618
Unless someone objects, I'll keep it as is and go with 4 spaces tab here (since my screen is not so wide, 8 spaces is a bit of luxury).

input/postgres/relation_stats_ext.go
keiko713: Ugh so tl;dr `COALESCE(se ->> 'exprs', '[]') AS exprs` won't work, not because you are wrong, but because I was wrong to start with; I only tested with the case of the column being `ARRAY[]::text[]` (aka exprs null), and now I test with the case with something in exprs, it started raising errors `"exprs": pq: unable to parse array; expected '{' at offset 0` in golang side, likely because it can't convert it to `pq.Array`.

Now I'm very much in the 🐇 🕳️ of converting `jsonb` array to the text array and just started hating the world. I kinda got something with LATERAL join but not sure if this is good, looks ugly. Also tried something with a subquery in the SELECT clause (by using something similar to [the function](https://stackoverflow.com/questions/45243186/how-to-cast-json-array-to-text-array)), but that also keeps giving me errors here and there with null case.
I'll sleep on this for now but if you have any good ideas, I'm all ears.

README.md
lfittl: How about we write it like this, for any fields that are not on all versions? (i.e. `exprs` and `inherited`)

```sql
SELECT statistics_schemaname, statistics_name, dependencies, n_distinct, most_common_val_nulls, most_common_freqs, most_common_base_freqs,
       row_to_json(se.*)::jsonb ->> 'inherited' AS inherited,
       row_to_json(se.*)::jsonb ->> 'exprs' AS exprs
  FROM pg_catalog.pg_stats_ext se
```

(note I intentionally included n_distinct and the most_common_* fields except the text one, since they may prove useful in the future, and we already consider those safe in the context of pg_stats)

input/postgres/relation_stats_ext.go
```diff
@@ -0,0 +1,105 @@
...
+		   %s inherited,
+		   se ->> 'n_distinct' n_distinct,
+		   se ->> 'dependencies' dependencies
+	  FROM %s
+)
+SELECT c.oid,
+	   n.nspname,
+	   s.stxname,
+	   (SELECT array_agg(k) FROM unnest(s.stxkeys) k) stxkeys,
+	   sd.exprs,
```
lfittl, line 32: As discussed, we should change that back to `pg_get_statisticsobjdef_expressions(pg_stats_ext.exprs)` to not require extra permissions for getting the definition of the statistics object.

README.md
msakrejda: Works for me.

input/postgres/relation_stats_ext.go
msakrejda: Oh man, this is tricky. Starting from your LATERAL hint, I came up with:

```sql
WITH stats_ext AS (                                               
  SELECT
    row_number() over() AS row_id,
    statistics_schemaname,
    statistics_name,
    dependencies,
    n_distinct,
    most_common_val_nulls,
    most_common_freqs,
    most_common_base_freqs,
    row_to_json(se.*)::jsonb ->> 'inherited' AS inherited,
    row_to_json(se.*)::jsonb -> 'exprs' AS exprs
  FROM
    pg_stats_ext se
)
SELECT
  se1.statistics_schemaname,
  se1.statistics_name,
  se1.dependencies,
  se1.n_distinct,
  se1.most_common_val_nulls,
  se1.most_common_freqs,
  se1.most_common_base_freqs,
  se1.inherited,
  se_exprs.exprs
FROM stats_ext AS se1, LATERAL (
  SELECT
    coalesce(array_agg(exprs), array[]::text[]) AS exprs
  FROM (
    SELECT
      jsonb_array_elements_text(coalesce(se2.exprs, '[]')) AS exprs
    FROM
      stats_ext se2 WHERE se1.row_id = se2.row_id
  )
) AS se_exprs;
```

I had to add a `row_number()` window function to join these, since I'm not sure if there's a join key. I guess we could use (statistics_schemaname, statistics_name) instead? I guess that should work. This seems to do the right thing for me both when exprs is and is not available.

README.md
keiko713: Uhmm actually this will return the non defined table/type, so we will need to use `RETURNS TABLE` with a function definition which will make the function even longer 🤔

keiko713: I'll push the update too but here is what I got:

```sql
CREATE OR REPLACE FUNCTION pganalyze.get_relation_stats_ext() RETURNS TABLE(
  statistics_schemaname text, statistics_name text,
  inherited boolean, n_distinct pg_ndistinct, dependencies pg_dependencies,
  most_common_val_nulls boolean[], most_common_freqs float8[], most_common_base_freqs float8[]
) AS
$$
  /* pganalyze-collector */ SELECT statistics_schemaname, statistics_name,
  (row_to_json(se.*)::jsonb ->> 'inherited')::boolean AS inherited, n_distinct, dependencies, 
  most_common_val_nulls, most_common_freqs, most_common_base_freqs
  FROM pg_catalog.pg_stats_ext se;
$$ LANGUAGE sql VOLATILE SECURITY DEFINER;
```

input/postgres/relation_stats_ext.go
keiko713: Ah sorry I should have updated, but yeah this is going to be tricky but I'm actually going to use the Lukas's original way of using `pg_get_statisticsobjdef_expressions` function to obtain this instead of trying to deal with jsonb thingy. In this way, I don't need to get `exprs` from pg_stats_ext

keiko713: (Lukas kinda said it in https://github.com/pganalyze/collector/pull/476#discussion_r1454370154)

keiko713:
This is ready for re-review! 🙏

input/postgres/relation_stats_ext.go
msakrejda: Ah got it. I saw the comment but didn't put together that it meant we don't need to handle `exprs` like this.

input/postgres/relation_stats_ext.go
```diff
@@ -0,0 +1,96 @@
...
+		inheritedField = extendedStatisticsSQLpg15InheritedField
+	} else {
+		inheritedField = extendedStatisticsSQLInheritedField
+	}
+
+	if StatsHelperExists(ctx, db, "get_relation_stats_ext") {
+		logger.PrintVerbose("Found pganalyze.get_relation_stats_ext() stats helper")
+		sourceTable = "pganalyze.get_relation_stats_ext()"
+	} else {
+		if systemType != "heroku" && !connectedAsSuperUser(ctx, db, systemType) && globalCollectionOpts.TestRun {
```
msakrejda, line 56: I was wondering if we should still warn if you don't actually _have_ any extended statistics, but I think we should, since they often come in handy sooner or later, and it's a good opportunity to have this set up sooner rather than later.

msakrejda (approved):
Looks good :+1: .

README.md
lfittl: > Good point and I agree, I actually thought about this while creating this, though I went with VOLATILE as the other functions that we define here are all VOLATILE. However, I think all of these functions actually can be STABLE. Should we update them all?

I was trying to recall why we mark these as volatile, but came up empty - I thought that its maybe related to them being SECURITY DEFINER, but reviewing the Postgres documentation there doesn't appear to be any reason why a security definer function needs to be marked as volatile.

lfittl (commented):
:+1: (thanks for completing the work on the PR @keiko713!)

README.md
keiko713: Ah yeah thanks for following up re: VOLATILE and STABLE. I think it's still the best to double check if changing this to `STABLE` works without problems, but I'd say that let's address this separately. I also have been thinking if we should update our guided setup script, I think they don't include lots of functions defined here. Well, let's also address that separately.

PR #477, opened 2023-11-21
Allow pg_stat_statements failures and continue snapshot processing

lfittl:
Some situations can cause pg_stat_statements data collection to fail, e.g. a timeout when the query text file got too large. Previously we treated the whole snapshot as having failed in such situations, only reporting an error snapshot to pganalyze, without any statistics.

This caused two issues:

(1) Unrelated data, such as schema statistics, got lost in the face of
    temporary pg_stat_statements issues

(2) The early return from the collection logic caused us to skip over
    the optional pg_stat_statements_reset() facility in the collector,
    which otherwise would have allowed the collector to recover from
    an overly large file (by resetting it, allowing the next collection
    to succeed)

Fix by treating pg_stat_statements errors as a collector error in the snapshot, but continuing afterwards instead of returning early.

### DIFF
--- a/input/full.go
+++ b/input/full.go
@@ -50,8 +50,18 @@ func CollectFull(ctx context.Context, server *state.Server, connection *sql.DB, 
 	}
 	ts.Statements, ts.StatementTexts, ps.StatementStats, err = postgres.GetStatements(ctx, server, logger, connection, globalCollectionOpts, ts.Version, true, systemType)
 	if err != nil {
-		err = fmt.Errorf("Error collecting pg_stat_statements: %s", err)
-		return
+		// Despite query performance data being an essential part of pganalyze, there are
+		// situations where it may not be available (or it timed out), so treat it as a
+		// non-fatal error, and continue snapshot collection.
+		//
+		// Importantly this also make sure that we may execute a pg_stat_statements_reset
+		// (if configured) despite pg_stat_statements data retrieval failing, allowing
+		// recovery from situations where the query text file got too large.
+		//
+		// Note that we do log it as an error, which gets added automatically to the snapshot's
+		// CollectorErrors information.
+		logger.PrintError("Error collecting pg_stat_statements: %s", err)
+		err = nil
 	}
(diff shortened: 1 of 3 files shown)


input/full.go
```diff
@@ -64,15 +81,13 @@ func CollectFull(ctx context.Context, server *state.Server, connection *sql.DB,
 		ps.StatementResetCounter = 0
 		err = postgres.ResetStatements(ctx, logger, connection, systemType)
 		if err != nil {
-			// This is a non-fatal error, so continue snapshot collection but do log it as an error
-			// (this gets automatically added to the snapshot's CollectorErrors information)
 			logger.PrintError("Error calling pg_stat_statements_reset() as requested: %s", err)
 			err = nil
 		} else {
 			_, _, ts.ResetStatementStats, err = postgres.GetStatements(ctx, server, logger, connection, globalCollectionOpts, ts.Version, false, systemType)
```
keiko713, line 87: Maybe add a log for successful pg_stat_statements_reset() run here

input/full.go
```diff
@@ -64,15 +81,13 @@ func CollectFull(ctx context.Context, server *state.Server, connection *sql.DB,
...
 			if err != nil {
-				err = fmt.Errorf("Error collecting pg_stat_statements: %s", err)
-				return
+				logger.PrintError("Error collecting pg_stat_statements: %s", err)
```
keiko713, line 89: I wonder if it would be nice to explicitly mention "after reset". This error is the same message as the before-reset one but I'd say that this has a bit different nature

keiko713 (approved):
Several comments but overall looks good, thanks for working on this 👍
I am a bit worried if this won't create any bad full snapshot though (by missing some data). Unless you are very confident, it might be good to do some test

lfittl:
> I am a bit worried if this won't create any bad full snapshot though (by missing some data). Unless you are very confident, it might be good to do some test

FWIW, since we already had this behavior in case of shared_preload_libraries missing pg_stat_statements, I feel confident that this doesn't cause any unexpected problems that we wouldn't have encountered already (worst case it makes them more likely).

PR #480, opened 2023-11-28
Expand support of Crunchy Bridge system metrics

keiko713:
Currently, Crunchy Bridge system metrics is using the same logic as self-hosted one, using gopsutil to grab the system metrics of its own host.
This mostly works well, however, it doesn't provide an accurate metrics for the "disk partitions", aka storage information. Instead of providing the metric of database disk, it returns the metric of the container that is running the collector. This results the disk size to be wrongly reported as the total 2GB (which is the size of the Container Apps's container).

With this PR, we still use the gopsutil for most of the metrics, but use Crunchy Bridge API to obtain the disk metrics as well as some basic information of the cluster, such as region, plan, etc..


input/system/crunchy_bridge/client.go
```diff
@@ -0,0 +1,258 @@
...
+	Points  []MetricPoint `json:"points"`
+	Title   string        `json:"title"`
+}
+
+type MetricPoint struct {
+	Time  string  `json:"time"`
+	Value float64 `json:"value"`
+}
+
+type CPUMetrics struct {
```
keiko713, line 50: [note for reviewer] Technically we don't need types (as well as corresponding functions) from CPUMetrics to LoadAverageMetrics as we use gopsutil for that, not API.
I initially implemented it assuming that we had to use the API for them as we were not able to get correct metrics of them using gopsutil, but I was wrong and we can actually get accurate metrics using gopsutil.
I could simply delete them but I figured that it won't hurt to have them here.

input/system/crunchy_bridge/client.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+	}
+
+	return &metrics, err
+}
+
+func average(points []MetricPoint) float64 {
+	// With metric-views endpoint, it returns metrics for the last 15 minutes
+	// The latest data point(s) often returns value 0 as there is some lag within the metrics collection on Crunchy side
+	// When calculating average, ignore last 3 data points
```
keiko713, line 248: [note for reviewer] Hope this makes sense, but here is some example API response:

```
{
  "name": "disk-usage",
  "series": [
    {
      "is_empty": false,
      "max": null,
      "min": 0,
      "name": "postgres_databases_size_bytes",
      "points": [
        {
          "time": "2023-11-28T07:56:04Z",
          "value": 30542436
        },
        {
          "time": "2023-11-28T07:57:04Z",
          "value": 30542436
        },
        {
          "time": "2023-11-28T07:58:04Z",
          "value": 30542436
        },
        {
          "time": "2023-11-28T07:59:04Z",
          "value": 30542436
        },
        {
          "time": "2023-11-28T08:00:04Z",
          "value": 30542436
        },
        {
          "time": "2023-11-28T08:01:04Z",
          "value": 30542436
        },
        {
          "time": "2023-11-28T08:02:04Z",
          "value": 30542436
        },
        {
          "time": "2023-11-28T08:03:04Z",
          "value": 30542436
        },
        {
          "time": "2023-11-28T08:04:04Z",
          "value": 30542436
        },
        {
          "time": "2023-11-28T08:05:04Z",
          "value": 30542436
        },
        {
          "time": "2023-11-28T08:06:04Z",
          "value": 30542436
        },
        {
          "time": "2023-11-28T08:07:04Z",
          "value": 30542436
        },
        {
          "time": "2023-11-28T08:08:04Z",
          "value": 30542436
        },
        {
          "time": "2023-11-28T08:09:04Z",
          "value": 30542436
        },
        {
          "time": "2023-11-28T08:10:04Z",
          "value": 0
        }
      ],
      "title": "Size of Postgres databases",
      "unit": "byte"
    },
    ...
  ]
}
```

I picked up 3 as a random number (also I didn't really see more than 3 zeros), happy to tweak if others have some opinions.

input/system/crunchy_bridge/system.go
```diff
@@ -0,0 +1,62 @@
...
+
+	diskUsageMetrics, err := client.GetDiskUsageMetrics()
+	if err != nil {
+		logger.PrintError("CrunchyBridge/System: Encountered error when getting cluster disk usage metrics %v\n", err)
+		return
+	}
+
+	system.DataDirectoryPartition = "/"
+	system.DiskPartitions = make(state.DiskPartitionMap)
+	// Manually specify the disk name to "md0" as that's the main disk
```
keiko713, line 52: [note for reviewer] We pass all the disk stats to pganalyze, and on the pganalyze side, we only persist the disk stat of the disk that matches to the data directory partition.
In this case, with `system.DataDirectoryPartition = "/"`, this is the data dir with the disk name `"md0"`. Within transformation, we then assign the index of `"md0"` disk as `DiskIdx`.
https://github.com/pganalyze/collector/blob/def695671efc4670f80a1867238177bbfcc36f60/output/transform/system.go#L278

Crunchy Bridge instance has several disks, and if we don't specify this part well, `0` will be used as `DiskIdx` and we'll incorrectly get the disk stats of `loop0` (which is Container Apps disk).

```
[13] pry[stag](main)> full.system.disk_references
=> [<Pganalyze::Collector::DiskReference: device_name: "loop0">,
 <Pganalyze::Collector::DiskReference: device_name: "md0">,
 <Pganalyze::Collector::DiskReference: device_name: "nvme0n1">,
 <Pganalyze::Collector::DiskReference: device_name: "nvme0n1p1">,
 <Pganalyze::Collector::DiskReference: device_name: "nvme0n1p2">,
 <Pganalyze::Collector::DiskReference: device_name: "nvme1n1">,
 <Pganalyze::Collector::DiskReference: device_name: "nvme2n1">,
 <Pganalyze::Collector::DiskReference: device_name: "nvme3n1">,
 <Pganalyze::Collector::DiskReference: device_name: "nvme4n1">]
 ```

File: config/read.go [hunk 4]
```diff
@@ -221,6 +221,9 @@ func getDefaultConfig() *ServerConfig {
 	if gcpProjectID := os.Getenv("GCP_PROJECT_ID"); gcpProjectID != "" {
 		config.GcpProjectID = gcpProjectID
 	}
+	if crunchyBridgeAPIKey := os.Getenv("CRUNCHY_BRIDGE_API_KEY"); crunchyBridgeAPIKey != "" {
+		config.CrunchyBridgeAPIKey = crunchyBridgeAPIKey
+	}
```
lfittl, line 226: We should probably also add this for setting `crunchy_bridge_cluster_id` through an environment variable, looks like I missed that at the time.

input/system/crunchy_bridge/client.go
lfittl: Hmm. Couldn't we just check if the value is 0, and then ignore it?

Its essentially impossible for a database to have 0 size once its initialized, so that would make it easy to ignore those data points.

Not a fan of using a hard coded number like "ignore the last 3", since that could break if Crunchy changes their API.

input/system/crunchy_bridge/system.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/pganalyze/collector/input/system/selfhosted"
+	"github.com/pganalyze/collector/state"
+	"github.com/pganalyze/collector/util"
+)
+
+// GetSystemState - Gets system information about a Crunchy Bridge instance
+func GetSystemState(config config.ServerConfig, logger *util.Logger) (system state.SystemState) {
+	// With Crunchy Bridge, we are assuming that the collector is deployed on the Container Apps
+	// Most of the metrics can be obtained using the same way as the selfhosted
+	// (as the collector runs on the database server), except disk usage metrics
```
lfittl, line 16:
```suggestion
	// With Crunchy Bridge, we are assuming that the collector is deployed on Container Apps,
	// which run directly on the database server. Most of the metrics can be obtained
	// using the same way as a self hosted server, since the container receives a bind mount
	// of /proc and /sys from the host. Note this excludes disk usage metrics, which we instead
	// get from the API.
```

input/system/crunchy_bridge/system.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// When API key is provided, use API to obtain extra info including disk usage metrics
+	if config.CrunchyBridgeAPIKey == "" {
+		return
+	}
+	client := Client{Client: *config.HTTPClientWithRetry, BaseURL: apiBaseURL, BearerToken: config.CrunchyBridgeAPIKey, ClusterID: config.CrunchyBridgeClusterID}
+
+	clusterInfo, err := client.GetClusterInfo()
+	if err != nil {
+		logger.PrintError("CrunchyBridge/System: Encountered error when getting a cluster info %v\n", err)
```
lfittl, line 28:
```suggestion
		logger.PrintError("CrunchyBridge/System: Encountered error when getting cluster info %v\n", err)
```

lfittl (approved):
Looks good - a few small comments, but I think we can ship this!

input/system/crunchy_bridge/client.go
keiko713: > Couldn't we just check if the value is 0, and then ignore it?

Uhmm yes and no for this. Definitely yes for the disk size, so good call on that one. However, I've seen that the load average and cpu metrics being zero in the middle (not at the last) of the metrics, so simply ignoring zero is not "accurate" for _all_ metrics.

That said, after giving a thought of pros/cons, I think it's okay to simplify this to ignore zero.
If load average or cpu load is being zero on some data points, I think it's quite safe to assume that the other data points are also close to zero, so calculating average ignoring zero won't change the average that much. We won't have the accurate data even if we do the drop 3 data points way (or, there is also no definition of "accurate" here as we get last 15 mins metrics to show the 10 mins metrics).

PR #481, opened 2023-11-29
Compact snapshot stats: Fix incorrect counting for current snapshot

lfittl:
Due to the way the code was written, we would either count the current snapshot, or emit statistics - but even if we emit statistics we still need to increment the snapshot count.

### DIFF
--- a/output/compact.go
+++ b/output/compact.go
@@ -142,26 +142,27 @@ func submitCompactSnapshot(ctx context.Context, server *state.Server, collection
 		if server.CompactLogTime.IsZero() {
 			server.CompactLogTime = time.Now().Truncate(time.Minute)
 			server.CompactLogStats = make(map[string]uint8)
-		} else if time.Now().Sub(server.CompactLogTime) > time.Minute {
-			var keys []string
-			for k := range server.CompactLogStats {
-				keys = append(keys, k)
-			}
-			sort.Strings(keys)
-			details := ""
-			for i, kind := range keys {
-				details += fmt.Sprintf("%d %s", server.CompactLogStats[kind], kind)
-				if i < len(keys)-1 {
-					details += ", "
-				}
-			}
-			if len(details) > 0 {
-				logger.PrintInfo("Compact snapshots submitted: " + details)
-			}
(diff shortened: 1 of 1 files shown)


output/compact.go
```diff
@@ -142,26 +142,27 @@ func submitCompactSnapshot(ctx context.Context, server *state.Server, collection
...
+				sort.Strings(keys)
+				details := ""
+				for i, kind := range keys {
+					details += fmt.Sprintf("%d %s", server.CompactLogStats[kind], kind)
+					if i < len(keys)-1 {
+						details += ", "
+					}
+				}
+				if len(details) > 0 {
+					logger.PrintInfo("Compact snapshots submitted: " + details)
```
lfittl, line 161: Whilst looking at this I also considered whether we should reword this to "Submitted compact snapshots" (to be more consistent with the previous wording, and what's used for full snapshots), but I suspect this was intentionally worded this way in #467 since we're talking about what happened in the last minute (vs what just happened), so I think this is okay to leave worded as is.

cc @msakrejda in case you have any wording thoughts on this message.

for context, right now it is:

Compact snapshots submitted: 5 activity, 3 logs

and I'm wondering if it shouldn't be:

Submitted compact snapshots: 5 activity, 3 logs

keiko713: Here is how it looks like currently, so I can see that a previous wording matches to the full snapshot and might be easier to also scan through.
Alternatively, we could also update the "Submitted snapshot successfully" and make it "Full snapshot submitted", though it doesn't look good 🤔 

```
 I [default] Compact snapshots submitted: 5 activity, 2 logs
 I [default] Compact snapshots submitted: 5 activity, 2 logs
 I [default] Submitted snapshot successfully
 I [default] Compact snapshots submitted: 5 activity, 2 logs
 ```

Comment (lfittl) [reply output/compact.go:161]: If we were to change it, I'm thinking this would be most sensible:

```
 I [default] Submitted compact snapshots successfully: 5 activity, 2 logs
 I [default] Submitted full snapshot successfully
```

Let me make a separate PR for this, and we can discuss there.

lfittl: See https://github.com/pganalyze/collector/pull/482

PR #482, opened 2023-11-29
Snapshot submission: Streamling wording for success message

lfittl:
This adjusts the recently changed compact snapshot information to be closer to what it was before, and adds "full" to the full snapshot message for clarity:

```
 I [default] Submitted compact snapshots successfully: 5 activity, 2 logs
 I [default] Submitted full snapshot successfully
```

Note that the compact snapshot message continues to emitted once per minute, not continuously as in past releases.

### DIFF
--- a/output/compact.go
+++ b/output/compact.go
@@ -158,7 +158,7 @@ func submitCompactSnapshot(ctx context.Context, server *state.Server, collection
 					}
 				}
 				if len(details) > 0 {
-					logger.PrintInfo("Compact snapshots submitted: " + details)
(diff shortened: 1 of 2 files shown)


keiko713 (approved):
I'm somewhat indifferent about this, so might be nice to have +1 from someone who has more opinion

lfittl:
Since there are no strong opinions either way, I'll go ahead and merge this - I think it feels slightly better to me to be consistent here (and I've wished for a better way to search for "full snapshot" in the logs previously, so that helps with that too).

PR #483, opened 2023-11-30
Release 0.52.3

keiko713:
Blocked by https://github.com/pganalyze/collector/pull/484 (we should make a release after this gets merged)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,27 @@
 # Changelog
 
+## 0.52.3      2023-11-30
+
+* Update the snapshot submission log format
+  - Previously, compact (activity and log) snapshots submission log were emitted
+    every time it's completed. This was causing lots of log lines as an activity
+    snapshot submission happens every 10 seconds
+  - With this update, we group the log and only emit once per minute
+* OpenTelemetry integration:
+  - Support `pganalyze` tracestate to set start time of the span
+  - Start time can be specified with `t` member key, the Unix time in second,
+    allowing decimals to present up to nano seconds
+  - This allows to specify a better span start and end time in case the accurate
+    time is unable to be obtained through the log, like RDS
+* Allow pg_stat_statements failures and continue snapshot processing
+  - Previously, when pg_stat_statements data collection was failing (e.g. a
+    timeout when the query text file got too large), the whole snapshot was
+    treated as failed and only reporting an error snapshot to pganalyze, without
+    any statistics
+  - Instead, treating pg_stat_statements errors as a collector error in the
+    snapshot, but continuing afterwards and report other statistics that were
+    collected successfully
+
 ## 0.52.2      2023-10-26
 
 * OpenTelemetry integration:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,15 @@
 
 ## 0.52.3      2023-11-30
 
-* Reduce frequency of collector log generation
-  - Previously, real-time snapshots would generate log lines every 10 seconds,
-    which made errors hard to find
+* Collector log output: Reduce frequency of some snapshot log events
+  - Previously, near real-time "compact" snapshots would generate log lines
+    every 10 seconds, which made errors hard to find
   - Now, a single log line is printed once a minute with a summary of snapshots
     submitted
   - Note that `--verbose` will still log every snapshot as it's submitted
+* Collector log output: Add "full" prefix for full snapshots sent every 10 minutes
+  - This changes the "Submitted snapshot successfully" message to read
+    "Submitted full snapshot successfully" instead
 * OpenTelemetry integration:
   - Support `pganalyze` tracestate to set start time of the span
   - Start time can be specified with `t` member key as Unix time in seconds,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,12 @@
 
 ## 0.52.3      2023-11-30
 
-* Update the snapshot submission log format
-  - Previously, compact (activity and log) snapshots submission log were emitted
-    every time it's completed. This was causing lots of log lines as an activity
-    snapshot submission happens every 10 seconds
-  - With this update, we group the log and only emit once per minute
+* Reduce frequency of collector log generation
+  - Previously, real-time snapshots would generate log lines every 10 seconds,
+    which made errors hard to find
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,27 @@
 # Changelog
 
+## 0.52.3      2023-11-30
+
+* Update the snapshot submission log format
+  - Previously, compact (activity and log) snapshots submission log were emitted
+    every time it's completed. This was causing lots of log lines as an activity
+    snapshot submission happens every 10 seconds
+  - With this update, we group the log and only emit once per minute
```
seanlinsley, line 9: Another option, in an attempt to make the intent of the change clear and to generalize activity/log/compact snapshots to "real-time".

* Reduce frequency of collector log generation
  - Previously, real-time snapshots would generate log lines every 10 seconds, which made errors hard to find.
  - Now, a single log line is printed once a minute with a summary of snapshots submitted.
  - Note that `--verbose` will still log every snapshot as it's submitted.

keiko713: Nice, thanks for the suggestion. Looks much better, I'll use it 👍

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+* Reduce frequency of collector log generation
+  - Previously, real-time snapshots would generate log lines every 10 seconds,
+    which made errors hard to find
+  - Now, a single log line is printed once a minute with a summary of snapshots
+    submitted
+  - Note that `--verbose` will still log every snapshot as it's submitted
+* OpenTelemetry integration:
+  - Support `pganalyze` tracestate to set start time of the span
+  - Start time can be specified with `t` member key, the Unix time in second,
+    allowing decimals to present up to nano seconds
```
msakrejda, line 14: I think the current wording is a little unclear; here is one alternative suggestion:
```suggestion
  - Start time can be specified with `t` member key as Unix time in seconds,
    with decimals to specify precision down to nano seconds
```

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+  - This allows to specify a better span start and end time in case the accurate
+    time is unable to be obtained through the log, like RDS
```
msakrejda, line 16: A couple of minor wording suggestions:
 - since it's the first time using it in this section, I think we should qualify RDS as "Amazon RDS" just to make it clear
 - it's precision, not accuracy, that's the issue with RDS (this feature could also address accuracy problems to some extent, but the driver is the RDS precision problems)
 - make it clearer that, by default, we're looking at log line timestamps in the logs to establish trace timings
```suggestion
  - This allows specifying a better span start and end time in case precise timestamps
    are not present in the Postgres logs, like with Amazon RDS
```
you could also add "or the log timestamps are not accurate" to point out that use case as well, but I think that's niche enough and obvious enough that it's not worth calling out here.

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+* Allow pg_stat_statements failures and continue snapshot processing
+  - Previously, when pg_stat_statements data collection was failing (e.g. a
+    timeout when the query text file got too large), the whole snapshot was
+    treated as failed and only reporting an error snapshot to pganalyze, without
+    any statistics
```
msakrejda, line 21:
```suggestion
  - Previously, when pg_stat_statements data collection failed (e.g. a
    timeout when the query text file got too large), the whole snapshot was
    treated as failed and only reported an error snapshot to pganalyze, without
    any statistics
```

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+  - Instead, treating pg_stat_statements errors as a collector error in the
+    snapshot, but continuing afterwards and report other statistics that were
+    collected successfully
```
msakrejda, line 24:
```suggestion
  - Instead, treat pg_stat_statements errors as a collector error in the
    snapshot, but continue afterwards and report other statistics that were
    collected successfully
```

msakrejda (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
 # Changelog
 
+## 0.52.3      2023-11-30
+
+* Reduce frequency of collector log generation
+  - Previously, real-time snapshots would generate log lines every 10 seconds,
+    which made errors hard to find
+  - Now, a single log line is printed once a minute with a summary of snapshots
+    submitted
+  - Note that `--verbose` will still log every snapshot as it's submitted
```
lfittl, line 10: I realize that I made a related change in #482 that's probably worth mentioning, in case anyone has regular expressions on the logs set up:

```
* Collector log output: Add "full" prefix for full snapshots sent every 10 minutes
  - This changes the "Submitted snapshot successfully" message to read "Submitted full snapshot successfully" instead
```

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
 # Changelog
 
+## 0.52.3      2023-11-30
+
+* Reduce frequency of collector log generation
```
lfittl, line 5: For consistency with the other message I proposed below, maybe reword this:

```suggestion
* Collector log output: Reduce frequency of some snapshot log events
```

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+  - Previously, real-time snapshots would generate log lines every 10 seconds,
```
lfittl, line 6: I realize this contradicts @seanlinsley's feedback, but I'm not a fan of introducing a new term here that is different from what we document (e.g. on https://pganalyze.com/docs/collector#data-collection-and-snapshot-submission-frequency).

Maybe a compromise for now could be to use the term "near real-time" (which we do use in-app on some check pages), and specifically reference the "compact" term:

```suggestion
  - Previously, near real-time "compact" snapshots would generate log lines every 10 seconds,
```

lfittl (approved):

CHANGELOG.md
keiko713: Nice, I think it's good combination of near real-time + compact (and I was mildly panicking that the doc is not having updated format of the log while I was 100% sure Maciek already fixed it, but I just realized that we just haven't shipped it yet).

benoittgt:
Is it possible have the docker image pushed? :)

lfittl:
> Just realized that I forgot to add things about Crunchy to the changelog 🤦

Good catch - I think we can just revise this with another PR.

PR #484, opened 2023-12-01
Support tracestate query tag with marginalia shape

keiko713:
While I was testing, I realized that tracestate is not propagating well when it's passed with marginalia shape.
This PR should fix it.
Honestly, I'm not 100% sure if "With sqlcommenter format (key='value'), key shouldn't include ":"" is accurate, but I think it's a safe assumption to make.

With sqlcommenter, the tracestate might be URL escaped, but let's assume that it's not for marginalia here.

### DIFF
--- a/logs/querysample/tags.go
+++ b/logs/querysample/tags.go
@@ -55,8 +55,8 @@ func parseTags(query string) map[string]string {
 
 func sqlcommenterFormat(str string) bool {
 	keyAndValue := strings.SplitN(str, "=", 2)
-        // With sqlcommenter format (key='value'), key shouldn't include ":"
-        return len(keyAndValue) == 2 && !strings.Contains(keyAndValue[0], ":")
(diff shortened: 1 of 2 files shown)


logs/querysample/tags.go
```diff
@@ -53,6 +53,17 @@ func parseTags(query string) map[string]string {
...
 
+func sqlcommenterFormat(str string) bool {
+	keyAndValue := strings.SplitN(str, "=", 2)
+	if len(keyAndValue) == 2 {
+		// With sqlcommenter format (key='value'), key shouldn't include ":"
+		if !strings.Contains(keyAndValue[0], ":") {
+			return true
+		}
+	}
+	return false
```
msakrejda, line 64: NBD, but this could be simplified to
```suggestion
        // With sqlcommenter format (key='value'), key shouldn't include ":"
        return len(keyAndValue) == 2 && !strings.Contains(keyAndValue[0], ":")
```

PR #486, opened 2023-12-07
Fall back to managed identity with workload identity credential error

keiko713:
The current logic looks like falling back to the `NewManagedIdentityCredential` when `NewWorkloadIdentityCredential` isn't created, but actually it's not working like that way. Instead, it's erroring out when `NewWorkloadIdentityCredential` failed to be created, making it impossible for the VM deployed collector to try out managed identity credential (workload identity credential is used for AKS setup).

With this change, avoid raising an error and early return when `NewWorkloadIdentityCredential` returns an error. Instead, try `NewManagedIdentityCredential` then raise an error when both of them failed.

For the reference, if `AZURE_AD_CLIENT_ID` is not specified and multiple identities are assigned to the VM, you'd see the following error:

```
E [default] ERROR - Could not get logs through Azure Event Hub: failed to connect to the Event Hub management node: ChainedTokenCredential: failed to acquire a token.
Attempted credentials:
        ManagedIdentityCredential: no default identity is assigned to this resource
```

### Reference

https://github.com/Azure/azure-sdk-for-go/blob/837ae161d16716d778dd414fbf963f5756c9d5cc/sdk/azidentity/default_azure_credential.go#L62
https://pganalyze.com/docs/collector/settings#azure-settings
https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash

### DIFF
--- a/input/system/azure/logs.go
+++ b/input/system/azure/logs.go
@@ -77,24 +77,7 @@ func getEventHubConsumerClient(config config.ServerConfig) (*azeventhubs.Consume
 			return nil, fmt.Errorf("failed to set up client secret Azure credentials: %s", err)
 		}
 	} else {
-		workloadIdentityCredential, err := azidentity.NewWorkloadIdentityCredential(nil)
-		if err != nil {
-			return nil, fmt.Errorf("failed to set up workload identity Azure credentials: %s", err)
-		}
-		var managedIdentityOptions *azidentity.ManagedIdentityCredentialOptions
-		if config.AzureADClientID != "" {
-			managedIdentityOptions = &azidentity.ManagedIdentityCredentialOptions{
-				ID: azidentity.ClientID(config.AzureADClientID),
-			}
-		}
-		managedIdentityCredential, err := azidentity.NewManagedIdentityCredential(managedIdentityOptions)
-		if err != nil {
-			return nil, fmt.Errorf("failed to set up managed identity Azure credentials: %s", err)
-		}
-		credential, err = azidentity.NewChainedTokenCredential([]azcore.TokenCredential{
-			workloadIdentityCredential,
-			managedIdentityCredential,
-		}, nil)
+		credential, err = azidentity.NewDefaultAzureCredential(nil)
 		if err != nil {
 			return nil, fmt.Errorf("failed to use default Azure credentials: %s", err)
 		}
--- a/input/system/azure/logs.go
+++ b/input/system/azure/logs.go
@@ -77,9 +77,33 @@ func getEventHubConsumerClient(config config.ServerConfig) (*azeventhubs.Consume
 			return nil, fmt.Errorf("failed to set up client secret Azure credentials: %s", err)
 		}
 	} else {
-		credential, err = azidentity.NewDefaultAzureCredential(nil)
-		if err != nil {
-			return nil, fmt.Errorf("failed to use default Azure credentials: %s", err)
+		var creds []azcore.TokenCredential
+		var errorMessages []string
+		workloadIdentityCredential, err := azidentity.NewWorkloadIdentityCredential(nil)
+		if err == nil {
+			creds = append(creds, workloadIdentityCredential)
+		} else {
+			errorMessages = append(errorMessages, "WorkloadIdentityCredential: "+err.Error())
+		}
+		var managedIdentityOptions *azidentity.ManagedIdentityCredentialOptions
+		if config.AzureADClientID != "" {
+			managedIdentityOptions = &azidentity.ManagedIdentityCredentialOptions{
+				ID: azidentity.ClientID(config.AzureADClientID),
+			}
+		}
+		managedIdentityCredential, err := azidentity.NewManagedIdentityCredential(managedIdentityOptions)
+		if err == nil {
+			creds = append(creds, managedIdentityCredential)
+		} else {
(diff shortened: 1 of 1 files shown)


lfittl:
> The current logic looks like falling back to the `NewManagedIdentityCredential` when `NewWorkloadIdentityCredential` isn't created, but actually it's not working like that way. Instead, it's erroring out when `NewWorkloadIdentityCredential` failed to be created, making it impossible for the VM deployed collector to try out managed identity credential (workload identity credential is used for AKS setup).
> 
> Let's use [`azidentity.NewDefaultAzureCredential`](https://github.com/Azure/azure-sdk-for-go/blob/837ae161d16716d778dd414fbf963f5756c9d5cc/sdk/azidentity/default_azure_credential.go#L62) instead for the manually creating a chained credential, as it does what we need it to do well.
> 
> I think we needed to do this manually because in our doc, [we tell to use `AZURE_AD_CLIENT_ID` to specify the client ID when multiple managed identities are assigned to the VM](https://pganalyze.com/docs/collector/settings#azure-settings). However, instead of trying to do this manually, I propose that we update the document to say that they should specify it as `AZURE_CLIENT_ID`, which the `azidentity.NewDefaultAzureCredential` will take care of it well. Docs PR: [pganalyze/pganalyze-docs#205](https://github.com/pganalyze/pganalyze-docs/pull/205)

I'm not opposed to changing from AZURE_AD_CLIENT_ID to AZURE_CLIENT_ID, especially since Microsoft recently renamed Azure AD to Microsoft Entra ID (and so using an "AZURE_AD" prefix doesn't really make sense).

However, I don't think we should rely on the default credentials logic for two reasons:

1. It unnecessarily breaks backwards compatibility for anyone who manually set AZURE_AD_CLIENT_ID before
2. It doesn't support setting the Client ID through the configuration file, which one may want to use when using one Azure VM that connects out to different servers to be monitored, with different client IDs

If I understand correctly the main problem here is that we don't do what the defaults logic is doing, which is to gather the errors and then return them together, instead of stopping early on an error. I'd be in favor of fixing just that bug instead.

For reference, we do something similar in the AWS case, where the SDK provides a defaults logic, but we instead manually initialize it, so we take all the configuration from the pganalyze config structs (https://github.com/pganalyze/collector/blob/main/util/awsutil/amazon.go#L46).

keiko713:
I'm gonna go with only fixing the bug and not changing the env var name. I think changing the env var name should occur with a bit more bigger change as it will break things and we also want to have more warnings printed out and etc., which would be good to do as a separate PR if we want to do so in the future (which does make sense to do at some point, but doesn't need to be now).

----

> I'm not opposed to changing from AZURE_AD_CLIENT_ID to AZURE_CLIENT_ID, especially since Microsoft recently renamed Azure AD to Microsoft Entra ID (and so using an "AZURE_AD" prefix doesn't really make sense).
> It unnecessarily breaks backwards compatibility for anyone who manually set AZURE_AD_CLIENT_ID before

I wanted to clarify a bit for why I was proposing to use `AZURE_CLIENT_ID` (though it's not relevant anymore as I'm not going to do this): I wanted to introduce specifying `AZURE_CLIENT_ID` for managed identity + VM, but keeping `AZURE_AD_CLIENT_ID` for other cases. I specifically wanted to do this to support the backwards compatibility, as the current collector always fails with the managed identity + VM setup, meaning specifying `AZURE_AD_CLIENT_ID` doesn't work with these people, therefore changing how to specify the client ID won't break the backwards compatibility -- was my justification.
However, I think I was wrong on this. I thought the managed identity + VM setup was introduced with https://github.com/pganalyze/collector/pull/415 (which was merged in May), but I found that [this doc change](https://github.com/pganalyze/pganalyze-docs/pull/145/files) to mention that you can use `AZURE_AD_CLIENT_ID` to specify the Managed Identity's Client ID was made in April, so the collector before #415 was actually supporting the managed identity + VM + `AZURE_AD_CLIENT_ID` okay, so it does break backwards compatibility for these customers who are using the collector below 0.50.0. Also, your second point about it doesn't support setting the Client ID through the configuration file is totally correct.

input/system/azure/logs.go
```diff
@@ -88,15 +92,18 @@ func getEventHubConsumerClient(config config.ServerConfig) (*azeventhubs.Consume
...
+			errorMessages = append(errorMessages, "ManagedIdentityCredential: "+err.Error())
 		}
-		credential, err = azidentity.NewChainedTokenCredential([]azcore.TokenCredential{
-			workloadIdentityCredential,
-			managedIdentityCredential,
-		}, nil)
-		if err != nil {
-			return nil, fmt.Errorf("failed to use default Azure credentials: %s", err)
+		if len(creds) == 0 {
+			return nil, fmt.Errorf("failed to set up Azure credentials:\n\t" + strings.Join(errorMessages, "\n\t"))
```
keiko713, line 101: [note for reviewer] interestingly, even when I pass a bad client ID or not specifying client ID with multiple identities, it won't really fail here but will fail when actually trying to pull the log from the event hub. This is because this part is purely creating a cred, not really validating that cred. In other words, I don't know if this condition will ever meet, as managedIdentityCredential will always be created.
I think it's okay to leave it as is, but just wanted to leave a memo here.

keiko713:
@lfittl it's ready for another 👀 when you have time!

msakrejda (approved):

lfittl (approved):

PR #488, opened 2023-12-08
Avoid error collecting schema stats on Citus tables with no indexes

msakrejda:
The Citus `run_command_on_shards` function we use to collect
distributed index stats appears to return an empty string when the
query passed in returns zero rows. We currently try to cast that to
jsonb, which fails (this is not a valid jsonb value).

Ignore any empty strings returned by `run_command_on_shards`.


lfittl (approved):

lfittl:
I wonder if its worth modifying https://github.com/pganalyze/collector/blob/main/integration_test/citus-pgbench-ddl.sql to cover this test case? (I suspect it might be as simple as dropping an index on one of the pgbench tables before distributing them)

msakrejda:
@lfittl yeah, good call. I was just thinking about how to test this. We could alternately just create another table, right? I'll see if I can reproduce the failure first.

lfittl:
> @lfittl yeah, good call. I was just thinking about how to test this. We could alternately just create another table, right? I'll see if I can reproduce the failure first.

Yeah, good point creating another table (without indexes) and then distributing it should also show the error I imagine.

msakrejda:
Ok, the good news is that with the test fix in #489, I can reproduce the error on `main` with the test I just added, and I don't see the error on this branch. The bad news is that our test does not fail because of our error handling around this and the way the test is structured (the error is only visible in inspecting the test log output). I'll try to expand the test to actually catch this before I merge this.

msakrejda:
Alright, the updated test case fails on main and passes here.

PR #489, opened 2023-12-09
Fix Citus integration test

msakrejda:
The current test was running without the Citus extension, since we
were passing stdin to `docker exec`, to pass to psql, but without
running it in interactive mode. Docker discarded stdin, and the
test was running without Citus-specific functionality.

Pass the -i flag to `docker exec` so it relays stdin to psql.  Also
create the Citus extension (which is not created by default in the
Citus Docker image), and update the expected output file.


integration_test/citus.expected
```diff
@@ -1,7 +1,11 @@
 "INSERT INTO pgbench_history (tid, bid, aid, delta, mtime) VALUES ($1, $2, $3, $4, CURRENT_TIMESTAMP)"
 "SELECT abalance FROM pgbench_accounts WHERE aid = $1"
+"SELECT assign_distributed_transaction_id($1, $2, $3)"
+"SELECT count(*) AS count FROM public.pgbench_branches_102053 pgbench_branches WHERE $1"
```
msakrejda, line 4: It looks like this sharded table identifier is not stable; will filter it.

integration_test/citus-pgbench-ddl.sql
```diff
@@ -1,8 +1,9 @@
+CREATE EXTENSION citus;
 ALTER TABLE pgbench_accounts DROP CONSTRAINT pgbench_accounts_pkey;
 ALTER TABLE pgbench_accounts ADD PRIMARY KEY (aid, bid);
 ALTER TABLE pgbench_tellers DROP CONSTRAINT pgbench_tellers_pkey;
 ALTER TABLE pgbench_tellers ADD PRIMARY KEY (tid, bid);
 SELECT create_distributed_table('pgbench_accounts', 'bid');
```
lfittl, line 6: Curious that these statements didn't fail, since that would have required the Citus extension to exist?

lfittl: Ah, I missed the part that explains this in the PR description. Nevermind :)

msakrejda: Yeah it was basically https://youtu.be/aI0euMFAWF8?si=sHdBguJhj_aDwsVN

lfittl (approved):

Issue #328, opened 2022-10-20
Helm chart repo


jalaziz:
With #273, the helm chart was added to this repo. However, it's not usable without a helm repository to install from.

The possibility of hosting the repo on packages.pganalyze.com was discussed in the PR. Other options include Github Pages or even a dedicated Github repo.

Github also has [helm repo support on the roadmap](https://github.com/github/roadmap/issues/122) but it doesn't appear to be scheduled yet.

jalaziz:
Instead of hosting the helm chart repo via static site, Helm 3.8 supports OCI registries: https://helm.sh/docs/topics/registries/

This means that you should be able to use the GHCR, GCR, ECR, etc as the helm repo.

ViktorTverdun:
+1 to this. I was also looking for the helm chart that can be installed from repo, but so far need to copy the source.

PR #490, opened 2023-12-13
Add helm chart package support

keiko713:
Closes https://github.com/pganalyze/collector/issues/328

🗒️ I changed the direction quite a bit, so updating the PR summary here too.

With this PR, I added the following steps in the release CI:

* build_charts (new!): this creates a helm package of pganalyze-collector (e.g. pganalyze-collector-0.55.0.tgz)
* release (existing): added the section to upload created pganalyze-collector-0.55.0.tgz to the GitHub release
* release_charts (new!): this uploads a helm package, as well as update index.yaml file and update it to s3

In theory, CI will do the helm chart release for us so we don't need to do anything further.

I also did the following to set up charts.pganalyze.com:

* created s3 bucket and did the CNAME thingy, etc. to make sure the host works
* uploaded index.yaml, pganalyze-collector-0.55.0.tgz, and index.html manually to make sure that charts.pganalyze.com works

Other notes:

* looks like lots of companies are using https://github.com/helm/chart-releaser-action (e.g. datadog, new relic, etc.)
* above is convenient and does everything for us, we could consider adopting
  * with this, the package will be placed in GitHub release
  * the website is hosted using GitHub pages
  
I'm currently kinda doing these github action steps manually, but I can also remove lots things because our case is really simple.

<details>
  <summary>previous description</summary>
I admit that I don't have much knowledge around here so maybe I'm saying something pretty off, though at least publishing the chart worked (currently publishing as a private so it doesn't work for everyone yet).

Lukas said that we might be able to push this chart to the same place as the current collector docker image, like `quay.io/pganalyze/collector`. However, I'm not quite sure about this from several points, and I'd say that we might be okay to publish the helm chart under `quay.io/pganalyze/pganalyze-collector`:

* In quay.io UI, I don't really see how this can be supported. You can't `helm pull quay.io/pganalyze/collector` and also can't `docker pull quay.io/pganalyze/pganalyze-collector`, though I don't see how this can be coexist in the same "repo" and easily tell which one is in which state in the UI, so it might be safe to publish this separately
* when you run `helm package contrib/helm/pganalyze-collector`, it will automatically crate a package using the name `pganalyze-collector`, so if we want to publish the helm chart with `pganalyze/collector`, we need to change lots of place under `contrib/helm` to support it
</details>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -250,16 +250,6 @@ jobs:
         asset_name: pganalyze-collector_${{ steps.get_version.outputs.version }}_arm64.deb
         asset_content_type: application/octet-stream
 
-    - name: Upload helm package
-      uses: actions/upload-release-asset@v1
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      with:
-        upload_url: ${{ steps.create_release.outputs.upload_url }}
-        asset_path: ./pganalyze-collector-${{ steps.get_version.outputs.version }}.tgz
-        asset_name: pganalyze-collector-${{ steps.get_version.outputs.version }}.tgz
-        asset_content_type: application/octet-stream
-
   release_charts:
     needs: [release]
     runs-on: ubuntu-latest
@@ -286,7 +276,7 @@ jobs:
     - name: Configure AWS credentials
       uses: aws-actions/configure-aws-credentials@v4
       with:
-        role-to-assume: arn:aws:iam::123456789012:role/HelmChartPackageRole
+        role-to-assume: arn:aws:iam::793741702295:role/CollectorChartsPackage
         aws-region: us-east-1
 
     - name: Upload package to S3
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -119,8 +119,32 @@ jobs:
           packages/tmp/pganalyze-collector_*_arm64.deb
         if-no-files-found: error
 
+  build_charts:
+    runs-on: ubuntu-latest
+    # ensure we don't release on create events for branches (only tags)
+    if: ${{ startsWith( github.ref, 'refs/tags' ) }}
+
+    steps:
+    - name: Install Helm
+      uses: azure/setup-helm@v4
+
+    - name: Check out code
+      uses: actions/checkout@v3
+      with:
+        submodules: true
+
+    - name: Create helm package
+      run: helm package contrib/helm/pganalyze-collector
+
+    - name: Upload packages as artifacts
+      uses: actions/upload-artifact@v3
+      with:
+        name: packages
+        path: packages/tmp/pganalyze-collector-*.tgz
+        if-no-files-found: error
+
   release:
-    needs: [build, build_arm, build_packages]
+    needs: [build, build_arm, build_packages, build_charts]
     runs-on: ubuntu-latest
     # ensure we don't release on create events for branches (only tags)
     if: ${{ startsWith( github.ref, 'refs/tags' ) }}
@@ -218,4 +242,14 @@ jobs:
         upload_url: ${{ steps.create_release.outputs.upload_url }}
         asset_path: ./pganalyze-collector_${{ steps.get_version.outputs.version }}_arm64.deb
         asset_name: pganalyze-collector_${{ steps.get_version.outputs.version }}_arm64.deb
+        asset_content_type: application/octet-stream
+
+    - name: Upload helm package
+      uses: actions/upload-release-asset@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        upload_url: ${{ steps.create_release.outputs.upload_url }}
+        asset_path: ./pganalyze-collector-${{ steps.get_version.outputs.version }}.tgz
+        asset_name: pganalyze-collector-${{ steps.get_version.outputs.version }}.tgz
         asset_content_type: application/octet-stream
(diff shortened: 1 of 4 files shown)


keiko713:
Lukas and I had a chat about the naming of the helm chart. Currently, we're using `pganalyze-collector` with the repo name `pganalyze`, but this can be quite confusing.

* quay.io/pganalyze/collector => hosting the docker images of the collector
* quay.io/pganalyze/pganalyze-collector => hosting the helm chart

We talked about several options, but after researching more, I'm leaning towards to one option Lukas mentioned, which is creating a new repo for the helm charts:

* quay.io/pganalyze/collector => hosting the docker images of the collector
* quay.io/pganalyzecharts/collector => hosting the helm chart

This is somewhat similar to what bitnami is doing:

* https://hub.docker.com/u/bitnami => for hosting images
* https://hub.docker.com/u/bitnamicharts => for hosting charts

Example in Artifact Hub: https://artifacthub.io/packages/helm/bitnami/redis

Alternative idea is (which was mentioned during a chat too):

* quay.io/pganalyze/collector => hosting the docker images of the collector
* quay.io/pganalyze/collector-charts => hosting the helm chart

I think this works okay too, but adding the `charts` (or `chart`) in the chart name is not exactly cool (there are only a few helm charts having `charts` with the name, by searching Artifact Hub).
@lfittl thoughts? If you +1 for the `pganalyzecharts` org/repo (or, `pganalyze-charts`), can you please go ahead and create it?

keiko713:
chat with Lukas: we're going to actually push this to charts.pganalyze.com

README.md
```diff
@@ -320,6 +320,37 @@ The script will also have the following environment variables set:
...
 
 
+Helm Chart
+----------
+
+The pganalyze collector uses the OCI ([Open Container Initiative](https://opencontainers.org/)) registry to publish the Helm chart.
+You can install the Helm chart for the collector like the following:
+
+```bash
+helm install pga-collector oci://quay.io/pganalyze/pganalyze-collector --values=myvalues.yml
```
keiko713, line 330: @lfittl Ok coming back here to just close the loop about the naming of the chart. With the `helm repo` way, here is how the install process will look like with the current name "pganalyze-collector":

```
helm repo add pganalyze https://charts.pganalyze.com/
helm install pga-collector pganalyze/pganalyze-collector
```

It is repeating pganalyze twice, but I think it's okay, as other providers are repeating the name as well like `helm install newrelic-bundle newrelic/nri-bundle`, `helm install my-datadog-operator datadog/datadog-operator`. With that, I'll keep the name as is.

.github/workflows/release.yml
```diff
@@ -219,3 +249,51 @@ jobs:
...
+      run: echo "version=${GITHUB_REF#refs/tags/v}" >> $GITHUB_OUTPUT
+
+    - name: Load the current index.yaml
+      run: curl -o index.yaml https://charts.pganalyze.com/index.yaml
+      continue-on-error: true
+
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
+        role-to-assume: arn:aws:iam::123456789012:role/HelmChartPackageRole
```
keiko713, line 289: TODO: I need to update this, but before doing so, I'd like to ask @lfittl if this sounds okay to him

.github/workflows/release.yml
```diff
@@ -219,3 +249,51 @@ jobs:
         asset_path: ./pganalyze-collector_${{ steps.get_version.outputs.version }}_arm64.deb
         asset_name: pganalyze-collector_${{ steps.get_version.outputs.version }}_arm64.deb
         asset_content_type: application/octet-stream
+
+    - name: Upload helm package
```
keiko713, line 253: We could consider removing this step actually, because we will upload the package to s3. The downside of doing this is:
* the same package will be in GitHub release and s3
* the name will show up like pganalyze-collector-0.55.0.tgz in GitHub release, which might not be so obvious that it's about the helm chart

.github/workflows/release.yml
lfittl: If the concern here is the AWS account ID being public (since this is a public repo), that is fine - Account IDs aren't by themselves something sensitive in my understanding.

We should however double check that there isn't a way to exploit the OIDC connection from GH actions by opening a rogue PR that we approve to run by accident (e.g. that modifies the GH action to run on a PR instead of the main branch).

My understanding is that this can be restricted on the AWS side (in the IAM policy), so only the `main` branch (or possibly even just a tagged release?) can use the role: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#configuring-the-role-and-trust-policy

(we can discuss this in more detail in Slack)

keiko713: Ok I used the following as a trust policy (per https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#filtering-for-a-specific-tag), I think it works.

```
"token.actions.githubusercontent.com:sub": "repo:pganalyze/collector:ref:refs/tags/*"
```

With this, even if some bad actor is opening a rogue PR to force running this part of code, it won't work because the GitHub Action is not triggered by the tag.

lfittl (approved):
:+1:

keiko713:
I'm going to merge this 🎉
* We (well, I) should be watching if the release step goes well after merging this (maybe nice to cut a release sooner than later to find out if release CI is not broken)
  * For the worst case, even if the release step (esp. push to s3) is broken, we can still push the package manually to recover
* We (well, I) will update our docs to talk about k8s setup

PR #492, opened 2023-12-20
Release 0.52.4

keiko713:
Likely this will be the final release of 2023

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,23 @@
 # Changelog
 
+## 0.52.4      2023-12-21
+
+* Log Insights: Add support for receiving syslog over TLS
+  - You can configure a TLS certificate for the collector syslog server using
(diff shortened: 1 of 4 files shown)


msakrejda (approved):
Should we try to squeeze in support for the Azure CosmosDB default (and possibly unchangeable) log_line_prefix? I think I could do that later today.

keiko713:
> Should we try to squeeze in support for the Azure CosmosDB default (and possibly unchangeable) log_line_prefix? I think I could do that later today.

I actually don't know which one you're talking about, if it's small, sure let's squeeze, though I'd say no rush. We likely want to make a new release with https://github.com/pganalyze/collector/pull/476 early next year so it can go with that release too.

msakrejda:
Ok, let's just see if I can get more flexible log_line_prefix parsing in with #476 in the following release.

PR #493, opened 2023-12-22
Rename rel stats collection func for Citus not to use ext

keiko713:
Some refactor that I would like to split from https://github.com/pganalyze/collector/pull/476

Currently, in order to support collecting relation stats information for Citus, the name "ext" is used, like `handleRelationStatsExt`. In https://github.com/pganalyze/collector/pull/476, I would like to use the term `RelationStatsExtended` for the extended statistic info.
To avoid these two mixing together, I would like to rename the Citus one specific to Citus.

### DIFF
--- a/input/postgres/relation_stats.go
+++ b/input/postgres/relation_stats.go
@@ -195,7 +195,7 @@ func GetRelationStats(ctx context.Context, db *sql.DB, postgresVersion state.Pos
 		return
 	}
 
-	relStats, err = handleRelationStatsCitus(ctx, db, relStats, postgresVersion, server)
+	relStats, err = handleRelationStatsAux(ctx, db, relStats, postgresVersion, server)
 
 	return
 }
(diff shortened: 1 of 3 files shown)


msakrejda (approved):
+1 for renaming away from `...Ext`, but I think the original idea here was to use these functions to gather additional custom stats from any Postgres-based system, not just Citus. Since this is not part of a public API, though, which makes renames pretty cheap, I'm okay with making this Citus-specific and renaming if it comes to that. Alternately we could call this something like `handleRelationStatsExtra` or `handleRelationStatsAux` (for "auxiliary").

keiko713:
> Alternately we could call this something like `handleRelationStatsExtra` or `handleRelationStatsAux` (for "auxiliary").

Ah nice, I think Aux works well here. I was thinking that we could make this more generic name when we actually add something other than Citus, but using Aux here also works well 👍 (won't be using Extra as it's still kinda confusing to have both Extra and Extended)

PR #494, opened 2023-12-27
Add alternate mechanism for parsing log lines

msakrejda:
Our current log line parsing mechanism relies on matching each log
line to one of a set of recognized prefix patterns, then pulling
relevant information out of the prefix "manually" according to their
regex groups.

This works, but is difficult to maintain and requires customers to use
one of our prefixes. Occasionally, we add prefixes (e.g., to support a
provider that does not allow changing log_line_prefix), but this is
error-prone and requires the customer to wait for a collector release.

This patch takes a different approach: we inspect the log_line_prefix,
generate a regular expression to match lines coming with that specific
prefix pattern, and use that to pull relevant metadata out of the log
stream. The log_line_prefix is re-checked every full snapshot.

This branch supports both the old and the new parsing code. The code
path is controlled with the `db_log_line_prefix` collector setting: this
can be either `auto` (the new mechanism) or `legacy` (the old). It defaults
to the new code, but we could switch that.

TODO:

- [x] Get server's log_line_prefix and store where it can be passed to
      ParseLogLineWithPrefix
- [x] Avoid re-compiling the prefix regex on every call to
      ParseLogLineWithPrefix
- [x] Add handling for syslog log line format
- [x] Support calling both old and new ParseLogLineWithPrefix via
      collector setting to have an escape hatch in case of issues
  - we could do this via a db_log_line_prefix setting or similar, and
        set that to 'auto' (the new behavior), 'legacy' (the current behavior)
        or a literal prefix (the new behavior but prefix can't be read from
        settings for whatever reason)
- [x] Fix naming (right now, the core method is just called
      ParseLogLineWithPrefix2, but that's not ideal)
- [x] Add tests for untested prefixes (some of the existing
      allegedly-supported prefixes do not have tests in the old code,
      so it's not clear if the new code supports them—they should just
      work, but it'd be good to verify this.)
- [x] Warn when prefix does not include database or role
- [x] Benchmark (this is almost certainly faster, but it'd be good to
      confirm)


msakrejda:
Oh, another thing not covered is parsing lines received through syslog. I don't understand how syslog interacts with `log_line_prefix`--do we only support a single prefix when syslog is used?

msakrejda:
I should also add the case from #449; I think this fixes it.

msakrejda:
I'm marking this ready for review, since it's effectively complete. Sean volunteered to do the benchmark, but that shouldn't change things much.

seanlinsley:
Unsurprisingly, this log parser is significantly faster than the code it replaces (which had to rediscover the log line prefix for every log line individually). This simple benchmark shows the legacy parser takes 10 thousand nanoseconds per log line (0.01 milliseconds) while the new parser takes 64 nanoseconds. I didn't try to find a worst case log line, so the real-world performance of the old parser is likely worse.

```go
// File name: log_test.go

package main

import (
	"testing"
	"time"
	"github.com/pganalyze/collector/logs"
)

func BenchmarkLogParserLegacy(b *testing.B) {
	tz := time.FixedZone("UTC", 0)
	text := "2024-08-10 14:51:59 UTC:10.0.242.2(41322):role_name@database_name:[5840]:LOG: could not receive data from client: Connection reset by peer\n"
	for i := 0; i < b.N; i++ {
		logs.ParseLogLineWithPrefix("", text, tz)
	}
}

func BenchmarkLogParserNew(b *testing.B) {
	tz := time.FixedZone("UTC", 0)
	text := "2024-08-10 14:51:59 UTC:10.0.242.2(41322):role_name@database_name:[5840]:LOG: could not receive data from client: Connection reset by peer\n"
	parser := logs.NewLogParser(text, tz, false, false)
	for i := 0; i < b.N; i++ {
		parser.ParseLine(text)
	}
}
```

```sh
go test -bench=. -count 5

BenchmarkLogParserLegacy-10    	  105687	     10608 ns/op
BenchmarkLogParserLegacy-10    	  112977	     10605 ns/op
BenchmarkLogParserLegacy-10    	  113160	     10597 ns/op
BenchmarkLogParserLegacy-10    	  113565	     10557 ns/op
BenchmarkLogParserLegacy-10    	  113646	     10587 ns/op
BenchmarkLogParserNew-10       	18397794	        64.05 ns/op
BenchmarkLogParserNew-10       	18273276	        64.31 ns/op
BenchmarkLogParserNew-10       	18642622	        64.16 ns/op
BenchmarkLogParserNew-10       	18834459	        63.25 ns/op
BenchmarkLogParserNew-10       	18938060	        63.34 ns/op
```

msakrejda:
Great, thanks for benchmarking! The `useLegacyFallback` param just [calls the old code](https://github.com/pganalyze/collector/pull/494/files#diff-b0823247f6e9e6c3f1033cc07a941a0435fea5c09f64ae4352b92b6ef8246614R358-R360) directly (to have an escape hatch in case there's an issue with the new code), so the perf there is expected.

logs/parse.go
```diff
@@ -1,519 +1,249 @@
...
-	ok = true
-
-	return
+func (lp *LogParser) Matches(prefix string, tz *time.Location, isSyslog bool) bool {
+	return lp.prefix == prefix && tz.String() == lp.tz.String() && lp.isSyslog == isSyslog
 }
 
-func GetOccurredAt(timePart string, tz *time.Location, rsyslog bool) time.Time {
-	if tz != nil && !rsyslog {
+func (lp *LogParser) GetOccurredAt(timePart string) time.Time {
```
keiko713, line 239: memo: this is pretty much the same logic as previous one (well, I guess it's obvious from this diff)

logs/parse.go
```diff
@@ -1,519 +1,249 @@
...
-
-func IsSupportedPrefix(prefix string) bool {
-	for _, supportedPrefix := range SupportedPrefixes {
-		if supportedPrefix == prefix {
-			return true
-		}
-	}
-	return false
+// not included: %q and %%, which are easier to handle by special-casing
+var EscapeMatchers = map[rune]PrefixEscape{
```
keiko713, line 24: I wonder if it's nice to have a reference here (e.g. link to https://www.postgresql.org/docs/current/runtime-config-logging.html#GUC-LOG-LINE-PREFIX)
(things like "see Section 74.1" is pretty mystery without it)

logs/parse.go
```diff
@@ -569,66 +299,149 @@ func GetOccurredAt(timePart string, tz *time.Location, rsyslog bool) time.Time {
 	return ts
 }
 
-type LineReader interface {
-	ReadString(delim byte) (string, error)
+func parseSyslogLine(line string, tz *time.Location) (logLine state.LogLine, ok bool) {
```
keiko713, line 302: memo: this is actually using the same way as legacy code

logs/parse.go
keiko713: Curious: how did you know if the value is optional or not?

keiko713 (approved):
Sorry took time but finally had some brain power to review this 🧠 Overall, looking great and changes make sense. It's nice to see that the benchmark is also looking good.
Re: default, I feel like it's okay to have `auto` as a default. It would be really ideal to test with multiple platforms, though. 

> The log_line_prefix is re-checked every full snapshot

Since logs are parsed more frequently than full snapshots, this means that it's possible to miss some logs after the `log_line_prefix` change until the next snapshot?

Also TODO (likely already in your TODO list): you'll need to update docs quite a lot

main.go
```diff
@@ -619,9 +619,11 @@ func checkOneInitialCollectionStatus(ctx context.Context, server *state.Server,
...
 	}
 
-	server.SetLogTimezone(settings)
-	if server.LogTimezone == nil {
-		logger.PrintWarning("Could not determine log timezone for this server: %s")
+	logs.SyncLogParser(server, settings)
+	parser := server.GetLogParser()
+	prefixErr := parser.ValidatePrefix()
+	if prefixErr != nil {
+		logger.PrintWarning("Checking log_line_prefix: %d", prefixErr)
```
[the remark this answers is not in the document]
msakrejda: Good catch, thanks.

logs/parse.go
msakrejda: Good call on the link.

And re: optional, I _think_ it was just things that will not be printed by non-session processes. Normally you'd use `%q` to omit that entirely, but that's not required, so you may end up with stuff in the prefix that just won't be filled in by some processes.

msakrejda:
Thanks for the review, @keiko713. Re:

>Since logs are parsed more frequently than full snapshots, this means that it's possible to miss some logs after the log_line_prefix change until the next snapshot?

Yes. But I think it's a safe bet that `log_line_prefix` does not change often, so the impact here is pretty limited.

msakrejda:
Oh, and good call on the docs, although this is basically backward-compatible with the current docs, so we can update that later.

msakrejda:
Tested successfully on RDS, self-managed, and Heroku.

msakrejda:
Tested successfully on GCP (after a couple of fixes to the refactoring that was needed to use the new parser).

keiko713:
I tested this with Azure but got bitten by the other thing. Despite I tested with Flexible server, the eventhub message was including `LogicalServerName` and wrongly tweaking the LogLineContent here (the log looked like "LOG:  2024-08-20 02:42:09.287 UTC [4708] [user=[unknown],db=[unknown],app=[unknown]] LOG:  connection received: host=127.0.0.1 port=36120"), hence the parse failed:

https://github.com/pganalyze/collector/pull/494/files#diff-6f5880431e90b30b7cc480560bbf2f8e4a80fcbe85c03ca3cacae1276cb3512eR245
https://github.com/pganalyze/collector/blob/1d1e4daae0f862b2729c4685c2698ef779d3884b/input/system/azure/logs.go#L242

It even fails with the current code so I think we should handle this in the separate PR. Looking at the doc, Flexible server shouldn't have `LogicalServerName`, so not sure why it's now having it. I'm planning to ask around.

https://learn.microsoft.com/en-us/azure/postgresql/single-server/concepts-server-logs
https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/concepts-logging

To check if this PR is working outside of this, I disabled problematic logic temporary and confirmed that it worked 👍

msakrejda:
Thanks for testing, Keiko. I agree; let's handle that issue separately.

PR #495, opened 2024-01-06
Docker image: Don't reload when calling "test" command

lfittl:
When we changed --test to automatically reload, we forgot to explicitly avoid that in the case of the Docker image. In the Docker situation we would typically find two separate containers during a test, the one used for the test, and the regular container running in the background. Thus, the test won't be able to reload the PID that's running in a separate container / cgroups namespace.

### DIFF
--- a/contrib/docker-entrypoint.sh
+++ b/contrib/docker-entrypoint.sh
@@ -12,7 +12,7 @@ if [ $(id -u) = 0 ]; then
 fi
 
 if [ "$1" = 'test' ]; then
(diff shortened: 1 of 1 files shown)


lfittl:
> Makes sense. Another situation we should fix is on Heroku (the actual collector is running in a separate dyno, so there's nothing to reload), but that's different enough that we can do it separately.

For Heroku that'd be more a documentation thing, right? (since we don't have any special reload task in the [Procfile](https://github.com/pganalyze/collector/blob/main/Procfile) currently)

PR #503, opened 2024-02-18
Self-hosted: Allow collector to receive logs via OpenTelemetry

lfittl:
This is mainly intended to support Kubernetes, by using fluentbit with the opentelemetry output plugin to send pod logs to the collector. The logs received can optionally be filtered by pod name, or via label selectors.

Note this requires the log output to be jsonlog (Postgres 15+), with optional K8s context added, or plain logs without any additional context.

An example fluentbit output configuration:

```
  [OUTPUT]
    name  opentelemetry
    match kube.*postgres*
    host  127.0.0.1
    port  4318
```

With a corresponding collector configuration:

```
  db_log_otel_server = 127.0.0.1:4318
```

When utilizing CloudNativePG, a filter like the following can be used to filter only to the primary's logs for a given cluster:

```
  db_log_otel_k8s_labels = cnpg.io/cluster=cluster-example,cnpg.io/instanceRole=primary
```

## TODO

* [ ] Test additional operators besides CloudNativePG (confirm whether we have any assumptions about container names/etc) -- if this works as expected consider dropping the "logger" TODO in the code
* [ ] Add explicit unit tests for K8s label selector handling
* [ ] Review whether jsonlog handling could share code with Tembo log handler


ianstanton (approved):

config/config.go
```diff
@@ -157,6 +157,12 @@ type ServerConfig struct {
 	LogSyslogServerKeyContents      string `ini:"db_log_syslog_server_key_contents"`
 	LogSyslogServerClientCAContents string `ini:"db_log_syslog_server_client_ca_contents"`
 
+	// Configures the collector to start a built-in OTLP HTTP server that
+	// receives logs and optionally filters them by Kubernetes pod or labels
+	LogOtelServer    string `ini:"db_log_otel_server"`
+	LogOtelK8SPod    string `ini:"db_log_otel_k8s_pod"`
+	LogOtelK8SLabels string `ini:"db_log_otel_k8s_labels"`
```
keiko713, line 164: It would be nice to have a way to specify this via config vars too, like adding the following in read.go:

```golang
	if logOtelServer := os.Getenv("LOG_OTEL_SERVER"); logOtelServer != "" {
		config.LogOtelServer = logOtelServer
	}
	if logOtelK8SPod := os.Getenv("LOG_OTEL_K8S_POD"); logOtelK8SPod != "" {
		config.LogOtelK8SPod = logOtelK8SPod
	}
	if logOtelK8SLabels := os.Getenv("LOG_OTEL_K8S_LABELS"); logOtelK8SLabels != "" {
		config.LogOtelK8SLabels = logOtelK8SLabels
	}
```

keiko713:
Tested with crunchy pgo. Looks like by default, the logs are stored in somewhere like `/pgdata/pg16/log/postgresql-Thu.log` in the database container, using `logging_collector` feature: 
https://github.com/CrunchyData/postgres-operator/issues/2088

This means, using fluentbit, at least this way, doesn't work because logs are not retrievable with the normal INPUT that fluentbit does (aka stdout of the container).
I turned off  `logging_collector` feature then it went to stdout, though the structure is really different from CloudNativePG one; especially, we assume that the message contains something like `"logger"=>"postgres"`, but it is not there. Here is the example:

```
[
    [1711018281.787688681, {}],
    {
        "time"=>"2024-03-21T10:51:21.787688681Z",
        "stream"=>"stderr",
        "_p"=>"F",
        "log"=>"2024-03-21 10:51:21.787 UTC [273] LOG:  duration: 13010.025 ms  statement: select pg_sleep(13);",
        "kubernetes"=>{...}
    }
]
```

Whereas CloudNativePG one is like this:

```
[
    [1711019024.635673287, {}],
    {
        "time"=>"2024-03-21T11:03:44.635673287Z",
        "stream"=>"stderr",
        "_p"=>"F",
        "level"=>"info",
        "ts"=>"2024-03-21T11:03:44Z",
        "logger"=>"postgres",
        "msg"=>"record",
        "logging_pod"=>"owls-db-3",
        "record"=>{
            "log_time"=>"2024-03-21 11:03:44.635 UTC",
            "process_id"=>"22",
            "session_id"=>"65faae9d.16",
            "session_line_num"=>"432",
            "session_start_time"=>"2024-03-20 09:38:37 UTC",
            "transaction_id"=>"0",
            "error_severity"=>"LOG",
            "sql_state_code"=>"00000",
            "message"=>"recovery restart point at 0/95000090",
            "detail"=>"Last completed transaction was at log time 2024-03-21 11:00:00.33255+00.",
            "backend_type"=>"checkpointer", "query_id"=>"0"
        },
        "kubernetes"=>{...}
    }
]
```

Uhmm, could also be the difference between jsonlog? Though if you have any input about this, please let me know. I'm planning to try out with Zalando tomorrow maybe.

keiko713:
Other point: I've been testing this with deploying everything in the same k8s cluster (postgres, fluent-bit, pganalyze collector). With this, fluent-bit and pganalyze collector won't be in the same host, so we shouldn't be able to specify it with `host  127.0.0.1` (I'm actually curious how you were testing without doing this, I might be missing something).
IOW, we likely need to introduce a Service for pganalyze-collector (e.g. `pganalyze-collector-otel-service`), then fluent-bit can specify it as `host`. I think this helm chart update (to include the service creation) can be added to this PR too.

input/system/selfhosted/otel_handler.go
```diff
@@ -0,0 +1,211 @@
...
+								if v.Key == "kubernetes" {
+									kubernetes = v.Value.GetKvlistValue()
+								}
+								if v.Key == "error_severity" {
+									hasErrorSeverity = true
+								}
+							}
+							// TODO: Does the logger name need to be configurable? (this is tested with CNPG,
+							// but would the logger be different with other operators/container names?)
+							if logger == "postgres" {
```
keiko713, line 178: I tweaked here to something like the following and I was able to make it work with Crunchy + `logging_collector` turned off:

```
							if handleWithK8sFilter(kubernetes, server) {
								prefixedLogger.PrintInfo("Received logs matching to the filter\n")
								// handle log as plain log for now
								// "log"=>"2024-03-22 09:40:37.089 UTC [320] [user=postgres,db=postgres,app=psql] LOG:  duration: 13013.199 ms  statement: select pg_sleep(13);"
								item := SelfHostedLogStreamItem{}
								item.Line = log
								item.OccurredAt = time.Unix(0, int64(l.TimeUnixNano))
								rawLogStream <- item
							} else if logger == "postgres" {
```

That said, I don't think this is something people would do (though I think it's still good to introduce the "match filter" that's what I did above with `handleWithK8sFilter`, as opposed to "deny filter" that this PR currently has).
I'd love to learn more how people are doing logs with Crunchy postgres operator. There is a blog post about it using vector (https://www.crunchydata.com/blog/log-export-examples-using-crunchy-postgres-for-kubernetes), which we can try but I think vector doesn't have otel sink.

keiko713:
My plan: I'm thinking to create a new PR as this PR gets a bit messy (due to my comments, sorry).
I'm going to focus on making the PR to support fluentbit + otel case with cnpg, and basically make all "TODO" tasks as a follow-up task. I'll do this tomorrow or next week

keiko713:
going to close and make a new PR as mentioned (still doing some testing so I'll open once the test is done)

PR #507, opened 2024-02-23
Only collect extended stats for postgres version 12 and above

keiko713:
The view `pg_stats_ext` was introduced from Postgres 12, so let's only try to collect extended stats data if the version is 12 and above.
https://www.postgresql.org/docs/current/view-pg-stats-ext.html


lfittl (approved):
:+1:

PR #508, opened 2024-02-23
Release v0.54.0

keiko713:


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 0.54.0      2024-02-23
+
+* Update pg_query_go to v5 / Postgres 16 parser
```
keiko713, line 5: When we bumped the major version previously, we were saying something more, but mostly something from the PR description (https://github.com/pganalyze/collector/pull/383). As we didn't have much to mention this time (https://github.com/pganalyze/collector/pull/502), I'd say that we're okay saying nothing here?

lfittl: Yeah, I think this is fine - technically we could mention things from here, but I don't think we need to: https://github.com/pganalyze/libpg_query/blob/16-latest/CHANGELOG.md (e.g. Windows support now exists in pg_query, but the collector isn't yet being built on Windows, etc)

msakrejda (approved):

PR #509, opened 2024-02-28
Add system metrics support for Tembo Cloud

ianstanton:
Allow for system metrics when the collector is connected to a Tembo Cloud instance:
- Add `TEMBO_NAMESPACE` environment variable
- If SystemType is `tembo`, collect system metrics
- Collect the following system metrics from Tembo API
  - CPU percentage
  - Total memory
  - Available memory
  - Disk capacity
  - Disk available

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -116,6 +116,7 @@ type ServerConfig struct {
 	AivenServiceID string `ini:"aiven_service_id"`
 
 	TemboAPIToken         string `ini:"tembo_api_token"`
+	TemboAPIURL           string `ini:"tembo_api_url"`
 	TemboInstanceID       string `ini:"tembo_instance_id"`
 	TemboMetricsNamespace string `ini:"tembo_metrics_namespace"`
 	TemboOrgID            string `ini:"tembo_org_id"`
--- a/config/config.go
+++ b/config/config.go
@@ -116,10 +116,10 @@ type ServerConfig struct {
 	AivenServiceID string `ini:"aiven_service_id"`
 
 	TemboAPIToken         string `ini:"tembo_api_token"`
-	TemboMetricsAPIURL    string `ini:"tembo_metrics_api_url"`
-	TemboInstanceID       string `ini:"tembo_instance_id"`
-	TemboMetricsNamespace string `ini:"tembo_metrics_namespace"`
 	TemboOrgID            string `ini:"tembo_org_id"`
+	TemboInstanceID       string `ini:"tembo_instance_id"`
+	TemboMetricsAPIURL    string `ini:"tembo_metrics_api_url"`
+	TemboMetricsNamespace string `ini:"tembo_metrics_namespace"`
 
 	SectionName string
 	Identifier  ServerIdentifier
--- a/config/config.go
+++ b/config/config.go
@@ -116,7 +116,7 @@ type ServerConfig struct {
 	AivenServiceID string `ini:"aiven_service_id"`
 
 	TemboAPIToken         string `ini:"tembo_api_token"`
-	TemboAPIURL           string `ini:"tembo_api_url"`
+	TemboMetricsAPIURL    string `ini:"tembo_metrics_api_url"`
 	TemboInstanceID       string `ini:"tembo_instance_id"`
 	TemboMetricsNamespace string `ini:"tembo_metrics_namespace"`
 	TemboOrgID            string `ini:"tembo_org_id"`
--- a/config/config.go
+++ b/config/config.go
@@ -115,6 +115,10 @@ type ServerConfig struct {
 	AivenProjectID string `ini:"aiven_project_id"`
 	AivenServiceID string `ini:"aiven_service_id"`
 
+	TemboAPIToken   string `ini:"tembo_api_token"`
+	TemboInstanceID string `ini:"tembo_instance_id"`
+	TemboOrgID      string `ini:"tembo_org_id"`
+
 	SectionName string
 	Identifier  ServerIdentifier
 
--- a/config/config.go
+++ b/config/config.go
@@ -115,10 +115,10 @@ type ServerConfig struct {
 	AivenProjectID string `ini:"aiven_project_id"`
 	AivenServiceID string `ini:"aiven_service_id"`
 
-	TemboAPIToken   string `ini:"tembo_api_token"`
-	TemboInstanceID string `ini:"tembo_instance_id"`
-	TemboNamespace  string `ini:"tembo_namespace"`
-	TemboOrgID      string `ini:"tembo_org_id"`
+	TemboAPIToken         string `ini:"tembo_api_token"`
+	TemboInstanceID       string `ini:"tembo_instance_id"`
+	TemboMetricsNamespace string `ini:"tembo_metrics_namespace"`
+	TemboOrgID            string `ini:"tembo_org_id"`
 
 	SectionName string
 	Identifier  ServerIdentifier
--- a/config/config.go
+++ b/config/config.go
@@ -117,6 +117,7 @@ type ServerConfig struct {
 
 	TemboAPIToken   string `ini:"tembo_api_token"`
 	TemboInstanceID string `ini:"tembo_instance_id"`
+	TemboNamespace  string `ini:"tembo_namespace"`
 	TemboOrgID      string `ini:"tembo_org_id"`
 
 	SectionName string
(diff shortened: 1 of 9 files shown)


input/system/tembo/system.go
```diff
@@ -0,0 +1,165 @@
...
+		return
+	}
+
+	system.CPUStats = make(state.CPUStatisticMap)
+	system.CPUStats["all"] = state.CPUStatistic{
+		DiffedOnInput: true,
+		DiffedValues: &state.DiffedSystemCPUStats{
+			UserPercent: cpuUsage,
+		},
+	}
```
ianstanton, line 63: @lfittl What should the format of `cpuUsage` look like here? Right now we're receiving something like `0.04305000823670472`. In the pganalyze app it shows a constant 100% utilization.

lfittl: This actually looks correct, we just need to adjust something inside the app to show this correctly based on the system type.

grant/logs.go
```diff
@@ -18,7 +18,8 @@ func GetLogsGrant(server *state.Server, globalCollectionOpts state.CollectionOpt
 
 	req.Header.Set("Pganalyze-Api-Key", server.Config.APIKey)
 	req.Header.Set("Pganalyze-System-Id", server.Config.SystemID)
-	req.Header.Set("Pganalyze-System-Type", server.Config.SystemType)
+	// TODO(ianstanton) Placeholder for local dev
+	req.Header.Set("Pganalyze-System-Type", "self_hosted")
```
lfittl, line 22: FYI, this should be good to drop now since we added the `tembo` system type.

config/identify_system.go
```diff
@@ -82,6 +82,15 @@ func identifySystem(config ServerConfig) (systemID string, systemType string, sy
 		if systemScopeFallback == "" {
 			systemScopeFallback = selfManagedSystemScope(config)
 		}
+	} else if (config.TemboInstanceID != "") || systemType == "tembo" {
+		systemType = "tembo"
+		// TODO(ianstanton) These should be tembo specific
```
lfittl, line 87: As discussed, something like this would make sense (in our call we considered whether it could make sense to auto-detect this from the hostname, but given that the hostname does not include these identifiers, and the collector is configured through Tembo, it seems better to require these to be set in the collector config).

```golang
		if systemID == "" {
			systemID = config.TemboInstanceID
		}
		if systemScope == "" {
			systemScope = config.TemboOrgID
		}
```

grant/default.go
```diff
@@ -18,7 +18,8 @@ func GetDefaultGrant(server *state.Server, globalCollectionOpts state.Collection
 
 	req.Header.Set("Pganalyze-Api-Key", server.Config.APIKey)
 	req.Header.Set("Pganalyze-System-Id", server.Config.SystemID)
-	req.Header.Set("Pganalyze-System-Type", server.Config.SystemType)
+	// TODO(ianstanton) Placeholder for local dev
+	req.Header.Set("Pganalyze-System-Type", "self_hosted")
```
lfittl, line 22: FYI, this should be good to drop now since we added the `tembo` system type.

input/system/tembo/system.go
```diff
@@ -0,0 +1,165 @@
...
+	headers := map[string]string{
+		"Authorization": "Bearer " + config.TemboAPIToken,
+		"accept":        "application/json",
+	}
+
+	client := http.Client{}
+
+	// Get CPU usage percentage
+	query := "sum(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate{ namespace=\"" + config.TemboNamespace + "\"}) / sum(kube_pod_container_resource_requests{job=\"kube-state-metrics\",  namespace=\"" + config.TemboNamespace + "\", resource=\"cpu\"})"
+	cpuUsage, err := getFloat64(query, "https://api.data-1.use1.tembo.io/"+config.TemboNamespace+"/metrics/query?query=", client, headers)
```
lfittl, line 51: As discussed both the `data-1.use1` and the `tembo.io` parts need to be configurable (for different regions, and dev/staging envs).

input/system/tembo/system.go
```diff
@@ -0,0 +1,165 @@
...
+		logger.PrintError("Tembo/System: Encountered error when getting memory info %v\n", err)
+		return
+	}
+
+	system.Memory.TotalBytes = memoryTotalBytes
+	system.Memory.AvailableBytes = memoryAvailableBytes
+	system.Memory.FreeBytes = memoryAvailableBytes
+
+	// Get disk capacity
+	//TODO(ianstanton) Check if volume claim names differ in cases like HA
```
lfittl, line 86: As discussed, would be good to double check before merging (and if not an issue to remove the TODO)

input/system/tembo/system.go
```diff
@@ -0,0 +1,165 @@
...
+	diskAvailable, err := getUint64(query, "https://api.data-1.use1.tembo.io/"+config.TemboNamespace+"/metrics/query?query=", client, headers)
+	if err != nil {
+		logger.PrintError("Tembo/System: Encountered error when getting disk info %v\n", err)
+		return
+	}
+
+	diskUsed := diskCapacity - diskAvailable
+	system.DiskPartitions = make(state.DiskPartitionMap)
+	system.DiskPartitions["/"] = state.DiskPartition{
+		DiskName:      "md0",
```
lfittl, line 106:
```suggestion
		DiskName: "default",
```

And I'd drop the PartitionName since its not needed here.

config/config.go
```diff
@@ -115,6 +115,11 @@ type ServerConfig struct {
 	AivenProjectID string `ini:"aiven_project_id"`
 	AivenServiceID string `ini:"aiven_service_id"`
 
+	TemboAPIToken   string `ini:"tembo_api_token"`
+	TemboInstanceID string `ini:"tembo_instance_id"`
+	TemboNamespace  string `ini:"tembo_namespace"`
```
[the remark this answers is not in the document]
ianstanton: Updated

config/identify_system.go
ianstanton: Updated

input/system/tembo/system.go
ianstanton: We do have separate PVCs when HA is enabled. I'd like to handle this as a follow-up if it's alright with you. Example:
```
⨠ kubectl get po -n org-demo-inst-pganalyze-test 
NAME                                                      READY   STATUS      RESTARTS   AGE
org-demo-inst-pganalyze-test-1                            1/1     Running     0          11h
org-demo-inst-pganalyze-test-2                            0/1     Running     0          102s
org-demo-inst-pganalyze-test-2-snapshot-recovery-s6dfb    0/1     Completed   0          119s
org-demo-inst-pganalyze-test-pganalyze-86f568fb59-k66lm   1/1     Running     0          3d9h


 ⨠ kubectl get pvc -n org-demo-inst-pganalyze-test 
NAME                             STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
org-demo-inst-pganalyze-test-1   Bound    pvc-2793b508-f3f5-44f2-8b50-dd3b3d387745   10Gi       RWO            gp3-enc        27d
org-demo-inst-pganalyze-test-2   Bound    pvc-3428da10-62a4-4670-b4ba-5b9e9e963ba5   10Gi       RWO            gp3-enc        3m34s

```

grant/logs.go
ianstanton: Updated

grant/default.go
ianstanton: Updated

input/system/tembo/system.go
ianstanton: I added a new env var for `TEMBO_METRICS_API_URL`:
- https://github.com/pganalyze/collector/pull/509/commits/397c4a4641fdfd677b3b92f0e2b664ed0bf44c8c
- https://github.com/pganalyze/collector/pull/509/commits/17dc38f92d47766a4fd768fce3944df3be9e4d8e

Let me know what you think of this approach

ianstanton:
@lfittl This one is ready for another pass. I'll run things again in the AM to make sure things are still working as expected after these most recent commits.

lfittl (approved):
Overall looks good!

A few more minor notes, but otherwise happy to merge this in.

ianstanton:
Thanks @lfittl! I've addressed all feedback in https://github.com/pganalyze/collector/pull/509/commits/491615c394401aaa3ae1e8dbe216b22085744697

ianstanton:
@lfittl Are we able to merge this one today?

lfittl:
> @lfittl Are we able to merge this one today?

Yep, done! Thanks for your work on this :)

PR #512, opened 2024-03-07
Relation stats: Call pg_stat_get_* directly instead of using system views

lfittl:
Due to the way how pg_stat_all_tables and pg_statio_all_tables are defined, its significantly more expensive to query through the views when a table filter is in place, since all the data gets pulled, only for the filter to be applied afterwards.

Instead, call the underlying pg_stat_get* functions directly, which has the same effect as querying the views (as they are simple views without any security barrier), and shows a major speedup in cases where a large portion of tables is filtered out. This also should provide a minor speedup in regular operations, even with no table filter in place.

### DIFF
--- a/input/postgres/relation_stats.go
+++ b/input/postgres/relation_stats.go
@@ -8,7 +8,7 @@ import (
 	"github.com/pganalyze/collector/state"
 )
 
-const relationStatsSQLInsertsSinceVacuumFieldPg13 string = "s.n_ins_since_vacuum"
+const relationStatsSQLInsertsSinceVacuumFieldPg13 string = "pg_stat_get_ins_since_vacuum(c.oid)"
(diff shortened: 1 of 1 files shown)


msakrejda:
For the record, this rewrite of the query did not perform faster for the customer for whom this was a problem (we suspect they had more general I/O issues). Is this still worth doing?

lfittl:
> For the record, this rewrite of the query did not perform faster for the customer for whom this was a problem (we suspect they had more general I/O issues). Is this still worth doing?

Maybe? I haven't identified a downside to this approach, and it does save a few cycles for all cases, and more so when a table filter is active (even if its not necessarily a big performance impact for regular hardware).

msakrejda:
Yeah, makes sense.

msakrejda (approved):
Double-checked the SQL against the source view definitions, and looks good to me.

PR #513, opened 2024-03-08
Colorize output

msakrejda:
First stab at colorizing collector output

Only colorize the log level, so that warnings and errors stand out
without making the output too nosiy.


msakrejda:
Sample output:
image

This is probably still too subtle (and the output needs cleanup in other ways to make this usable), but it's a start. We should also test this with light terminal backgrounds.

lfittl:
> This is probably still too subtle (and the output needs cleanup in other ways to make this usable), but it's a start. We should also test this with light terminal backgrounds.

Should we focus our efforts here on the `--test` command? (i.e. instead of doing this at the logger level, re-think how the `--test` command outputs its progress/result?)

I don't know how the color coding gets parsed by various log processing tools, but I suspect its of little added value to have for logs from the collector running in the background.

msakrejda:
> I don't know how the color coding gets parsed by various log processing tools, but I suspect its of little added value to have for logs from the collector running in the background.

The library I added checks if stdout is a tty and no-ops coloring if not, so there's no impact on "real" log output.

> Should we focus our efforts here on the --test command? (i.e. instead of doing this at the logger level, re-think how the --test command outputs its progress/result?)

But yeah. I wanted to add colorizing first and then revamp test output, but after sleeping on it, I don't think _just_ colorizing output will move the needle on the test output experience. I'll revamp the output instead, adding color if it makes sense.

PR #516, opened 2024-03-19
Accept PGA_API_BASE_URL env var in addition to PGA_API_BASEURL

msakrejda:
Currently, the env var corresponding to the config setting
api_base_url is PGA_API_BASEURL, which has different spacing.

Allow both (but prefer the documented version).

### DIFF
--- a/config/read.go
+++ b/config/read.go
@@ -55,6 +55,11 @@ func getDefaultConfig() *ServerConfig {
 	if apiKey := os.Getenv("PGA_API_KEY"); apiKey != "" {
 		config.APIKey = apiKey
 	}
+	// Since there's a discrepancy here between the config file key (api_base_url) and the
+	// env var key (PGA_API_BASEURL), also accept PGA_API_BASE_URL to make our lives easier.
+	if apiBaseURL := os.Getenv("PGA_API_BASE_URL"); apiBaseURL != "" {
+		config.APIBaseURL = apiBaseURL
+	}
 	if apiBaseURL := os.Getenv("PGA_API_BASEURL"); apiBaseURL != "" {
 		config.APIBaseURL = apiBaseURL
 	}
--- a/config/read.go
+++ b/config/read.go
@@ -55,12 +55,13 @@ func getDefaultConfig() *ServerConfig {
 	if apiKey := os.Getenv("PGA_API_KEY"); apiKey != "" {
 		config.APIKey = apiKey
 	}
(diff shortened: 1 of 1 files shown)


config/read.go
```diff
@@ -55,6 +55,11 @@ func getDefaultConfig() *ServerConfig {
 	if apiKey := os.Getenv("PGA_API_KEY"); apiKey != "" {
 		config.APIKey = apiKey
 	}
+	// Since there's a discrepancy here between the config file key (api_base_url) and the
+	// env var key (PGA_API_BASEURL), also accept PGA_API_BASE_URL to make our lives easier.
```
msakrejda, line 59: Alternately, we could standardize on this one for consistency (and still accept the old spelling).

lfittl (approved):

config/read.go
lfittl: Assuming you mean `baseurl` with this one (i.e. without the `_`), I think that will be a bit more annoying to handle for the ini field names, so it seems better to deprecate `PGA_API_BASEURL` and use `BASE_URL` going forward.

config/read.go
```diff
@@ -55,6 +55,11 @@ func getDefaultConfig() *ServerConfig {
 	if apiKey := os.Getenv("PGA_API_KEY"); apiKey != "" {
 		config.APIKey = apiKey
 	}
+	// Since there's a discrepancy here between the config file key (api_base_url) and the
```
lfittl, line 58: I wonder a little bit if instead of this comment we should have one before `PGA_API_BASEURL` below that says its deprecated but we keep this for backwards compatibility.

msakrejda: Oh, I did not intend to change the "canonical" key here: we'd need to change the docs to do that. But if you think that's worth doing, sure, that makes sense to me (and I'm happy to make the docs change as well).

config/read.go
msakrejda: No, I meant align with the config value (with the dash). So I'll go ahead and make the new spelling canonical (and update docs).

config/read.go
lfittl: Yeah, I think its worth doing the docs change as well. Maybe with a small note saying "This used to be called PGA_API_BASEURL before collector release 0.XX.X", in case someone is on an old collector version.

keiko713 (approved):

msakrejda:
See also https://github.com/pganalyze/pganalyze-docs/pull/231

PR #518, opened 2024-03-22
Fix spurious errors on Heroku

msakrejda:
- don't complain about a missing state file: this is expected since
   dynos are stateless and there's unlikely to have been a previous
   run of the collector in the same dyno

 - do nothing on reload: our reload mechanism looks for a running
   collector process, but on heroku, that's going to be in a different
   dyno

### DIFF
--- a/config/read.go
+++ b/config/read.go
@@ -736,7 +736,7 @@ func Read(logger *util.Logger, filename string) (Config, error) {
 			return conf, fmt.Errorf("Configuration file is empty, please edit %s and reload the collector", filename)
 		}
 	} else {
-		if os.Getenv("DYNO") != "" && os.Getenv("PORT") != "" {
+		if util.IsHeroku() {
 			for _, kv := range os.Environ() {
 				parts := strings.SplitN(kv, "=", 2)
 				parsedKey := parts[0]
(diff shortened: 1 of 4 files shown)


msakrejda:
One or both of these changes may also apply to Docker, but I'm not sure if we should no-op a reload in Docker, since the user would still need to restart the "main" container (whereas on Heroku, applying any config changes automatically restarts the running dyno).

lfittl (approved):

lfittl:
> One or both of these changes may also apply to Docker, but I'm not sure if we should no-op a reload in Docker, since the user would still need to restart the "main" container (whereas on Heroku, applying any config changes automatically restarts the running dyno).

In the Docker case we actually added `--no-reload` to the collector Docker entrypoint: https://github.com/pganalyze/collector/commit/551df82570109d60f96e64cf42dda0ea6005f3e0

I imagine we could do a similar change to the Procfile used for Heroku, instead of special casing it in the source?

msakrejda:
Ah, good point. Will do.

msakrejda:
Hmm, actually, it's the test command we'd need to change, which is [in the docs](https://pganalyze.com/docs/install/heroku_postgres/03_verify_configuration) rather than in the Procfile. We could add the flag there, but I don't know that we should make users type more to prevent an unwanted behavior that we can handle automatically...

lfittl:
> Hmm, actually, it's the test command we'd need to change, which is [in the docs](https://pganalyze.com/docs/install/heroku_postgres/03_verify_configuration) rather than in the Procfile. We could add the flag there, but I don't know that we should make users type more to prevent an unwanted behavior that we can handle automatically...

Ah, good point. Let's just do it in the code then (like you've done here) - the extra conditional is not a problem I think.

PR #519, opened 2024-03-26
Better test output

msakrejda:
This revamps collector test output to make it clearer when there are integration issues and, if there are none and everything is okay, when various parts of the UI can expect to be populated.

TODO:
 - [x] support format strings in SelfCheck methods to simplify call sites
 - [x] complete Logs check for RDS and Crunchy (and any other future providers that use log downloads)
 - [x] gate all self-test status tracking behind `--test` flag so it's skipped outside of tests (currently, the code does not do that consistently)
 - [x] complete Index Advisor check
 - [x] complete VACUUM Advisor check
 - [x] add pganalyze connection check (right now, if the test can't reach the pganalyze API, it fails correctly, but opaquely)
 - [x] add placeholder EXPLAIN check based on what we can check currently (i.e., we can point out when we know it's broken; we don't have a test in place to validate it works within the collector)
 - [x] go through remaining output from logger and determine if any of that should be added to the test run check
 - [x] add hints for common issues (e.g., shared_preload_libraries updated but server needs reboot for that to take effect)
 - [x] reformat output (consider inlining pganalyze URLs into server chunks)
 - [x] decide whether we want to add checks corresponding to other in-app top-level menu items (Connections, Config Settings, Alerts & Check-Up)
   - included connections, but we can ignore the others for now
 - [x] rename SelfCheck to SelfTest (tried this in https://github.com/pganalyze/collector/commit/c8a7ef71b7c9d792e943ce6281142cfa601df36d but that somehow broke the build with a complaint about an unreferenced symbol; will try again)
 - [x] address TODOs
 - [x] code clean up

Future work:
 - [ ] complete EXPLAIN check (add the necessary test code)
 - [ ] clean up obsolete test-run-only log output


msakrejda:
Okay, this turned out to be way trickier than I anticipated and is not quite ready, but I'd love some input at this point @pganalyze/dev and @jen.  The output looks more

msakrejda:
Okay, this turned out to be way trickier than I anticipated and is still not quite ready, but I'd love some input at this point @pganalyze/dev and @jensy (when you're back). The output looks something like this:

image

I originally wanted to replace the log output of the test run, but I realized that may not be a good idea (at least yet), since we're somewhat inconsistent about error handling and we could end up in a situation where the self-check error is never reported because of that.  I'd rather make this output purely additive for now, so users can always go back and sift through the logs if they run into an edge case not covered by the summary output.

The general idea behind the output format is, for each server, first a section for system information, then for low-level checks for individual aspects of collection, then a summary of those aspects in terms of how they relate to app features that are or are not expected to be working. With the `--verbose` flag, we also print per-database information for anything we gather per-database (without that flag, you just get the summary).

Note that the EXPLAIN check is incomplete due to [this issue](https://github.com/pganalyze/pganalyze/issues/3851#issuecomment-2018888586); I think we can fix that later.

Thoughts?

Note that one big missing piece here is hints: I've added the code to collect them, but I'm not sure where to place them yet: inline right after the highlighted issue, if a hint exists? In a per-server footer section? At the end, after the summary output for all servers?

state/self_check.go
```diff
@@ -0,0 +1,262 @@
+package state
+
+import (
+	"fmt"
+	"sync"
+)
+
+// The self-check mechanism is intended to catch errors and potenttial problems
+// encountered during a test surface them to users in a test summary, and
+// communicate how these errors will impact the various features of pganalyze.
```
msakrejda, line 10: For code review, start here. Most of the other code just calls this code with various messages.

state/self_check.go
```diff
@@ -0,0 +1,262 @@
...
+		AspectStatuses:   make(map[CollectionAspect]*CollectionAspectStatus),
+		AspectDbStatuses: make(map[DbCollectionAspect](map[string]*CollectionAspectStatus)),
+	}
+}
+
+// collection suspended (e.g., if replica)
+func (s *SelfCheckStatus) MarkCollectionSuspended(format string, args ...any) {
+	if s == nil {
+		return
+	}
```
msakrejda, line 72: Checking for nil like this means the method can be invoked on a nil value of the right type, so we can easily disable the self-check without having to guard on `options.TestRun` everywhere.

msakrejda:
I think https://github.com/pganalyze/collector/commit/c8a7ef71b7c9d792e943ce6281142cfa601df36d failed because `self_test.go` is considered a test file due to Go's conventions, and was silently excluded from the build. I'll try with `selftest.go`.

keiko713:
Not looking at the code too closely, but overall I really like the direction and looking good!

> I originally wanted to replace the log output of the test run, but I realized that may not be a good idea (at least yet)

I agree with you and I like your approach for this one 👍 

> With the `--verbose` flag, we also print per-database information for anything we gather per-database (without that flag, you just get the summary).

Sounds good 👍 Is your screenshot the result with the verbose flag? I believe so, but just wanted to double check. Also I was gonna ask "what if one of the db name is long?!" and I just saw `printDbStatus`, well done 👏

> Note that the EXPLAIN check is incomplete due to https://github.com/pganalyze/pganalyze/issues/3851#issuecomment-2018888586; I think we can fix that later.

Is the "unknown; check pganalyze EXPLAIN Plans page" final version? Or you just put something temporary? I think that wording is a bit confusing, I don't know if unknown is my fault or not (and sounds like it's our fault). If that's the case, I think removing completely sounds better (and come back to this when we fix that).

> Note that one big missing piece here is hints: I've added the code to collect them, but I'm not sure where to place them yet: inline right after the highlighted issue, if a hint exists? In a per-server footer section? At the end, after the summary output for all servers?

Yeah this is difficult, I personally don't wanna break the flow of your current output (it's neat and tidy). How about creating a docs page dedicated to this collector test output, then list the hint there? Whenever there is an issue, we simply place the link to the doc (so that we won't be creating a wall of texts there, but printing a simple one line).

One more feedback: `printAllTestSummary` (and other related code) is currently added to main.go but I wonder if we can move this to somewhere else. The main.go is getting bigger and getting complicated to follow things there.

msakrejda:
Yeah, part of the cleanup item in that checklist is definitely getting that code out of `main.go`. It's not that easy, unfortunately, because Go doesn't like multiple top-level `main` package files, and when I tried to put this in a different package, I got a bunch of circular import issues. I've reorganized things a bit since, so I'll try again and I'm pretty sure we can make something like that work.

msakrejda:
@keiko713
>  Is your screenshot the result with the verbose flag? 

It is, yeah. Currently, non-verbose output looks something like this:
image

> Is the "unknown; check pganalyze EXPLAIN Plans page" final version? Or you just put something temporary? I think that wording is a bit confusing, I don't know if unknown is my fault or not (and sounds like it's our fault). If that's the case, I think removing completely sounds better (and come back to this when we fix that).

It's intended to be final-ish for this PR, but hopefully something I can fix in this sprint. I noted the reason for that in [this comment](https://github.com/pganalyze/pganalyze/issues/3851#issuecomment-2018888586). We could drop it, but @jensy suggested just dropping the "unknown"--maybe that's a reasonable compromise.

>How about creating a docs page dedicated to this collector test output, then list the hint there?

That could work, yeah. I think it'd be nice to give users something directly actionable in the collector test output, especially since we may have enough information at that point to give them very specific hints. But more troubleshooting info with more in-depth explanations could also be useful.

keiko713:
@msakrejda can you please share the example output with hint text now that you implemented that too?

input/system/system.go
```diff
@@ -33,30 +33,38 @@ func DownloadLogFiles(ctx context.Context, server *state.Server, globalCollectio
...
 		system.Info.Type = state.AivenSystem
+		server.SelfTest.MarkCollectionAspectNotAvailable(state.CollectionAspectSystemStats, "not available on this platform")
 	} else if config.SystemType == "tembo" {
-		system = tembo.GetSystemState(config, logger)
+		system = tembo.GetSystemState(server, logger)
 	} else if dbHost == "" || dbHost == "localhost" || dbHost == "127.0.0.1" || config.AlwaysCollectSystemData {
-		system = selfhosted.GetSystemState(config, logger)
+		system = selfhosted.GetSystemState(server, logger)
 	} else {
+		// system state is _available_ here, just not
```
keiko713, line 60: did you mean to say something more?

keiko713 (commented):
Haven't checked the main two files (selftest/summary.go and state/selftest.go) yet, but other files are looking good 🎉 I'm gonna review again tomorrow for these two files

input/system/system.go
msakrejda: Yes :grimacing:. Will fix.

msakrejda:
> @msakrejda can you please share the example output with hint text now that you implemented that too?

Yes. Some of the hint output is probably too verbose right now. I'm hesitant to hard-wrap it, because when combined with soft-wrapping on a terminal emulator, that can become pretty illegible, but maybe we should try to at least have terser hints.

msakrejda:
With just `--test`:
image
image

With `--test --verbose`:
image
image

Note the first result does not actually include the per-database results because we could not connect to the database in this case.

msakrejda:
Maybe our hints should all point to docs and just say

> See https://pganalyze.com/some-troubleshooting-docs for more info

?

keiko713:
Ok I tested locally (macOS + terminal) and here is an output, got several points/questions:

image

* At the end, it says "Test successful" message. Is it expected?
  * I don't have a strong opinion on if your new output should be at the end or it's okay to have "Test successful" at the end, but just wanted to double check if that's your intention
* The "Column stats" section is saying "ok in postgres" but yet it is being ❌  (my db_name is `pgaweb, *`)
* The "Extended stats" section is saying "could not check databases" but not sure why

I also ran on verbose mode, but didn't give me much info:

image

Regarding HINT, I think it's okay to ship as is actually. Though, I still think it's nice to have some docs page to talk about all the collector installation troubles (helpful for support perspective too).

msakrejda:
- I should probably drop the "Test successful" message (or at least its "view in pganalyze" links)
 - Huh. Do you see any errors in the collector logs?
 - Same as above.

msakrejda:
Also, I just pushed some fixes. Could you try the latest version? I don't think it will fix everything, but it might give us more info. I haven't been able to repro what you're seeing.

input/postgres/relation_column_stats.go
```diff
@@ -15,14 +16,19 @@ SELECT schemaname, tablename, attname, inherited, null_frac, avg_width, n_distin
...
-func GetColumnStats(ctx context.Context, logger *util.Logger, db *sql.DB, globalCollectionOpts state.CollectionOpts, systemType string, dbName string) (state.PostgresColumnStatsMap, error) {
+func GetColumnStats(ctx context.Context, logger *util.Logger, db *sql.DB, globalCollectionOpts state.CollectionOpts, systemType string, dbName string, server *state.Server) (state.PostgresColumnStatsMap, error) {
 	var sourceTable string
 
 	if StatsHelperExists(ctx, db, "get_column_stats") {
 		logger.PrintVerbose("Found pganalyze.get_column_stats() stats helper")
 		sourceTable = "pganalyze.get_column_stats()"
+		server.SelfTest.MarkDbCollectionAspectOk(dbName, state.CollectionAspectColumnStats)
 	} else {
 		if systemType != "heroku" && !connectedAsSuperUser(ctx, db, systemType) && globalCollectionOpts.TestRun {
```
keiko713, line 27: ah found it, I am apparently using the superuser so it wasn't coming to this condition, therefore dbs that don't have the helper wasn't getting "OK".
same goes to column stats

msakrejda: Ah cool, thanks!

msakrejda: Ok, I pushed a fix. Could you try again? If you get any errors, can you share the full output (feel free to redact anything, obviously)?

keiko713: The latest version works well! 🎉

keiko713 (approved):
Looks good!

msakrejda:
I'll go ahead and merge this, and we can iron out any kinks with Heroku customers as a sort of early-access, since they will typically use the `main` version.

PR #520, opened 2024-03-26
Change `TEMBO_METRICS_NAMESPACE` to `TEMBO_NAMESPACE`

ianstanton:


lfittl (approved):
As discussed with Ian 1:1, this change is done to improve the default name that gets shown in the pganalyze dashboard. Since the Tembo integration is not yet publicly released, this breaking change of the API system ID is safe to do.

PR #526, opened 2024-04-08
Update pg_stat_statements logic to match field names in Postgres 17

lfittl:
Whilst 17 is still in development, this change landed a while ago, so its likely to stay in, and I think we're okay to take this for now (since otherwise statement stats don't work on 17 at all).

### DIFF
--- a/input/postgres/statements.go
+++ b/input/postgres/statements.go
@@ -14,21 +14,27 @@ import (
 )
 
 // pg_stat_statements 1.3+ (Postgres 9.5+)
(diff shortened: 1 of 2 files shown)


msakrejda (approved):
Can we add 17 to integration tests? Is there a 17/devel base image we could build our integration image from?

msakrejda:
To answer my own question, there don't seem to be 17/devel images, so I think we can just wait until (beta?) images are available.

lfittl:
> To answer my own question, there don't seem to be 17/devel images, so I think we can just wait until (beta?) images are available.

Yeah, exactly - I think those are only available once packages are produced (i.e. with beta1).

PR #527, opened 2024-04-08
Update pg_stat_progress_vacuum field names for Postgres 17

lfittl:


protobuf/compact_activity_snapshot.proto
```diff
@@ -356,6 +356,10 @@ message VacuumProgressStatistic {
   int64 heap_blks_scanned = 4;
   int64 heap_blks_vacuumed = 5;
   int64 index_vacuum_count = 6;
-  int64 max_dead_tuples = 7;
-  int64 num_dead_tuples = 8;
+  int64 max_dead_item_ids = 7; // Postgres 16 and older
+  int64 num_dead_item_ids = 8;
```
lfittl, line 360: Note that we can change these names without breaking backwards compatibility, as long as the field numbers don't change.

seanlinsley (approved):

PR #529, opened 2024-04-10
Amazon RDS/Aurora IAM Authentication: Default to DB port 5432

lfittl:
Previously IAM authentication would fail with "PAM authentication failed" when the port was not explicitly set in the collector configuration.

This instead falls back to the default (5432), which is the typical port used on Amazon RDS and Aurora.

In passing use the same logic in the fallback logic for identifying RDS instances by their hostname, since the DB instance list also always includes the port, and this was likely broken as well if the port wasn't set explicitly.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -415,6 +415,15 @@ func (config ServerConfig) GetDbPort() int {
 	return config.DbPort
 }
 
+// GetDbPortOrDefault - Gets the database port from the given configuration, otherwise returns the default Postgres port (5432)
+func (config ServerConfig) GetDbPortOrDefault() int {
+	port := config.GetDbPort()
+	if port == 0 {
+		port = 5432
+	}
+	return port
+}
+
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -415,6 +415,15 @@ func (config ServerConfig) GetDbPort() int {
 	return config.DbPort
 }
 
+// GetDbPortOrDefault - Gets the database port from the given configuration, otherwise returns the default Postgres port (5432)
+func (config ServerConfig) GetDbPortOrDefault() int {
```
lfittl, line 419: Note for reviewer: The reason I made a separate function is that there is another caller of `GetDbPort` that is used for defining the self-managed system scope, which would be affected by changing the result from `0` to `5432`, so its best kept there: https://github.com/pganalyze/collector/blob/main/config/identify_system.go#L111

keiko713 (approved):

PR #531, opened 2024-04-11
Migrate collector README items to docs

msakrejda:
The official install documentation already covers most of the README,
but is more current. Drop the duplicate, out-of-date information to
avoid confusion.

- Drop log-based EXPLAIN helper docs
- Drop pg_read_file helper docs
- Drop reset_stat_statements docs
- Drop get_stat_replication docs
- Drop relation_stats_ext docs
- Drop get_column_stats docs
- Drop get_stat_activity docs
- Drop get_stat_statements docs
- Drop the restricted monitoring user docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,55 +139,6 @@ $$
 $$ LANGUAGE sql VOLATILE SECURITY DEFINER;
 ```
 
-Setting up Log EXPLAIN helper
------------------------------
-
-This section is relevant only if you use the `enable_log_explain`
-setting; if you use the recommended autoexplain extension, or if you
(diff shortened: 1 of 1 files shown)


msakrejda:
See also https://github.com/pganalyze/pganalyze-docs/pull/242

keiko713:
@msakrejda are you going to create a PR for the app too? I think there are several places mentioning these sections

msakrejda:
Yes. I should also update the collector hints to point to the docs, but I don't want to conflict with #519 so maybe I'll do that in a follow-up (in the meantime, those links will still point to the README, which will point to the docs).

keiko713 (approved):

PR #532, opened 2024-04-16
Update links that point to README to point to official docs instead

msakrejda:
Some of the README documentation was out of date, and it's hard to
keep these two copies in sync, so for a more seamless onboarding
experience, we'll focus our documentation on https://github.com/pganalyze/pganalyze-docs/

In passing, drop the get_stat_replication helper, which is no longer
necessary with pg_monitor.

### DIFF
--- a/input/postgres/replication.go
+++ b/input/postgres/replication.go
@@ -93,7 +93,7 @@ func GetReplication(ctx context.Context, logger *util.Logger, db *sql.DB, postgr
 	} else {
 		if systemType != "heroku" && !connectedAsSuperUser(ctx, db, systemType) && !connectedAsMonitoringRole(ctx, db) {
 			logger.PrintInfo("Warning: You are not connecting as superuser. Please setup" +
-				" the monitoring helper functions (https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user)" +
+				" the monitoring helper functions (https://pganalyze.com/docs/install/aiven/01_create_monitoring_user)" +
 				" or connect as superuser, to get replication statistics.")
 		}
 		sourceTable = "pg_stat_replication"
--- a/input/postgres/statements.go
+++ b/input/postgres/statements.go
@@ -186,9 +186,9 @@ func GetStatements(ctx context.Context, server *state.Server, logger *util.Logge
 			server.SelfTest.MarkCollectionAspectWarning(state.CollectionAspectPgStatStatements, "monitoring user may have insufficient permissions to capture all queries")
 			server.SelfTest.HintCollectionAspect(state.CollectionAspectPgStatStatements, "Please set up"+
 				" the monitoring helper functions (%s)"+
-				" or connect as superuser to get query statistics for all roles.", selftest.URLPrinter.Sprint("https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user"))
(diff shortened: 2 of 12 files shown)


input/postgres/replication.go
```diff
@@ -87,19 +85,7 @@ func GetReplication(ctx context.Context, logger *util.Logger, db *sql.DB, postgr
...
-		if systemType != "heroku" && !connectedAsSuperUser(ctx, db, systemType) && !connectedAsMonitoringRole(ctx, db) {
-			logger.PrintInfo("Warning: You are not connecting as superuser. Please setup" +
-				" the monitoring helper functions (https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user)" +
-				" or connect as superuser, to get replication statistics.")
-		}
-		sourceTable = "pg_stat_replication"
-	}
-
-	rows, err := db.QueryContext(ctx, QueryMarkerSQL+fmt.Sprintf(replicationStandbySQL, sourceTable))
+	rows, err := db.QueryContext(ctx, QueryMarkerSQL+replicationStandbySQL)
```
msakrejda, line 88: Hmm, I realized we don't actually want to do this, because Aiven still does not grant pg_monitor as far as I know. I'll revert this. I guess we should add a get_stat_replication helper page in the official docs.

input/postgres/statements.go
```diff
@@ -186,9 +186,9 @@ func GetStatements(ctx context.Context, server *state.Server, logger *util.Logge
 			server.SelfTest.MarkCollectionAspectWarning(state.CollectionAspectPgStatStatements, "monitoring user may have insufficient permissions to capture all queries")
 			server.SelfTest.HintCollectionAspect(state.CollectionAspectPgStatStatements, "Please set up"+
 				" the monitoring helper functions (%s)"+
-				" or connect as superuser to get query statistics for all roles.", selftest.URLPrinter.Sprint("https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user"))
+				" or connect as superuser to get query statistics for all roles.", selftest.URLPrinter.Sprint("https://pganalyze.com/docs/install/aiven/03_create_pg_stat_statements_helpers"))
```
msakrejda, line 189: This is currently only covered in the Aiven docs, since Aiven is the only provider we support that does not grant `pg_monitor`. Should we add a separate section for this in the docs?

input/postgres/replication.go
```diff
@@ -93,7 +93,7 @@ func GetReplication(ctx context.Context, logger *util.Logger, db *sql.DB, postgr
 	} else {
 		if systemType != "heroku" && !connectedAsSuperUser(ctx, db, systemType) && !connectedAsMonitoringRole(ctx, db) {
 			logger.PrintInfo("Warning: You are not connecting as superuser. Please setup" +
-				" the monitoring helper functions (https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user)" +
+				" the monitoring helper functions (https://pganalyze.com/docs/install/aiven/01_create_monitoring_user)" +
```
msakrejda, line 96: Same as below; only Aiven needs this helper now and we don't document it separately. I don't think we need to until we add another provider that does not support pg_monitor (which seems unlikely), but I could be convinced.

keiko713 (approved):

PR #533, opened 2024-04-18
Restore "Setting up a Restricted Monitoring User" section

msakrejda:
Use it to point to our official docs. Users may still be pointed here from, e.g., log output of old collector versions, so we can make it easier to find the right documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,12 @@ You can repeat the configuration block with a different `[name]` if you have mul
 See https://pganalyze.com/docs/install for further details.
 
 
+Setting up a Restricted Monitoring User
+---------------------------------------
+See https://pganalyze.com/docs/install/self_managed/01_create_monitoring_user (or the corresponding
+instructions for your platform) for details.
+
+
 Additional Setup
 ----------------
 If you are using the Buffer Cache report in pganalyze, you will also need to create this additional helper method:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,6 +41,12 @@ You can repeat the configuration block with a different `[name]` if you have mul
 See https://pganalyze.com/docs/install for further details.
 
 
+Setting up a Restricted Monitoring User
+---------------------------------------
+See https://pganalyze.com/docs/install/self_managed/01_create_monitoring_user (or the corresponding
+instructions for your platform) for details.
+
+
 Additional Setup
```
lfittl, line 50: I wonder if its also worth dropping the "Additional Setup" part of the README? We don't support the report functionality anymore for which this is intended, and it may be confusing when someone comes here looking for helper SQL to copy.

lfittl (approved):

README.md
msakrejda: Should we drop the reporting code as well to tidy things a bit? We can go to git history if we decide to revisit that functionality.

lfittl: Yeah, we probably should. If you'd like I can go ahead and rip that out (since I authored most of it I can probably find the spots more easily).

msakrejda: Sure, but no rush. In any case, I'll leave that to a future PR.

lfittl: Done that way: https://github.com/pganalyze/collector/pull/537 :)

PR #534, opened 2024-04-19
Update a config file location with Docker

keiko713:
Currently, `/home/pganalyze/.pganalyze_collector.conf` is used to store the config file when `CONFIG_CONTENTS` env var is passed, with Docker. However, this was failing to work when using helm chart, because `/home/pganalyze` is readonly.

With this PR, update that location to `/config/pganalyze-collector.conf`, then create such dir in Docker side. Also make sure to add this dir in helm chart side too, so that it'll be properly written within the docker entrypoint.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 FROM golang:1.21-alpine as base
 MAINTAINER team@pganalyze.com
+LABEL org.opencontainers.image.authors="team@pganalyze.com"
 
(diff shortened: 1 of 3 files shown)


lfittl (approved):
:+1: As discussed, this should be fine from a backwards compatibility perspective, since writing that file manually, e.g. in a Kubernetes context, wouldn't have worked anyway in the common setup, due the read-only file system problem.

PR #535, opened 2024-04-19
Add --generate-stats-helper-sql command to generate helpers psql install script

lfittl:
Passing the new "--generate-stats-helper-sql" command to the collector will generate a SQL script that can be passed to "psql" to install stats helpers (e.g. for collecting column stats) on all configured databases for the specified server.

The command takes the name of the configuration section of a specific server in the collector configuration (e.g. "server1").

It can be used manually by saving the output to a file, or by piping it to psql like this:

```
pganalyze-collector --generate-stats-helper-sql=server1 | psql -f -
```

Make sure to execute the generated SQL as an administrative user or database owner (not the pganalyze user!) in order for the SECURITY DEFINER functions to have sufficient privileges assigned.

### DIFF
--- a/main.go
+++ b/main.go
@@ -120,6 +120,34 @@ func run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
 		}
 	}
 
+	if globalCollectionOpts.GenerateHelperSql != "" {
+		wg.Add(1)
+		testRunSuccess = make(chan bool)
+		go func() {
+			var matchingServer *state.Server
+			for _, server := range servers {
+				if globalCollectionOpts.GenerateHelperSql == server.Config.SectionName {
+					matchingServer = server
+				}
+			}
+			if matchingServer == nil {
+				fmt.Fprintf(os.Stderr, "ERROR - Specified configuration section name '%s' not known\n", globalCollectionOpts.GenerateHelperSql)
+				testRunSuccess <- false
+			} else {
+				output, err := runner.GenerateHelperSql(ctx, matchingServer, globalCollectionOpts, logger.WithPrefix(matchingServer.Config.SectionName))
+				if err != nil {
+					fmt.Fprintf(os.Stderr, "ERROR - %s\n", err)
+					testRunSuccess <- false
+				} else {
+					fmt.Println(output)
+					testRunSuccess <- true
+				}
+			}
+			wg.Done()
+		}()
+		return
+	}
+
 	state.ReadStateFile(servers, globalCollectionOpts, logger)
 
 	writeStateFile = func() {
@@ -260,6 +288,7 @@ func main() {
 	var testRunLogs bool
 	var testExplain bool
 	var testSection string
+	var generateHelperSql string
 	var forceStateUpdate bool
 	var configFilename string
 	var stateFilename string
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -120,6 +120,34 @@ func run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
...
+		testRunSuccess = make(chan bool)
+		go func() {
+			var matchingServer *state.Server
+			for _, server := range servers {
+				if globalCollectionOpts.GenerateHelperSql == server.Config.SectionName {
+					matchingServer = server
+				}
+			}
+			if matchingServer == nil {
+				fmt.Fprintf(os.Stderr, "ERROR - Specified configuration section name '%s' not known\n", globalCollectionOpts.GenerateHelperSql)
```
lfittl, line 134: Note for reviewer: I ended up not using the regular logger here, so we don't accidentally pass error message to `psql` when piping.

runner/generate_helper_sql.go
```diff
@@ -0,0 +1,65 @@
...
+func GenerateHelperSql(ctx context.Context, server *state.Server, globalCollectionOpts state.CollectionOpts, logger *util.Logger) (string, error) {
+	db, err := postgres.EstablishConnection(ctx, server, logger, globalCollectionOpts, "")
+	if err != nil {
+		return "", err
+	}
+	defer db.Close()
+
+	version, err := postgres.GetPostgresVersion(ctx, logger, db)
+	if err != nil {
+		server.SelfTest.MarkCollectionAspectError(state.CollectionAspectPgVersion, err.Error())
```
msakrejda, line 44: I think we could skip this, since this is not really a normal part of the test flow, right?

msakrejda (approved):

runner/generate_helper_sql.go
lfittl: Good point!

runner/generate_helper_sql.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	databases, _, err := postgres.GetDatabases(ctx, logger, db, version)
+	if err != nil {
+		return "", fmt.Errorf("error collecting pg_databases: %s", err)
+	}
+
+	output := ""
+	for _, dbName := range postgres.GetDatabasesToCollect(server, databases) {
+		output += fmt.Sprintf("\\c \"%s\"\n", dbName)
```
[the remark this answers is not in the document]
lfittl: Yeah, let me see if I can utilize an existing function for that.

runner/generate_helper_sql.go
```diff
@@ -0,0 +1,65 @@
...
+		server.SelfTest.MarkCollectionAspectError(state.CollectionAspectPgVersion, err.Error())
+		return "", fmt.Errorf("error collecting Postgres version: %s", err)
+	}
+
+	databases, _, err := postgres.GetDatabases(ctx, logger, db, version)
+	if err != nil {
+		return "", fmt.Errorf("error collecting pg_databases: %s", err)
+	}
+
+	output := ""
```
keiko713, line 53: `strings.Builder` would be preferable here, e.g.

```
output := strings.Builder{}
...
output.WriteString(fmt.Sprintf("\\c \"%s\"\n", dbName))
...
return output.String(), nil
```

keiko713 (approved):
Maciek beat me already but still wanted to say something

runner/generate_helper_sql.go
lfittl: Nice, didn't know about `strings.Builder`!

PR #536, opened 2024-04-19
Release 0.56.0

keiko713:
This will be the first release with Helm chart CI release, let's see if it goes well (or not).


lfittl (approved):
:+1: I made some slight revisions directly, and will merge shortly.

PR #542, opened 2024-04-24
Handle newline within auto_explain query text with text format

keiko713:
We saw some cases where the explain output text is including the part of query for the case with Heroku logs when the new line is added in the middle of query text because query text is too long.
This is purely a workaround and I'm not even sure if this is a good workaround (probably bad), but this at least will fix the issue, at least with the customer who is seeing the issue.

I think the change is pretty safe. If the explain output is not starting with the capital letter (the first word should be a node type), I think that's a problem anyways, so trying something different won't hurt.
I glanced https://github.com/postgres/postgres/blob/89ad3e1316bf0374f9e57456bf2888d14bf070f6/src/backend/commands/explain.c#L1360 to see what the possible explain output to justify. Technically I'm not matching with "???" case but I think that's also another edge case that should be failing to parse anyways.

### DIFF
--- a/logs/querysample/querysample.go
+++ b/logs/querysample/querysample.go
@@ -11,6 +11,7 @@ import (
 	"github.com/pganalyze/collector/logs/util"
 	"github.com/pganalyze/collector/output/pganalyze_collector"
 	"github.com/pganalyze/collector/state"
+	cUtil "github.com/pganalyze/collector/util"
 )
 
 func TransformAutoExplainToQuerySample(logLine state.LogLine, explainText string, queryRuntime string) (state.PostgresQuerySample, error) {
@@ -65,7 +66,7 @@ func transformExplainTextToQuerySample(logLine state.LogLine, explainText string
 	// likely it's hitting the Heroku's newline break in "Query Text:" chunk
 	// Handle the separation of the query and the explain output text with the tab for these cases
 	explainOutputFirstChar := explainParts[2][0]
-	if !(explainOutputFirstChar >= 'A' && explainOutputFirstChar <= 'Z') {
+	if cUtil.IsHeroku() && !(explainOutputFirstChar >= 'A' && explainOutputFirstChar <= 'Z') {
 		if parts := herokuAutoExplainWithTabRegexp.FindStringSubmatch(explainText); len(parts) == 3 {
 			explainParts = parts
 		}
--- a/logs/querysample/querysample.go
+++ b/logs/querysample/querysample.go
@@ -54,11 +54,21 @@ func transformExplainJSONToQuerySample(logLine state.LogLine, explainText string
 }
 
 var autoExplainTextPlanDetailsRegexp = regexp.MustCompile(`^Query Text: (.+)\s+([\s\S]+)`)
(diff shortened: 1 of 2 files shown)


msakrejda (approved):
Nice, I think this is a reasonable workaround. I wonder if we should maybe limit it to only `util.IsHeroku`, though?

logs/querysample/querysample.go
```diff
@@ -54,12 +55,22 @@ func transformExplainJSONToQuerySample(logLine state.LogLine, explainText string
...
 func transformExplainTextToQuerySample(logLine state.LogLine, explainText string, queryRuntimeMs float64) (state.PostgresQuerySample, error) {
 	explainParts := autoExplainTextPlanDetailsRegexp.FindStringSubmatch(explainText)
 	if len(explainParts) != 3 {
 		return state.PostgresQuerySample{}, fmt.Errorf("auto_explain output doesn't match expected format")
 	}
+	// If EXPLAIN output's first char is not a capital letter (e.g. not something like "Update on" or "Index Scan"),
+	// likely it's hitting the Heroku's newline break in "Query Text:" chunk
+	// Handle the separation of the query and the explain output text with the tab for these cases
+	explainOutputFirstChar := explainParts[2][0]
+	if cUtil.IsHeroku() && !(explainOutputFirstChar >= 'A' && explainOutputFirstChar <= 'Z') {
```
andyatkinson, line 69: Wonder if it's worth trying to avoid `IsHeroku()` conditional checks from a maintenance perspective?

If you always checked for the text being in the expected character range, could the Heroku check be dropped?

e.g. flipping to:
```
(explainOutputFirstChar <= 'A' && explainOutputFirstChar >= 'Z'
```

keiko713: Thanks for the comment!
Ah yeah, we can flip that condition, though it's a bit different from what you suggested, more like the following:

```golang
explainOutputFirstChar < 'A' || explainOutputFirstChar > 'Z'
```

Reads like "if the first char is smaller than A or bigger than Z" then "do the workaround". Though maybe I misunderstand your suggestion, let me know in that case.

Regarding `IsHeroku()`, I thought limiting to it was a good idea to minimize changes to the currently working flow. I'd say this workaround is a bit hacky so I wanted to limit it.
That said, definitely open to adjust this. Have you observed this actually happens out side of Heroku before?

andyatkinson: Sounds good. Take my comments with a grain of salt. I didn't realize the collector is open source, which is cool to see! I agree I didn't properly invert your original condition. Added a comment above about that.

I've only used PgAnalyze with Heroku PostgreSQL so far, although I'm working with another client that's considering PgAnalyze, and they're on AWS RDS PostgreSQL.

Thanks!

keiko713: Appreciate the input!

> I've only used PgAnalyze with Heroku PostgreSQL so far, although I'm working with another client that's considering PgAnalyze, and they're on AWS RDS PostgreSQL.

Nice! In general we recommend setting up auto_explain in JSON format so these customers (e.g. RDS) won't hit this issue (won't touch this part of code). This is a bit of limitation that Heroku doesn't support JSON format + Heroku logs shape is a bit awkward. I'll remove `IsHeroku()` if we notice this happens on other providers and/or feel more confident on this workaround.

Thanks!

PR #544, opened 2024-04-30
Self-hosted: Allow collector to receive logs via OpenTelemetry

keiko713:
Successor of #503 

This is mainly intended to support Kubernetes, by using fluentbit with the opentelemetry output plugin to send pod logs to the collector. The logs received can optionally be filtered by pod name, or via label selectors. 

Note this requires the log output to be jsonlog (Postgres 15+), with optional K8s context added, or plain logs without any additional context.

When the collector is deployed to k8s using the helm chart, update `service` of the values.yaml to `create: true` so that the service of otel server will be created.

```
service:
  create: true
  name: pganalyze-collector-otel-service
  type: ClusterIP
  port: 4318
  targetPort: 4318
```

An example fluentbit output configuration:

```
  [OUTPUT]
    name  opentelemetry
    match kube.*postgres*
    host  pganalyze-collector-otel-service
    port  4318
```

With a corresponding collector configuration:

```
  db_log_otel_server = 0.0.0.0:4318
```

When utilizing CloudNativePG, a filter like the following can be used to filter only to the primary's logs for a given cluster:

```
  db_log_otel_k8s_labels = cnpg.io/cluster=cluster-example,cnpg.io/instanceRole=primary
```


keiko713:
I also created some draft of docs: https://pganalyze-we-add-selfho-wbo8ie.herokuapp.com/docs/log-insights/setup/opentelemetry

.gitignore
```diff
@@ -1,6 +1,6 @@
-pganalyze-collector
-pganalyze-collector-helper
-pganalyze-collector-setup
+/pganalyze-collector
+/pganalyze-collector-helper
+/pganalyze-collector-setup
```
keiko713, line 3: This was causing the `contrib/helm/pganalyze-collector` dir to be ignored. Adding `/` will enforce that only the root dir one is going to be ignored.

input/system/selfhosted/otel_handler.go
```diff
@@ -0,0 +1,210 @@
...
+		if len(parts) == 2 {
+			if k8sNamespaceName != parts[0] || k8sPodName != parts[1] {
+				return true
+			}
+		} else if len(parts) == 1 {
+			if k8sPodName != parts[0] {
+				return true
+			}
+		} else {
+			prefixedLogger.PrintWarning("Pod specification for OTel server not valid (need zero or one / separator): \"%s\", skipping log record\n", server.Config.LogOtelK8SPod)
```
msakrejda, line 112: Maybe we should put this in `config/read.go`? It seems silly to warn about this for every log line. I think we could also consider erroring out on this at startup, no?

msakrejda: Maybe we should also `SplitN` this while parsing, and set values on Config that don't correspond to config settings directly (one for name, one for namespace)? I doubt it's a bottleneck here, but that might be easier to follow. Lines 102-114 could become

```golang
if server.Config.LogOtelK8SPod != "" {
  if server.Config.LogOtelK8SPodNamespace != "" && server.Config.LogOtelK8SPodNamespace != k8sNamespaceName {
    return true
  }
  if server.Config.LogOtelK8SPod != k8sPodName {
    return true
  }
}
```

input/system/selfhosted/otel_handler.go
```diff
@@ -0,0 +1,210 @@
...
+					if (selEq && v != selValue) || (selNotEq && v == selValue) {
+						return true
+					}
+				}
+			} else {
+				prefixedLogger.PrintWarning("Label selector for OTel server not valid: \"%s\", skipping log record\n", server.Config.LogOtelK8SLabels)
+				return true
+			}
+		}
+	}
```
msakrejda, line 137: Similar to above, I think it'd be better parse this into a structured form when reading the config, and make it easier to work with for consumers.

input/system/selfhosted/otel_handler.go
```diff
@@ -0,0 +1,210 @@
...
+								logLine, detailLine := logLineFromJsonlog(record, tz)
+								if skipDueToK8sFilter(kubernetes, server, prefixedLogger) {
+									continue
+								}
+
+								parsedLogStream <- state.ParsedLogStreamItem{Identifier: server.Config.Identifier, LogLine: logLine}
+								if detailLine != nil {
+									parsedLogStream <- state.ParsedLogStreamItem{Identifier: server.Config.Identifier, LogLine: *detailLine}
+								}
+							} else if logger == "" && hasErrorSeverity {
```
msakrejda, line 188: Why are we requiring `error_severity`? It might be good to add a comment here or above.

msakrejda (commented):
Some comments, but looks good (with the caveat that I am not familiar with otel logs). Does this work with `--test` and `--test-logs`? If not, what's our plan there?

keiko713:
Thanks for the review, @msakrejda !

> Does this work with `--test` and `--test-logs`? If not, what's our plan there?

No it does not work, and the plan is to support with the branch that I have in my local machine 😬 (aka planning to do that as a follow-up)

contrib/helm/pganalyze-collector/README.md
```diff
@@ -46,6 +46,11 @@ pganalyze statistics collector
 | securityContext.runAsNonRoot | bool | `true` |  |
 | securityContext.runAsUser | int | `1000` |  |
 | securityContext.seccompProfile.type | string | `"RuntimeDefault"` |  |
+| service.create | bool | `false` | Specifies whether a service should be created for receiving logs via OpenTelemtry This service is used when Postgres is running within the cluster and Postgres logs are sent out to the collector using log collectors like Fluent Bit |
+| service.name | string | `"pganalyze-collector-otel-service"` | The name of the service to use. If not set and create is true, a name is generated using the fullname template. This is the name referenced by the log sender like Fluent Bit |
+| service.port | int | `4318` | The port of service. This is the port referenced by the log sender like Fluent Bit |
+| service.targetPort | int | `4318` | The target port of the log OTEL server port. This should match to the port number specified with db_log_otel_server |
+| service.type | string | `"ClusterIP"` | The type of service to create. |
```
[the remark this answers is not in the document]
keiko713: Yeah, it's k8s terminology. In k8s, a Service is an abstraction which defines a logical set of Pods and a policy by which to access them, often over a network.
Services in k8s support different types, primarily: ClusterIP (default, providing internal cluster connectivity), NodePort (exposes the Service on each Node’s IP at a static port), and LoadBalancer (provides a fixed, external IP assigned by the cloud provider and routes traffic to the NodePort or ClusterIP).
Here, I'm putting "ClusterIP" as a sample/default value and it'll work when you put the Postgres, fluent-bit, and the collector in the same k8s cluster (and this helm chart is mainly assuming that it'll be the case, otherwise they'll need to make more changes).

input/system/selfhosted/otel_handler.go
keiko713: Good question. This part is written by Lukas and actually don't have good context. This is about the following part:

> Note this requires the log output to be jsonlog (Postgres 15+), with optional K8s context added, or plain logs without any additional context.


It is "plan logs without any additional context". I actually haven't tested this case. I think `error_severity` is somewhat "must be there" JSON key as a JSON format logs, therefore if that key exists, we can assume that it's a Postgres log? @lfittl does it sound right? 
https://www.postgresql.org/docs/current/runtime-config-logging.html#RUNTIME-CONFIG-LOGGING-JSONLOG

input/system/selfhosted/otel_handler.go
keiko713: Thanks for the idea! I refactored this, and also added the test for the label/selector matching part as it was actually in TODO of the original PR. Ideally we want to use the k8s code for this I'd say, but that'll involve lots of pulling deps so I decided not to do that at least in this PR.

input/system/selfhosted/otel_handler.go
keiko713: (I updated the comment assuming that it's correct, but happy to update)

util/kubernetes.go
```diff
@@ -0,0 +1,29 @@
+package util
+
+import (
+	"regexp"
+)
+
+var K8sSelectorRegexp = regexp.MustCompile(`\s*([^!=\s]+)\s*([!=]+)\s*([^\s]+)\s*`)
+
+// CheckLabelSelectorMismatch checks if selectors do not match the given labels
+func CheckLabelSelectorMismatch(labels map[string]string, selectors []string) bool {
```
msakrejda, line 10: Is this actually called anywhere (other than the tests)? I didn't see any call sites.

It might also be clearer to invert the condition here (i.e., `CheckLabelSelectorMatch` or `CheckLabelSelectorMatches`) and then negate it in the caller.

Separately, since we're dealing with multiple selectors and multiple labels, it might be good to include the full expected behavior (i.e., if I'm reading this correctly, that every selector must match its corresponding label, if any) in the comment.

util/kubernetes.go
```diff
@@ -0,0 +1,29 @@
...
+	// Potentially refactor this and selector regexp with
+	// https://github.com/kubernetes/apimachinery/blob/master/pkg/labels/selector.go
+	for _, selector := range selectors {
+		parts := K8sSelectorRegexp.FindStringSubmatch(selector)
+		if parts != nil {
+			selKey := parts[1]
+			selEq := parts[2] == "=" || parts[2] == "=="
+			selNotEq := parts[2] == "!="
```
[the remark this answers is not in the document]
keiko713: Actually I wrote this in the draft PR of docs update too, but it's based on "Comma-separated [Kubernetes label selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) to filter matching logs" (I actually need to update this and mention that it only supports equality-based selector)
https://pganalyze-we-add-selfho-wbo8ie.herokuapp.com/docs/collector/settings#self-managed-servers

Per their spec:

> Three kinds of operators are admitted =,==,!=. The first two represent equality (and are synonyms), while the latter represents inequality.

So this is the correct implementation. Though the regex indeed matches `!==` too, I'll update the regex from `([!=]+)` to `(=|==|!=)` to make this clear. (currently it technically matches to things like `=====` too, so better to be explicit)

util/kubernetes.go
keiko713: > Is this actually called anywhere (other than the tests)? I didn't see any call sites.

No, `CheckLabelSelectorMismatch` is not called (though `K8sSelectorRegexp` is called, and that's the reason behind of making this util). I'd like to use the k8s code for this and refactor, so that I can retire this whole util. As this was a good chunk, I decided to keep them together (also makes easier to write tests focusing on this).

> It might also be clearer to invert the condition here (i.e., `CheckLabelSelectorMatch` or `CheckLabelSelectorMatches` or maybe just `SelectorsMatchLabels`) and then negate it in the caller.

I thought about it, but the nice part of the current logic is that it can early return easily when something doesn't match (also logic is a bit simpler). Well, given neither labels and selectors won't be that long, maybe I don't need to worry about too much for the performance?

> Separately, since we're dealing with multiple selectors and multiple labels, it might be good to include the full expected behavior (i.e., if I'm reading this correctly, that every selector must match its corresponding label, if any) in the comment.

I'll point to k8s label selector docs (that I linked in the other comment) 👍 FYI, the short answer to this is the following:

> In the case of multiple requirements, all must be satisfied so the comma separator acts as a logical AND (&&) operator.

input/system/selfhosted/otel_handler.go
msakrejda: Did you push up those changes? I don't see them in this branch.

keiko713: Wow, I swear I did update this (well I did everything to prepare this, which is already pushed), but I don't see them too 🙃 Now I'm modifying, maybe I did in fact forget to update this part 🤦 Thanks for noticing! Update pushed 👍

msakrejda (approved):

PR #547, opened 2024-05-11
AWS: Allow setting both assume role and web identity/role ARN

lfittl:
In a situation where both the web identity token file, role ARN, and assume role ARN are set, we now first retrieve credentials via web identity, and then assume the role specified as AWS_ASSUME_ROLE.

This helps with cross-account configurations on AWS in combination with the collector running in EKS.

## TODO

* [x] Test that this works as expected
* [x] Think through whether there are any situations where this change is not desired (e.g. does AWS_ASSUME_ROLE get set by EKS automatically in some way?)
  * AWS only uses `AWS_ROLE_ARN`, not `AWS_ASSUME_ROLE`, so we're safe in that regard

### DIFF
--- a/util/awsutil/amazon.go
+++ b/util/awsutil/amazon.go
@@ -76,10 +76,23 @@ func GetAwsSession(cfg config.ServerConfig) (*session.Session, error) {
 		if err != nil {
 			return nil, err
 		}
-		if cfg.AwsAssumeRole != "" {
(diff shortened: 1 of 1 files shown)


msakrejda (approved):
Haven't tried it, but code looks good.

keiko713 (approved):
I've tested with eks and worked 🎉

PR #550, opened 2024-05-31
Work around RDS pg_stats and pg_stats_ext issue

msakrejda:
It looks like RDS has some custom patches that can run the permission
check for stats before running the related filter, causing a
permission failure fetching stats for tables that we don't want to see
in the first place (and that should be pruned by the `schemaname`
filter).

Ignore the 'pg_subscription' table explicitly.


lfittl (approved):
I think that (unfortunately) makes sense, based on how that filter is implemented for RDS (and the number of reports we've gotten). We should also update this in the docs.

msakrejda:
Yep, just wanted to confirm this is the right change first. I'll draft a docs patch.

PR #551, opened 2024-06-19
Release 0.57.0

keiko713:
I think we have enough changes to make a new release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Changelog
 
+## 0.57.0      2024-06-19
+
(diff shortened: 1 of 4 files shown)


msakrejda (approved):

seanlinsley:
Maybe we could also include https://github.com/pganalyze/collector/pull/555 and https://github.com/pganalyze/collector/pull/556?

lfittl:
We may also want to pull in #547 - just ran into this again with a different customer.

keiko713:
I'm planning to make a release tomorrow, let's see how many we can squeeze
We can also always make a new patch release too

PR #553, opened 2024-06-26
Collect statistics from pg_stat_io (Postgres 16+)

seanlinsley:
https://www.postgresql.org/docs/current/monitoring-stats.html#MONITORING-PG-STAT-IO-VIEW

In a future PR we'll want to handle `stats_reset`, but that's been skipped for now.


input/postgres/server_stats.go
```diff
@@ -69,8 +69,23 @@ COALESCE((
...
+	   object,
+	   context,
+	   reads,
+	   writes,
+	   extends,
+	   op_bytes,
+	   evictions,
+	   reuses,
+	   fsyncs
+  FROM pg_stat_io
```
lfittl, line 83: @seanlinsley We should probably add the `_time` fields as well here - AFAIK they didn't exist yet when I first prototyped this, but were added later (and are very useful to understand whether its likely the OS actually issued an I/O request to disk, vs getting something from the page cache).

We could probably also add `writebacks` and `writeback_time` (https://www.postgresql.org/docs/current/monitoring-stats.html#MONITORING-PG-STAT-IO-VIEW) - its a bit specialized, but potentially useful.

runner/diff.go
```diff
@@ -195,3 +196,25 @@ func diffDatabaseStats(new state.PostgresDatabaseStatsMap, prev state.PostgresDa
...
+		var s state.DiffedPostgresServerIoStats
+		prevStats, exists := prev[k]
+		if exists {
+			s = stats.DiffSince(prevStats)
+		} else if followUpRun { // New since the last run
+			s = stats.DiffSince(state.PostgresServerIoStats{})
+		}
+		// Skip over empty diffs (which can occur either because there was no activity, or for fixed entries that never saw activity)
+		if s.Reads != 0 || s.Writes != 0 || s.Extends != 0 ||
+			s.Evictions != 0 || s.Reuses != 0 || s.Fsyncs != 0 {
```
lfittl, line 214: Not a big deal, but we could add Writebacks and `Hits` here as well (it seems very unlikely that that'd be the only activity on a system, but easier to be consistent here for a future reader of the code).

lfittl (commented):
Looks good!

seanlinsley:
Changes since the previous iteration of this PR:
- `backend_type` is no longer a string, instead it reuses the `BackendCountStatistic.BackendType` enum
- `io_object` and `io_context` are no longer strings, instead using new enums
- The protobuf struct is longer prefixed with `Historic`. AFAIK that's a namign pattern used to distinguish between the many Go structs needed for diffing, but there's no reason to expose that internal complexity to the protobuf structure.
- `collected_interval_secs` has been renamed to `collected_secs`. I feel that communicates the same intent with fewer words, and I personally find "interval" confusing in this context.

output/transform/postgres.go
```diff
@@ -167,6 +170,83 @@ func transformPostgresServerStats(s snapshot.FullSnapshot, newState state.Persis
...
+			case "checkpointer":
+				stat.BackendType = snapshot.BackendCountStatistic_CHECKPOINTER
+			case "startup":
+				stat.BackendType = snapshot.BackendCountStatistic_STARTUP
+			case "walreceiver":
+				stat.BackendType = snapshot.BackendCountStatistic_WALRECEIVER
+			case "walsender":
+				stat.BackendType = snapshot.BackendCountStatistic_WALSENDER
+			case "walwriter":
+				stat.BackendType = snapshot.BackendCountStatistic_WALWRITER
```
lfittl, line 223: We're missing the "slotsync worker" backend type here, that was added in Postgres 17: https://www.postgresql.org/docs/current/logicaldecoding-explanation.html#LOGICALDECODING-REPLICATION-SLOTS-SYNCHRONIZATION

(I think to be able to account for all I/O its helpful to include it, even if an edge case)

runner/full_1min.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-func gatherQueryStatsForServer(ctx context.Context, server *state.Server, globalCollectionOpts state.CollectionOpts, logger *util.Logger) (state.PersistedState, error) {
+func gather1minStatsForServer(ctx context.Context, server *state.Server, globalCollectionOpts state.CollectionOpts, logger *util.Logger) (state.PersistedState, error) {
```
lfittl, line 15: The `1min` in the function name jumped out a bit at me at first glance, but not sure there is a better name here - using something like "high frequency" in the name is probably more confusing than helpful.

runner/full_1min.go
```diff
@@ -51,33 +51,45 @@ func gatherQueryStatsForServer(ctx context.Context, server *state.Server, global
...
+	}
+
 	// Don't calculate any diffs on the first run (but still update the state)
 	if len(server.PrevState.StatementStats) == 0 || server.PrevState.LastStatementStatsAt.IsZero() {
 		return newState, nil
 	}
 
-	diffedStatementStats := diffStatements(newState.StatementStats, server.PrevState.StatementStats)
-	collectedIntervalSecs := uint32(newState.LastStatementStatsAt.Sub(server.PrevState.LastStatementStatsAt) / time.Second)
+	timeKey := state.PostgresStatementStatsTimeKey{
```
lfittl, line 64: I wonder if we should rename the `PostgresStatementStatsTimeKey` struct to `HistoricStatsTimeKey` or something, so its clear its not specific to statements?

lfittl (commented):
Overall looks good - I want to do one more test together with the storage PR, but left a few minor notes in the meantime.

lfittl (approved):
:+1:

lfittl (approved):
:+1:

PR #556, opened 2024-07-01
Enable log filtering by default

seanlinsley:
To reduce the likelihood that logs contain passwords, this PR enables log filtering by default for credentials, parsing errors, and unidentified log lines.

Any users that change this setting need to take care to re-enable these filters, so for example if they wanted to add `statement_parameter` filtering they should set `statement_parameter,credential,parsing_error,unidentified`.

### DIFF
--- a/config/read.go
+++ b/config/read.go
@@ -318,6 +318,8 @@ func getDefaultConfig() *ServerConfig {
 	}
 	if filterLogSecret := os.Getenv("FILTER_LOG_SECRET"); filterLogSecret != "" {
 		config.FilterLogSecret = filterLogSecret
+	} else {
+		config.FilterLogSecret = "credential,parsing_error,unidentified"
 	}
 	if filterQuerySample := os.Getenv("FILTER_QUERY_SAMPLE"); filterQuerySample != "" {
 		config.FilterQuerySample = filterQuerySample


config/read.go
```diff
@@ -318,6 +318,8 @@ func getDefaultConfig() *ServerConfig {
 	}
 	if filterLogSecret := os.Getenv("FILTER_LOG_SECRET"); filterLogSecret != "" {
 		config.FilterLogSecret = filterLogSecret
+	} else {
+		config.FilterLogSecret = "credential,parsing_error,unidentified"
```
lfittl, line 322: I think we should also include `table_data` here - this information isn't typically necessary, and more likely to be forgotten.

That essentially means our defaults are to include statement texts, statement parameters (which are often useful, and whether they are sensitive or not depends on the system) and operational information (which is safe to see in the typical configuration), but exclude other kinds of secrets.

lfittl (approved):

config/read.go
seanlinsley: This PR is aimed at making sure database passwords aren't stored in the logs.

I don't think it makes sense to filter table data by default unless statement parameters were also filtered out, since both contain user data. It seems like we should leave that decision up to customers.

lfittl: I'd argue that we're trying to adjust our defaults here more broadly (not just in the context of credentials accidentally logged via syntax errors/etc), and I don't see a reason not to pick a reasonable default from the ground up.

I don't think many customers need the actual table data from the log events - and many would be surprised to see that data being logged by Postgres in the first place.

@msakrejda @keiko713 Any thoughts on this in terms of what's a sensible default?

seanlinsley: The table data in logs is arguably useful information. For example, knowing the exact keys causing a duplicate key constraint error could help you fix the issue faster.

I think conceptually it only makes sense to filter table data by default if it's somehow higher risk than statement parameters, or if we were also going to filter statement parameters by default.

Customers who want the most out of pganalyze are okay with a random sample of their data being in the logs. Customers who want maximum security / data privacy will set `filter_log_secret: all`

msakrejda: Hmm. I think this is tricky. Basically we're trying to manage a trade-off between three aspects of the default setting here:
 1. safe, preventing us from collecting any sensitive customer data accidentally
 2. usable, allowing our system to present useful information to users (and use it for analysis)
 3. consistent, making it easier to reason about the default

I do think (3) is a reasonable concern: there's undoubtedly a bunch of workloads like `INSERT INTO x VALUES($1, $2, $3)`, so it is kinda weird to filter these when they're parameters, but not right after the insert, when they become table data. On the other hand, _most_ inserts are very likely to be under the slow query logging threshold.

Looking at what's flagged as table data in [analyze.go](https://github.com/pganalyze/collector/blob/main/logs/analyze.go), I do think it's worth filtering this out by default: some of the table data errors only show the primary key, but several show the entire row.
```
maciek=# create table t(a int not null, b text);
CREATE TABLE
maciek=# insert into t(a,b) values(null, 'super secret');
ERROR:  null value in column "a" of relation "t" violates not-null constraint
DETAIL:  Failing row contains (null, super secret).
```
This or check constraints seem like they could be fairly common errors.

And while they're similar to table data, I do agree that statement parameters are probably too useful to filter out by default. Especially since most inserts (and probably updates) will not hit the slow query threshold.

So while it feels a little unprincipled to give up on consistency, it's reasonable to add `table_data` but not `statement_parameter` to the list. While technically `table_data` and `statement_parameter` are the same category of risk, I think pragmatically, `table_data` is more risky (and far less useful in this context).

That's my two cents.



Aside 1: we should really document `filter_log_secret` better: we document what values are allowed and they're sort of self-documenting, but I think it's pretty hard to tell what exactly this code is doing.

Aside 2: how do we update the list in `analyze.go`? How do we know when new log messages are introduced that we should be filtering, or when existing log messages are updated so that our regular expressions no longer match?

seanlinsley: I still think that table data is useful for debugging, and since pganalyze makes it easy to filter logs by category it's actually possible as an application developer to identify and diagnose not-null and unique constraint errors. Filtering this data out by default reduces the usefulness of Log Insights, and (as I argued above) is out of scope for this PR.

keiko713: Thanks all for answering my questions of "what does X mean" (esp. Maciek for providing an example).

> For example, knowing the exact values causing a duplicate key constraint error could help you fix the issue faster.

Interestingly, our prod collector has `filter_log_secret: credential, parsing_error, table_data` (and staging collector has `filter_log_secret: credential, parsing_error, table_data, statement_text, unidentified`), which is a good case to check out how `table_data` filtering is like.

I just checked our prod to see how the duplicate key constraint error looks like (as we have many, `V100` ones). Since the prod doesn't have `statement_text` in the filtering, but this error does have `STATEMENT`, it looked a bit funny and the `DETAIL` looked like `DETAIL: Key (name)=(XXXXXXXXXXXX) already exists.` where `STATEMENT` has all raw data.
So, for this specific case, having `table_data` wasn't preventing from getting the necessary debug data (one of Sean's concern), however, that also means that it's not hiding data much for certain cases which look inconsistent for what we do. There is a case that doesn't look so inconsistent (like Maciek's example case), though.

I'm honestly indifferent and we can go either way of having or not having `table_data`.


Outside of that, I am having a slight opinion for "unidentified". It is good to include "unidentified" (as we don't know what's in there) but I fear that the all Z00 category now becomes `XXXXXX` and no idea what these logs are. I wish we had a special category for "LOG: connection authenticated" (actually just realized that we do have it, then why all of our Z00 logs are not properly categorized as C21 🤔).

seanlinsley: > I just checked our prod to see how the duplicate key constraint error looks like (as we have many, `V100` ones). Since the prod doesn't have `statement_text` in the filtering, but this error does have `STATEMENT`, it looked a bit funny and the `DETAIL` looked like `DETAIL: Key (name)=(XXXXXXXXXXXX) already exists.` where `STATEMENT` has all raw data.

That may need a similar adjustment to #555 so `table_data` also redacts the statement.

>Outside of that, I am having a slight opinion for "unidentified". It is good to include "unidentified" (as we don't know what's in there) but I fear that the all Z00 category now becomes XXXXXX and no idea what these logs are.

Maybe Lukas can chime in here, but I think the idea is we might temporarily disable `unidentified` log filtering when a new Postgres version is released (on staging or in local development) in order to add parsing for new log types added.

lfittl: > I just checked our prod to see how the duplicate key constraint error looks like (as we have many, `V100` ones). Since the prod doesn't have `statement_text` in the filtering, but this error does have `STATEMENT`, it looked a bit funny and the `DETAIL` looked like `DETAIL: Key (name)=(XXXXXXXXXXXX) already exists.` where `STATEMENT` has all raw data. So, for this specific case, having `table_data` wasn't preventing from getting the necessary debug data (one of Sean's concern), however, that also means that it's not hiding data much for certain cases which look inconsistent for what we do. There is a case that doesn't look so inconsistent (like Maciek's example case), though.
 
Hmm, yeah, that's an interesting point.

The cases I was thinking of where including `table_data` in the default filter makes a difference (but not the `statement_` filters) is where someone consciously utilizes bind parameters to send the INSERT data out of band, but I suppose even in those cases a slow INSERT would have its parameters logged, which is controlled by the `statement_parameter` filter (not `table_data`).

I guess what we should be doing here is specifically normalizing `VALUES` expressions when `table_data` is on, but nothing else in the statement. Or normalizing the whole statement (and masking parameters) if the statement is a write statement (which would have to account for wCTEs).

> I'm honestly indifferent and we can go either way of having or not having `table_data`.

Given that you've shown that even for our own systems this doesn't really address the concern in different edge cases, I suspect its better if we keep this as is in the PR for now, and clearly document that both `table_data` and `statement_` filters are needed to filter out all potentially sensitive information from the logs.

I still feel that we may want to include `table_data` in a future revision of this filter, but unless we close the gap noted above, it may provide a false sense of security.

> Outside of that, I am having a slight opinion for "unidentified". It is good to include "unidentified" (as we don't know what's in there) but I fear that the all Z00 category now becomes `XXXXXX` and no idea what these logs are. I wish we had a special category for "LOG: connection authenticated" (actually just realized that we do have it, then why all of our Z00 logs are not properly categorized as C21 🤔).

Yeah, there isn't really a way around this - `unidentified` could include new types of syntax / etc errors, so the only way to be consistent with "filter anything by default that could be credential related" is to filter out unidentified lines as well.

If needed for a specific customer question we could ask for running the original logs through the `--analyze-logfile` option (which will emit all unidentified lines).

> Aside 1: we should really document filter_log_secret better: we document what values are allowed and they're sort of self-documenting, but I think it's pretty hard to tell what exactly this code is doing.

Agreed, I think we should probably have a dedicated documentation page for it (not just document it inline on the settings page, like we do currently).

> Aside 2: how do we update the list in analyze.go? How do we know when new log messages are introduced that we should be filtering

That's really where `unidentified` comes in, and why I think adding this to the default filter makes sense - essentially an unknown log line has to be treated as any of the possible secret kinds.

> or when existing log messages are updated so that our regular expressions no longer match?

Typically that has caused them to become `unidentified`, at least in the example that I can think of (autovacuum logs), since the regular expressions are written in a narrow enough way that e.g. new data added at the end should cause the regexp to fail to match.

lfittl: @seanlinsley To clarify, given the above discussion, whilst I'm still inclined to push for including `table_data` here, I feel that to make that work we need to do more (i.e. to cover the other cases where insert values show up), and so I'm okay with deferring that decision, and going with the filter proposed here as is for now.

lfittl: > > Outside of that, I am having a slight opinion for "unidentified". It is good to include "unidentified" (as we don't know what's in there) but I fear that the all Z00 category now becomes XXXXXX and no idea what these logs are.
> 
> Maybe Lukas can chime in here, but I think the idea is we might temporarily disable `unidentified` log filtering when a new Postgres version is released (on staging or in local development) in order to add parsing for new log lines.

I think for our own systems we can continue to allow `unidentified` on production - its not like there is a big risk that there is anything credentials-related in there (its just a chance, and a small one at that, especially on a DBaaS provider).

PR #558, opened 2024-07-02
Add workaround for CentOS 7

keiko713:
Currently, build_packages CI is failing because of the EOL of CentOS 7 and the retirement of its mirror for packages.
https://serverfault.com/questions/1161816/mirrorlist-centos-org-no-longer-resolve

I've placed several workarounds, seems to be working locally. Until we have a better plan for how to retire CentOS 7 support, maybe we can use this workaround?

### DIFF
--- a/packages/repo/Dockerfile.repo-rpm
+++ b/packages/repo/Dockerfile.repo-rpm
@@ -1,7 +1,5 @@
 FROM amazonlinux:2
 
-# Replace with ARM base image when building on ARM
-
 # Build arguments
 ARG VERSION
 ENV NAME pganalyze-collector
--- a/packages/repo/Dockerfile.repo-rpm
+++ b/packages/repo/Dockerfile.repo-rpm
@@ -1,7 +1,6 @@
-FROM centos:7
+FROM amazonlinux:2
 
-# Replace with ARM base image when building on ARM:
-#FROM arm64v8/centos:7
+# Replace with ARM base image when building on ARM
 
 # Build arguments
 ARG VERSION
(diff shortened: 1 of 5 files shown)


packages/src/Dockerfile.build.rpm-systemd
```diff
@@ -8,10 +8,16 @@ ENV PATH $PATH:/usr/local/go/bin
...
+# Workaround for mirrorlist.centos.org
+# https://serverfault.com/questions/1161816/mirrorlist-centos-org-no-longer-resolve
+RUN sed -i s/mirror.centos.org/vault.centos.org/g /etc/yum.repos.d/*.repo
+RUN sed -i s/^#.*baseurl=http/baseurl=http/g /etc/yum.repos.d/*.repo
+RUN sed -i s/^mirrorlist=http/#mirrorlist=http/g /etc/yum.repos.d/*.repo
+
 # Packages required for both building and packaging
 #
 # This uses Ruby 3.0 since 2.7 (which still used for Debian builds) does not work with fpm: https://github.com/jordansissel/fpm/issues/2048#issuecomment-1972196104
-RUN yum install -y centos-release-scl scl-utils tar make git rpmdevtools gcc && yum install -y rh-ruby30 rh-ruby30-ruby-devel
```
keiko713, line 14: While I was testing, I noticed that running `yum install` twice is not working and second time always fail.
I believe that this is because files in `/etc/yum.repos.d/` are modified with the first run of `yum install` and the workaround will no longer be there.
Making this as one command worked.

packages/test/Makefile
```diff
@@ -9,7 +9,8 @@ docker_test_and_clean = docker exec pga-collector-test /root/systemd_test.sh &&
 
 # Note: The default list excludes rockylinux9 since there is an open issue with
 # /sbin/init not being present, see https://github.com/rocky-linux/sig-cloud-instance-images/issues/39
-DISTROS=centos7 rhel8 rockylinux8 rhel9 fedora36 fedora37 amazonlinux2 amazonlinux2023 ubuntu-focal ubuntu-jammy ubuntu-noble debian-bullseye debian-bookworm
+# Note: The default list also excludes centos7 since it's EOL, we will drop the support near future
```
keiko713, line 12: I could try to support centos7 here but felt like too much work, so I simply dropped this.

lfittl (approved):

packages/repo/Dockerfile.repo-rpm
```diff
@@ -1,7 +1,6 @@
-FROM centos:7
+FROM amazonlinux:2
 
-# Replace with ARM base image when building on ARM:
-#FROM arm64v8/centos:7
+# Replace with ARM base image when building on ARM
```
lfittl, line 3: I think we can actually drop that comment now, because `amazonlinux:2` has packages for both ARM and x86 on the same name (see https://hub.docker.com/_/amazonlinux/tags).

PR #559, opened 2024-07-03
Address docker warnings and GH Action Node.js 16 deprecation

keiko713:
I'm not addressing all the warnings for both Docker and GH Action, but we have too many now and I'm addressing some.
For GH Action version bump, I made sure that every bump is pretty safe. The `actions/upload-artifact` action had some breaking changes (which didn't affect to us), but otherwise all bumps are pretty much to support Node.js 16 deprecation.

https://github.com/pganalyze/collector/actions/runs/9770203430

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,9 @@
-FROM golang:1.21-alpine as base
-MAINTAINER team@pganalyze.com
+FROM golang:1.21-alpine AS base
 LABEL org.opencontainers.image.authors="team@pganalyze.com"
 
-ENV GOPATH /go
-ENV HOME_DIR /home/pganalyze
-ENV CODE_DIR $GOPATH/src/github.com/pganalyze/collector
+ENV GOPATH=/go
+ENV HOME_DIR=/home/pganalyze
+ENV CODE_DIR=$GOPATH/src/github.com/pganalyze/collector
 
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,10 +1,9 @@
-FROM golang:1.21-alpine as base
-MAINTAINER team@pganalyze.com
```
msakrejda, line 2: For other reviewers, this was [superseded by LABEL](https://docs.docker.com/reference/dockerfile/#maintainer-deprecated).

Dockerfile
```diff
@@ -1,10 +1,9 @@
...
+FROM golang:1.21-alpine AS base
 LABEL org.opencontainers.image.authors="team@pganalyze.com"
 
-ENV GOPATH /go
-ENV HOME_DIR /home/pganalyze
-ENV CODE_DIR $GOPATH/src/github.com/pganalyze/collector
+ENV GOPATH=/go
```
msakrejda, line 4: For other reviewers, the space-based syntax here is [now deprecated](https://docs.docker.com/reference/dockerfile/#env).

msakrejda (approved):

Dockerfile
keiko713: Thanks for the note! Yeah I added `label` a while back so it's safe to remove/retire: https://github.com/pganalyze/collector/pull/534/files#r1571635098

PR #560, opened 2024-07-04
Support multiline logs with AlloyDB

keiko713:
With AlloyDB, when the log line contains a new line, it was cutting anything after a new line unintentionally.
This updates the regex to make sure that all content will be kept.

Confirmed that it'll match: https://go.dev/play/p/ym6RhMU26ud


lfittl (approved):
👍 Nice catch!

PR #562, opened 2024-07-09
Add option to avoid collecting distributed index stats for Citus

msakrejda:
In some situations, the problems that DISABLE_CITUS_SCHEMA_STATS is
intended to solve are actually only with indexes, not with other
schema stats. By adding an index-specific setting, we can support more
functionality for affected users.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -131,9 +131,8 @@ type ServerConfig struct {
 	SystemTypeFallback  string `ini:"api_system_type_fallback"`
 	SystemScopeFallback string `ini:"api_system_scope_fallback"`
 
-	AlwaysCollectSystemData bool `ini:"always_collect_system_data"`
-	DisableCitusSchemaStats bool `ini:"disable_citus_schema_stats"`
-	DisableCitusIndexStats  bool `ini:"disable_citus_index_stats"`
+	AlwaysCollectSystemData bool   `ini:"always_collect_system_data"`
+	DisableCitusSchemaStats string `ini:"disable_citus_schema_stats"` // none/all/index
 
 	// Configures the location where logfiles are - this can either be a directory,
 	// or a file - needs to readable by the regular pganalyze user
--- a/config/config.go
+++ b/config/config.go
@@ -133,6 +133,7 @@ type ServerConfig struct {
 
 	AlwaysCollectSystemData bool `ini:"always_collect_system_data"`
 	DisableCitusSchemaStats bool `ini:"disable_citus_schema_stats"`
+	DisableCitusIndexStats  bool `ini:"disable_citus_index_stats"`
 
 	// Configures the location where logfiles are - this can either be a directory,
 	// or a file - needs to readable by the regular pganalyze user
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -133,6 +133,7 @@ type ServerConfig struct {
 
 	AlwaysCollectSystemData bool `ini:"always_collect_system_data"`
 	DisableCitusSchemaStats bool `ini:"disable_citus_schema_stats"`
+	DisableCitusIndexStats  bool `ini:"disable_citus_index_stats"`
```
lfittl, line 136: I was wondering whether we should instead make `disable_citus_schema_stats` have three options? `all`, `index`, `none` (with the default being `none`)

Though I guess technically you could enable index stats and not table stats, so maybe this approach works too? (at the cost of adding an additional config setting)

msakrejda: That would be simpler. Filtering out table but not index stats is an unlikely use case, so I don't think we need to support that (and if we do, we could always allow `disable_citus_schema_stats=table`).

It would be a backward compatibility break, though. Or I guess we could map truthy values to "all" and falsey ones to "none"? Maybe I'll try that; that seems reasonable.

msakrejda: Okay, did that. Our existing tests don't cover this, and it's probably not worth the CI cycles, but I tried adding various values of the setting to the `citus` integration test and it behaves as expected.

input/postgres/relation_stats_aux.go
```diff
@@ -21,37 +21,43 @@ SELECT logicalrelid::oid,
  WHERE ($1 = '' OR (n.nspname || '.' || c.relname) !~* $1)
 `
 
+func collectCitusRelationStats(postgresVersion state.PostgresVersion, server *state.Server) bool {
+	return postgresVersion.IsCitus && server.Config.DisableCitusSchemaStats != "all"
+}
```
msakrejda, line 26: This file best reviewed with whitespace diffs disabled. I decided to break the guards out into separate functions and use early returns to simplify the original function bodies.

lfittl (approved):
:+1: That looks good, thanks for tackling this!

config/read.go
```diff
@@ -291,7 +305,7 @@ func getDefaultConfig() *ServerConfig {
 		config.AlwaysCollectSystemData = parseConfigBool(alwaysCollectSystemData)
 	}
 	if disableCitusSchemaStats := os.Getenv("DISABLE_CITUS_SCHEMA_STATS"); disableCitusSchemaStats != "" {
-		config.DisableCitusSchemaStats = parseConfigBool(disableCitusSchemaStats)
+		config.DisableCitusSchemaStats = parseConfigDisableCitusSchemaStats(disableCitusSchemaStats)
```
msakrejda, line 308: @lfittl while working on #494, I realized that this was not the right way to write this because of how we load config. This is only the default config plus config from env, and would be overridden by anything in the config file, if one is present. This should probably be in `preprocessConfig` instead, right?

lfittl: Ah yeah, good point - I think that's right.

PR #563, opened 2024-07-11
Test run: Avoid collector hanging when interrupted via CTRL+C (SIGINT)

### LINKED ISSUE #561: Can't ctrl-c pganalyze-collector
When I am trying to ctrl-c a failed `pganalyze-collector --test` run, it hangs:

```
Test successful

^C
```

Seems like signals might not be handled correctly when interactive?

lfittl:
Due to the way the signal handling is written, we stop checking the testRunSuccess channel once a SIGINT is received. However the code that marks the wait groups as done could only return once it finished writing to the channel, since it was unbuffered. Make the channel buffered so that the code can return as intended.

Fixes #561


msakrejda (approved):
A cleaner fix might be to propagate an exit signal to all goroutines (via channel) and use select, but that'd be way more complicated, and this seems robust.

PR #569, opened 2024-07-16
Avoid using apt-key command for Debian

keiko713:
As `apt-key add` is deprecated in Debian 11 (“Bullseye”) and Ubuntu 20.04 LTS (“Focal Fossa”), which are our minimal supported versions of the collector, let's remove the usage of it.

This thread has a great answer of why we shouldn't use `apt-key add`, and I updated the command to make sure to follow the best practice here.
https://askubuntu.com/questions/1286545/what-commands-exactly-should-replace-the-deprecated-apt-key

As a reference, the following message was shown during the installation process:

```
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).

W: https://packages.pganalyze.com/ubuntu/noble/dists/stable/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details.
```

### DIFF
--- a/contrib/install.sh
+++ b/contrib/install.sh
@@ -162,26 +162,14 @@ metadata_expire=300" | $maybe_sudo tee -a /etc/yum.repos.d/pganalyze_collector.r
   $maybe_sudo yum $yum_opts install pganalyze-collector <$user_input
 elif [ "$pkg" = deb ];
 then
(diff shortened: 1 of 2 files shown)


msakrejda (approved):
Awesome, thanks for updating this!

PR #575, opened 2024-07-18
Replace use of github.com/satori/go.uuid with github.com/google/uuid 

lfittl:
The old library we've utilized is unmaintained, and there is an old CVE (https://github.com/advisories/GHSA-33m6-q9v5-62r7), that, whilst not a problem for the way the collector uses UUIDs, is still good to clear out to avoid scanner complaints.

Additionally this PR first cleans up some duplicative UUID generation, to make the actual library switch easier, see individual commits.


input/system/selfhosted/otel_handler.go
```diff
@@ -28,8 +27,6 @@ import (
 
 func logLineFromJsonlog(record *common.KeyValueList, tz *time.Location) (state.LogLine, *state.LogLine) {
 	var logLine state.LogLine
-	logLine.CollectedAt = time.Now()
-	logLine.UUID = uuid.NewV4()
```
keiko713, line 32: These lines don't look like having duplicate. Can you tell me where these are generated? (the same for tembo)

keiko713 (approved):
One question but otherwise looking good

input/system/selfhosted/otel_handler.go
lfittl: See commit message for an explanation - we assign it in the log runner.

output/full.go
```diff
@@ -47,7 +47,11 @@ func submitFull(ctx context.Context, s snapshot.FullSnapshot, server *state.Serv
 	var err error
 	var data []byte
 
-	snapshotUUID := uuid.NewV4()
+	snapshotUUID, err := uuid.NewRandom()
```
msakrejda, line 50: As long as we're making changes here, should we consider [v7](https://pkg.go.dev/github.com/google/uuid#NewV7)?

msakrejda (approved):
This will probably conflict with #494, but it shouldn't be too bad; I can rebase.

output/full.go
lfittl: I thought about that, but figured it might be best to keep that as a separate PR for clarity - though we could certainly ship it at the same time. @seanlinsley any concerns with switching to v7?

lfittl:
> This will probably conflict with #494, but it shouldn't be too bad; I can rebase.

Yeah, sorry about causing any conflicts there - it was easiest to clean up some of that logic now, before doing a lot of unnecessary search and replace.

output/full.go
seanlinsley: No, having timestamp-prefixed UUIDs is generally good. We currently replace some random UUIDs at snapshot processing time with timestamp-prefixed versions because the collector doesn't.

PR #579, opened 2024-07-26
Update AWS SDK to v1.55.3

cyclops23:
Trying to query RDS APIs across AWS accounts from within EKS fails when using pod identities.  

```
ERROR - Could not download logs: could not collect logs: Error finding instance for cluster ID "foo-production-primary": NoCredentialProviders: no valid providers in chain. Deprecated.
For verbose messaging see aws.Config.CredentialsChainVerboseErrors
HINT - This may occur if you have not assigned an IAM role to the collector EC2 instance, and have not provided AWS credentials through another method
```

However I have set up the pod identity and set the appropriate `AWS_ACCOUNT_ID` and `AWS_ASSUME_ROLE` environment variables for the collector to use.

Turns out that the AWS SDK used by the collector currently is a much earlier version than that which [supports pod identities](https://docs.aws.amazon.com/eks/latest/userguide/pod-id-minimum-sdk.html).

I updated the SDK version, tested with a custom collector image and the issue was resolved.  I did not observe any side-effects so contributing the code change back here.


lfittl:
> I updated the SDK version, tested with a custom collector image and the issue was resolved. I did not observe any side-effects so contributing the code change back here.

Nice, thanks for the contribution!

We'll review this shortly, and agreed that it makes sense to update the SDK to support pod identity.

msakrejda (approved):
Looks good. I re-vendored the updated dependency to ensure it matches, and tested a bit with an RDS instance. Thanks again for the contribution.

Issue #21, opened 2018-05-08
Error collecting replication statistics


declaresub:
I'm getting this warning in the log:

Error collecting replication statistics: sql: Scan error on column index 14: converting driver.Value type <nil> ("<nil>") to a int64: invalid syntax.

I'm running postgresql 10.3 with replication enabled for backup to barman.

lfittl:
@declaresub Thanks for the report - did you create the monitoring user functions, in particular this one, as a super user:

```
CREATE OR REPLACE FUNCTION pganalyze.get_stat_replication() RETURNS SETOF pg_stat_replication AS
$$
  /* pganalyze-collector */ SELECT * FROM pg_catalog.pg_stat_replication;
$$ LANGUAGE sql VOLATILE SECURITY DEFINER;
```

declaresub:
Yes; that function exists in my pganalyze schema.

declaresub:
In my db, SELECT * FROM pg_catalog.pg_stat_replication returns a single row, with the value of replay_lsn being NULL.  I expect that's the offending column.

lfittl:
@declaresub Which collector version are you using?

It seems that this is the bug that got fixed about a year ago in https://github.com/pganalyze/collector/commit/1fa72e27c6f18ee6027192670acf0a6f37883738 (which first got released in v0.9.13)

declaresub:
I'm using 0.12.0.

lfittl:
@declaresub My apologies, this was actually a bug - this is now fixed in the newly released 0.13.1 collector version.

Let me know in case that doesn't fix the issue you were seeing!

dexterous:
@lfittl I'm seeing this in the current `stable`, which AFAICT is `v0.17.1`

lfittl:
@dexterous Thanks for your note, and sorry for the late reply.

Based on the report in #34 this is now more flexible and should not error out anymore. We'll have a new release soon that includes this bugfix.

PR #581, opened 2024-08-07
Inline protobuf definitions into collector repo

seanlinsley:
Historically the snapshot protobuf definitions were stored in a [separate repo](https://github.com/pganalyze/collector-snapshot/) so that the main application could include them as a submodule. But now that the the main application includes the collector itself as a submodule, the additional maintenance burden to have this as a separate repo no longer makes sense.

For reference, I followed [these steps](https://stackoverflow.com/a/1789374) to inline the protobuf definitions.


msakrejda (commented):
+1 in general, but I wonder if instead we shouldn't
 1. create a branch in this repo
 2. drop the submodule in this repo
 3. move everything in the snapshot repo into a `protobuf` directory
 4. add the snapshot repo as a remote to this repo
 5. merge the snapshot repo main branch into this repo's branch

I haven't tried this yet, but I think that would let us preserve the full history of the snapshot repo. Thoughts?

In any case, after we merge this, we should remember to archive https://github.com/pganalyze/collector-snapshot to avoid confusion.

seanlinsley:
Since the snapshot definitions are largely append-only and all of the actual implementation was in the collector repo, I'm not sure it's worth creating a confusing git history here just to preserve git blames.

Anecdotally, I can't think of a single time when I needed to know the provenance of a history field by looking at the protobuf definitions themselves; I'd always look at the collector repo or the main repo. And if we did merge the history of the two repos like you suggest, if I were to look at a git blame of the protobuf it wouldn't include the collector code changes since that'd still be a separate commit, so there wouldn't be a simple way to see all of the relevant changes.

msakrejda:
I tried this locally (had to add `--allow-unrelated-histories`, but otherwise, it went smoothly), and the history looks pretty clean to me. All the protobuf history is basically just in the branch. You're right about `git blame`, but since the protobuf repo moves slowly, it shouldn't be hard to correlate by timestamp.

I'd still prefer this approach, but I'll let others weigh in.

seanlinsley:
Let's walk through a possible scenario: you reach [this commit](https://github.com/pganalyze/collector-snapshot/commit/37171e773e957c09b4eed5a95ceb845b49e9f76e) from a git blame and want to find the related collector changes. Screenshot for reference:


I see two problems:

- The `#21` in the commit message would either no longer be a link, or it would incorrectly link to collector issue [#21](https://github.com/pganalyze/collector/issues/21).
- GitHub doesn't give you a way to find commits around the same timestamp as this commit; you can only navigate to its parent commit.

If the history was instead squashed, you'd reach this PR (#581) and then navigate to the [archived repo](https://github.com/pganalyze/collector-snapshot/), and perform the same git blame there. I think that leaves much less potential for confusion.

msakrejda:
Thanks for illustrating the use case; that makes sense. I prefer to investigate history through `git log`, and I think that'd be easier with a combined history, but I concede the GH magic does not play well with that. I suspect external contributors would also prefer something GH-friendly, so maybe you're right. I'll leave it open for comment for a bit, but I think if no one chimes in, we can go with your approach.

keiko713:
tl;dr I'm indecisive, and I'm okay with either way.

I was initially commenting here saying that we could go with Sean's approach, given the downside Sean illustrated + the history of will-be-archived collector-snapshot repo is likely going to be enough for me.
However, reading more about the stackoverflow post that Sean pasted in the PR summary (people in general wanting to preserve the history, including Maciek, and I do think it's a good practice too), also looking at the git blame UI of the current collector-snapshot, I wonder if the case Sean mentioned will be that common. IOW, if the downside is minimal, I want to support Maciek's way. That `#21` is squashed, and out of 30+ merged PR, I only see a few squashed ones... actually more like ~10 PRs squashed. The squashed ones will have that issue more indeed, as the commit message will have something like `(#21)`. Well, now I have not so small number in front of me, maybe the downside is not so minimal (also unsquashed ones still have the problem that it's not easy to find the PR from the commit).

Well, if we go with Sean's way, I may add one request. It'd be great if you can add something like "see https://github.com/pganalyze/collector-snapshot for previous history" to the commit message of this PR, so that when I do git blame or see the history of these files/directories in the future, I will have more easy pointer there.

msakrejda (approved):
Thanks for weighing in, @keiko713. I'm going to approve this because I realized that if we go with @seanlinsley's approach, we can always go back and merge histories later, and having this merged will simplify his other work.

+1 on mentioning the old snapshot repo in the commit message, though.

PR #582, opened 2024-08-14
fix: Handle unexpected status code in GetClusterInfo

irfanhabib:
This handles the case where the HTTP response status code is not `200 OK` or the response body is empty in `client.go`. Instead of returning a `nil` error, it now returns a formatted error message with the status code and response body. This improves error handling and provides more information about the issue.

I hit this issue when PGAnalyze collector stopped working for me and the snapshots failed to be proccesed due the error 

```
Could not process server: runtime error: invalid memory address or nil pointer dereference
```

Due to the `nil` error, the subsequent code (https://github.com/pganalyze/collector/blob/5e8eab6bb9a6cdf41e8874fdd1e8b95fd1ee18fc/input/system/crunchy_bridge/system.go#L35) panics due to `clusterInfo` being `nil`. It turned out my CrunchyBridge API Key had expired resulting in this issue

### DIFF
--- a/input/system/crunchy_bridge/client.go
+++ b/input/system/crunchy_bridge/client.go
@@ -104,7 +104,7 @@ func (c *Client) GetClusterInfo(ctx context.Context) (*ClusterInfo, error) {
 	}
 
 	if resp.StatusCode != http.StatusOK || len(body) == 0 {
-		return nil, err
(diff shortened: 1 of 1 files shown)


msakrejda:
Thanks, @irfanhabib! The change looks right to me. I'll kick off a CI run and we'll try to get this into the next collector release.

msakrejda (approved):

PR #583, opened 2024-08-15
Add snapshot WebSocket API

seanlinsley:
This introduces a persistent WebSocket connection for all server communication. Additional details:
- If unable to start a WebSocket connection, the collector falls back to the legacy API.
- `grant_logs` is still used for every log snapshot, since it's needed to generate a S3 upload URL. This will be deprecated in a separate PR.
- All messages are zlib-compressed protobuf. The collector sends snapshots to the server, and the server sends `ServerMessage`s.
- The existing `state.GrantConfig` struct has been replaced with the same structure inside the protobuf, so the server and collector stay in sync.
- For future use, `ServerMessage` includes a `Pause` message that will let us disable duplicate collector instances for the same server, and an `ExplainRun` message for the Explain Upload feature that's currently in-progress.


runner/websocket.go
```diff
@@ -0,0 +1,103 @@
...
+	headers["Pganalyze-System-Id-Fallback"] = []string{server.Config.SystemIDFallback}
+	headers["Pganalyze-System-Type-Fallback"] = []string{server.Config.SystemTypeFallback}
+	headers["Pganalyze-System-Scope-Fallback"] = []string{server.Config.SystemScopeFallback}
+	headers["User-Agent"] = []string{util.CollectorNameAndVersion}
+	headers["Sec-WebSocket-Protocol"] = []string{"websocket"}
+	conn, response, err := websocket.DefaultDialer.DialContext(connCtx, url.String(), headers)
+	if err != nil {
+		cancelConn()
+		time.Sleep(10 * time.Second)
+		logger.PrintError("Error starting websocket: %s %v", err, response)
```
seanlinsley, line 54: I wonder if there's a better output to show when doing a test run?

state/state.go
```diff
@@ -303,7 +288,10 @@ func MakeServer(config config.ServerConfig, testRun bool) *Server {
 		ActivityStateMutex:    &sync.Mutex{},
 		CollectionStatusMutex: &sync.Mutex{},
 		LogTimezoneMutex:      &sync.Mutex{},
+		SnapshotStream:        make(chan []byte),
 	}
+	server.Pause = pganalyze_collector.ServerMessage_Pause{Pause: false}
+	server.Grant.Config.Features = &pganalyze_collector.ServerMessage_Features{}
```
seanlinsley, line 294: This was needed to avoid a nil pointer error, since the protobuf library doesn't have an option to nest structs without using pointers https://github.com/golang/protobuf/issues/1225

state/state.go
```diff
@@ -278,6 +259,10 @@ type Server struct {
 
 	SelfTest *SelfTestResult
 
+	WebSocket      *websocket.Conn
+	SnapshotStream chan []byte
+	Pause          pganalyze_collector.ServerMessage_Pause
```
seanlinsley, line 264: These probably need a synchronization primitive, though I wasn't sure which one is the best fit, and this PR adds data-race detection to the `make run` command in the Makefile which didn't find an issue here.

state/state.go
seanlinsley: Hmm, apparently the `integration_test` command [failed on CI](https://github.com/pganalyze/collector/actions/runs/10406859790/job/28820868142) for the same reason. Any idea how to fix this?

runner/websocket.go
```diff
@@ -0,0 +1,103 @@
...
+			select {
+			case <-connCtx.Done():
+				err := server.WebSocket.Close()
+				if err != nil {
+					logger.PrintError("Error closing websocket: %s", err)
+					return
+				}
+				server.WebSocket = nil
+				return
+			case snapshot := <-server.SnapshotStream:
```
keiko713, line 69: I wonder if it would be nice to either have a separate stream or pass type in addition to the snapshot bytes data, so that we will know which snapshot we're processing here. (e.g. to have a log of which snapshot went well and which didn't, like we log currently).
Also `logger.PrintError` is used for the below log but I think it's not an error level log

keiko713 (commented):
Still haven't tested by myself yet but looks pretty straight forward and promising.
Left some comments for things I noticed so far. I'm planning to do more review next week 👀

state/state.go
keiko713: It's failing because that integration test is doing the `--dry-run --force-state-update` (you can repro locally, actually even without `--force-state-update`).

With that test, it runs the full snapshot dry run test first, then move on to the activity snapshot dry run test. After the successful full snapshot test, `server.Grant` is updated with `newGrant, err = grant.GetDefaultGrant(ctx, server, globalCollectionOpts, logger)`, so it will become something like `state.Grant{}`. As you worked around here, you likely need to do similar operation for other places too to make sure that `Grant.Config.Features` won't be nil.

Is it possible to implement this part without introducing this change? Feels like that might be simpler. Or, keep the struct in a previous way and update it more manually (instead of doing `server.Grant.Config = *message.GetConfig()`, update one by one, for example).

runner/full.go
```diff
@@ -89,13 +89,19 @@ func processServer(ctx context.Context, server *state.Server, globalCollectionOp
 	var collectionStatus state.CollectionStatus
 	var err error
 
+	if server.Pause.Pause == true {
+		logger.PrintWarning("Snapshot processing disabled by pganalyze server: %s", server.Pause.Reason)
+		return newState, newGrant, collectionStatus, nil
```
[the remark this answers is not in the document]
seanlinsley: My thought was that we don't want to preserve the existing state, since the diffs (e.g. for query stats) would become out of date while snapshot collection is paused.

This certainly could return an error instead of printing a warning. I'm curious what others think.

runner/websocket.go
seanlinsley: In 8e0234d0bc7d536f5750d094ac90d0eccf6d10d4 I updated the code to print similar log lines to what you see with the existing API. The `PrintError` I was using for testing; I've changed that to `PrintVerbose`.

state/state.go
seanlinsley: Thanks, that's been fixed in 98b6cdd10b86a0cf919b6bdba092f0013f9e256a.

Yeah this nil pointer issue is unfortunate. I changed the `state.Grant` to contain the protobuf struct directly to avoid the two versions of the struct from getting out of sync. At the time I didn't realize this would be an issue.

runner/websocket.go
```diff
@@ -0,0 +1,104 @@
...
+			message := &pganalyze_collector.ServerMessage{}
+			err = proto.Unmarshal(data.Bytes(), message)
+			if err != nil {
+				logger.PrintError("Error parsing ServerMessage: %s", err)
+			} else if message.GetConfig() != nil {
+				server.Grant.Config = *message.GetConfig()
+			} else if message.GetPause() != nil {
+				server.Pause = *message.GetPause()
+			} else if message.GetExplainRun() != nil {
+				logger.PrintError("ExplainRun: %v", message.GetExplainRun()) // TODO
```
msakrejda, line 100: This isn't really an error, and we don't expect the server to send this message until we actually implement it, so I think we could just `PrintVerbose` for now:
```suggestion
				logger.PrintVerbose("ExplainRun: %v", message.GetExplainRun()) // TODO
```

state/state.go
msakrejda: Oof, the existing races are a mess. I wonder if we shouldn't rethink our flow rather than try to fix those directly. No strong feelings on the new mechanism.

msakrejda (commented):
Works as advertised, and reconnects automatically, but I did run into a panic on Ctrl+C:

```
2024/08/16 14:03:51 V [server1] Uploading snapshot to websocket
^C2024/08/16 14:03:58 I Exiting...
2024/08/16 14:03:58 V Log file fsnotify watcher received stop signal
2024/08/16 14:03:58 E [server1] Error closing websocket: close tcp 127.0.0.1:38368->127.0.0.1:5200: use of closed network connection
2024/08/16 14:03:58 V Stopping log tail for /var/log/postgresql/postgresql-15-main.log (stop requested)
2024/08/16 14:03:58 V Log file fsnotify watcher received stop signal
2024/08/16 14:03:58 V Stopping log tail for /var/log/postgresql/postgresql-16-main.log (stop requested)
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xe487ca]

goroutine 9032 [running]:
github.com/gorilla/websocket.(*Conn).Close(...)
	/home/maciek/code/collector/vendor/github.com/gorilla/websocket/conn.go:352
github.com/pganalyze/collector/runner.connect.func1()
	/home/maciek/code/collector/runner/websocket.go:63 +0x16a
created by github.com/pganalyze/collector/runner.connect in goroutine 126
	/home/maciek/code/collector/runner/websocket.go:59 +0x78f
```

runner/websocket.go
seanlinsley: Yeah, I was thinking it'd specifically be useful for enterprise users who e.g. have a proxy configured that doesn't support websockets.

runner/websocket.go
```diff
@@ -0,0 +1,104 @@
...
+	headers["Pganalyze-System-Scope"] = []string{server.Config.SystemScope}
+	headers["Pganalyze-System-Id-Fallback"] = []string{server.Config.SystemIDFallback}
+	headers["Pganalyze-System-Type-Fallback"] = []string{server.Config.SystemTypeFallback}
+	headers["Pganalyze-System-Scope-Fallback"] = []string{server.Config.SystemScopeFallback}
+	headers["User-Agent"] = []string{util.CollectorNameAndVersion}
+	headers["Sec-WebSocket-Protocol"] = []string{"websocket"}
+	conn, response, err := websocket.DefaultDialer.DialContext(connCtx, url.String(), headers)
+	if err != nil {
+		cancelConn()
+		time.Sleep(10 * time.Second)
```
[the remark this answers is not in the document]
seanlinsley: Hmm, I don't remember. This was probably from an earlier version of the code. I'll remove it.

runner/websocket.go
keiko713: Maybe you can reuse `MarkCollectionAspectError` like:

```
server.SelfTest.MarkCollectionAspectError(state.CollectionAspectApiConnection, "error starting websocket with API: %s", err)
```

This is currently set by `GetDefaultGrant` but I think it won't be called via websocket way.

main.go
```diff
@@ -255,6 +255,8 @@ func run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
 		wg.Done()
 	}, logger, "high frequency query statistics of all servers", schedulerGroups["stats"])
 
+	runner.SetupWebsocketForAllServers(ctx, servers, globalCollectionOpts, logger)
```
keiko713, line 258: Should we have some sort of "close websocket" shutdown? This `run` func is returning the `shutdown`), we could tweak/reuse that.

runner/websocket.go
```diff
@@ -0,0 +1,103 @@
...
+				err := server.WebSocket.Close()
+				if err != nil {
+					logger.PrintError("Error closing websocket: %s", err)
+					return
+				}
+				server.WebSocket = nil
+				return
+			case snapshot := <-server.SnapshotStream:
+				logger.PrintVerbose("Uploading snapshot to websocket")
+				server.WebSocket.WriteMessage(websocket.BinaryMessage, snapshot)
```
keiko713, line 71: Looks like this WriteMessage returns an error, would it be nice to log the error when it happens?

runner/full.go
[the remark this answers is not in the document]
keiko713: I was thinking that this is somewhat close to the ignore replica case, where we handle it with an error. Though more I think about it, I think it's different and it's fine as is. I was also a bit worried about the consequences of storing the empty struct as a `server.PrevState` but that's how it is when you boot up the collector without a state file, so seems fine to me.

main.go
seanlinsley: Shutdown should happen automatically AFAIK because the websocket runner calls `context.WithCancel`. I could be wrong though.

runner/websocket.go
seanlinsley: 👍. I'm not sure what kind of error could occur here that wouldn't be fatal, so I've updated it to also close the websocket.

main.go
keiko713: Ah yeah just connected the dots, you created `connCtx, cancelConn := context.WithCancel(ctx)` and doing `server.WebSocket.Close()` within `<-connCtx.Done()` so it should be fine.
We could write `defer cancelConn()` and `defer conn.Close()` over there also but I guess it's somewhat optional given they are handled within `<-connCtx.Done()`.

runner/websocket.go
```diff
@@ -0,0 +1,104 @@
...
+	go func() {
+		for {
+			select {
+			case <-connCtx.Done():
+				err := server.WebSocket.Close()
+				if err != nil {
+					logger.PrintError("Error closing websocket: %s", err)
+					return
+				}
+				server.WebSocket = nil
```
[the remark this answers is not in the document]
seanlinsley: fixed in 8d076e63f47d3a44b1fa376abd05f8e9571b5070

protobuf/server_message.proto
```diff
@@ -0,0 +1,47 @@
...
+    // Statement timeout for pg_stat_statements query text requests (defaults to 120s)
+    int32 statement_timeout_ms_query_text = 3;
+  }
+
+  message Pause {
+    bool pause = 1;
+    string reason = 2;
+  }
+
+  message ExplainRun {
```
seanlinsley, line 41: Todo: add a comment that this isn't currently used.

seanlinsley:
It looks like the test suite failed because it's trying to open a WebSocket connection to the production servers, but that side of the code hasn't been merged yet.

runner/websocket.go
```diff
@@ -0,0 +1,131 @@
...
+		NoProxy:    server.Config.NoProxy,
+	}
+	dialer := websocket.Dialer{
+		ReadBufferSize:  10240,
+		WriteBufferSize: 10240,
+		Proxy: func(req *http.Request) (*url.URL, error) {
+			return proxyConfig.ProxyFunc()(req.URL)
+		},
+	}
+	url, _ := url.Parse(server.Config.APIBaseURL + "/v2/snapshots/websocket")
```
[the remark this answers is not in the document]
msakrejda: To answer my own question, the error message can be confusing, but it's basically okay because of the error check below:

```
2024/09/23 16:11:55 W [server1] Error starting websocket: tls: first record does not look like a TLS handshake <nil>
2024/09/23 16:11:55 W [server2] Error starting websocket: tls: first record does not look like a TLS handshake <nil>
2024/09/23 16:12:00 E [server1] Could not collect activity for server: could not get default grant for activity snapshot: "htp://localhost:5200/v2/snapshots/grant": unsupported protocol scheme "htp"
2024/09/23 16:12:00 E [server2] Could not collect activity for server: could not get default grant for activity snapshot: "htp://localhost:5200/v2/snapshots/grant": unsupported protocol scheme "htp"
```
or
```
2024/09/23 16:13:37 W [server2] Error starting websocket: websocket: bad handshake &{400 Bad Request 400 HTTP/1.1 1 1 map[Content-Length:[166] Content-Type:[text/html] Date:[Mon, 23 Sep 2024 23:13:37 GMT] Server:[nginx/1.18.0 (Ubuntu)]] {0xc000c54060} 166 [] true false map[] 0xc0004d5b00 <nil>}
2024/09/23 16:13:37 W [server1] Error starting websocket: websocket: bad handshake &{400 Bad Request 400 HTTP/1.1 1 1 map[Content-Length:[166] Content-Type:[text/html] Date:[Mon, 23 Sep 2024 23:13:37 GMT] Server:[nginx/1.18.0 (Ubuntu)]] {0xc000e7e180} 166 [] true false map[] 0xc0001fe120 <nil>}
2024/09/23 16:13:55 E [server1] Could not collect activity for server: could not get default grant for activity snapshot: "http: //localhost:5200/v2/snapshots/grant": http: no Host in request URL
2024/09/23 16:13:55 E [server2] Could not collect activity for server: could not get default grant for activity snapshot: "http: //localhost:5200/v2/snapshots/grant": http: no Host in request URL
```

It is possible to hit a panic (I added a `%` in the host name):
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xec36af]

goroutine 107 [running]:
github.com/pganalyze/collector/runner.connect({0x18b7dc8?, 0xc0007a2000?}, 0xc000872a88, {{0xc1b49b7efae697be, 0xe90074, 0x22e87e0}, 0x1, 0x1, 0x1, 0x1, ...}, ...)
	/home/maciek/code/collector/runner/websocket.go:49 +0x24f
github.com/pganalyze/collector/runner.SetupWebsocketForAllServers.func1(0xc000872a88)
	/home/maciek/code/collector/runner/websocket.go:26 +0x1ab
created by github.com/pganalyze/collector/runner.SetupWebsocketForAllServers in goroutine 1
	/home/maciek/code/collector/runner/websocket.go:22 +0x8c
```

msakrejda (commented):
I tested this with https://github.com/pganalyze/pganalyze/pull/4234 and it worked fine. I left some minor comments on the code, but it looks fine, too.

runner/websocket.go
```diff
@@ -0,0 +1,131 @@
...
+				}
+			}
+		}
+	}()
+	// Reader goroutine
+	go func() {
+		for {
+			_, compressedData, err := conn.ReadMessage()
+			if err != nil {
+				if !websocket.IsCloseError(err, 1005) { // Normal close event
```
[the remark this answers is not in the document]
seanlinsley: 1005 and 1006 are the only close events I've been able to cause from the server side. The server-side websocket implementation doesn't seem to let us provide a close message.

1006 (abnormal closure) happened before the recent server-side commit that added graceful shutdown.

runner/websocket.go
```diff
@@ -0,0 +1,131 @@
...
+				}
+				server.SelfTest.MarkCollectionAspectError(state.CollectionAspectWebSocket, "error starting WebSocket: %s", err)
+				cancelConn()
+				return
+			}
+			server.SelfTest.MarkCollectionAspectOk(state.CollectionAspectWebSocket)
+			var data bytes.Buffer
+			r, err := zlib.NewReader(bytes.NewReader(compressedData))
+			if err != nil {
+				logger.PrintWarning("Error decompressing ServerMessage: %s", err)
```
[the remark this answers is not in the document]
seanlinsley: Good catch. Updated in 218f824c0b1a922a10922423b7ef802b14c0497c

runner/websocket.go
```diff
@@ -0,0 +1,131 @@
...
+				return
+			}
+			defer r.Close()
+			io.Copy(&data, r)
+			message := &pganalyze_collector.ServerMessage{}
+			err = proto.Unmarshal(data.Bytes(), message)
+			if err != nil {
+				logger.PrintWarning("Error parsing ServerMessage: %s", err)
```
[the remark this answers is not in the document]
seanlinsley: New messages would be added to the [protobuf enum](https://github.com/pganalyze/collector/pull/583/files#diff-4cba84f43462652699ada1f4d9e63a7d83cef4bc51554df1f8ee29458e6e6ac0R6-R14), so the message would still be parsed, though presumably would be empty because its contents didn't match a known `oneof` type.

runner/websocket.go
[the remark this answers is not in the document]
seanlinsley: Note that the event codes are from the WebSocket spec not from gorilla/websocket, so if we link to something it might be better to use https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code for example.

protobuf/server_message.proto
```diff
@@ -0,0 +1,47 @@
...
+    Config config = 1;
+    // Server request to pause collection of data (e.g. in case of duplicate collectors)
+    Pause pause = 2;
+    // Request for the collector to generate an explain plan for a query
+    ExplainRun explain_run = 3;
+  }
+
+  message Config {
+    string server_id = 1;
+    string server_url = 2;
```
lfittl, line 18: We should check this server URL instead in the part where we output the URL to view the server:

https://github.com/pganalyze/collector/blob/f9752507c41c0fab792d5760beaea17504054372/selftest/summary.go#L222

(and drop server.PGAnalyzeURL)

lfittl: Also when we do that we should make sure that we still see the server URL at the bottom of `--test` for both the HTTP and the WebSocket mechanism.

runner/activity.go
```diff
@@ -41,7 +46,9 @@ func processActivityForServer(ctx context.Context, server *state.Server, globalC
 		}
 	}
 
-	if !globalCollectionOpts.ForceEmptyGrant {
+	if server.WebSocket.Load() != nil {
+		newGrant = *server.Grant.Load()
```
lfittl, line 50: There could be a case here where the websocket is initialized, but we didn't get the grant server message yet (or it wasn't processed yet by the Go routine).

Should we be blocking here, or falling back to HTTP if the websocket grant is not yet valid?

runner/full.go
```diff
@@ -229,7 +238,7 @@ func CollectAllServers(ctx context.Context, servers []*state.Server, globalColle
 					}
 				}
 			} else {
-				server.Grant = grant
+				server.Grant.Store(&grant)
```
lfittl, line 241: As discussed, this might be a problem in the websocket case because `grant` could be the previous grant, and `server.Grant` might already have been replaced by a newer grant. I think it seems safe to just not do this store here (and in the error case 12 lines up) if we're on a websocket connection.

runner/full.go
```diff
@@ -88,29 +89,37 @@ func processServer(ctx context.Context, server *state.Server, globalCollectionOp
...
-			if server.Grant.Valid {
+			if server.Grant.Load().Valid {
 				logger.PrintVerbose("Could not acquire snapshot grant, reusing previous grant: %s", err)
 			} else {
-				return state.PersistedState{}, state.Grant{}, state.CollectionStatus{}, err
+				return newState, newGrant, collectionStatus, err
 			}
 		} else {
-			server.Grant = newGrant
+			server.Grant.Store(&newGrant)
```
lfittl, line 115: It seems like we actually do this twice, once here, and then once in the caller of this function. I think we should be good to drop this here? (or in the parent function?)

lfittl (commented):
Overall looks good - a few minor things pointed out, but seems close to be ready to merge.

state/state.go
```diff
@@ -303,8 +288,11 @@ func MakeServer(config config.ServerConfig, testRun bool) *Server {
 		LogStateMutex:         &sync.Mutex{},
 		ActivityStateMutex:    &sync.Mutex{},
 		CollectionStatusMutex: &sync.Mutex{},
+		SnapshotStream:        make(chan []byte),
 		LogParseMutex:         &sync.RWMutex{},
 	}
+	server.Grant.Store(&Grant{Config: pganalyze_collector.ServerMessage_Config{Features: &pganalyze_collector.ServerMessage_Features{}}})
```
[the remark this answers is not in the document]
seanlinsley: Several parts of the code expect the `Grant` to already exist. Additionally due to protobuf putting `Features` behind a pointer, it has to be explicitly allocated. I felt it was better to allocate these ahead of time than add nil checks everywhere.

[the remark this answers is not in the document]
lfittl: Right, I think we clarified this in our conversation but I didn't remove this comment - sorry for the noise!

runner/websocket.go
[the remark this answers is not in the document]
keiko713: I'm not Maciek but I _think_ what Maciek meant was changing this to below?

```suggestion
				if !websocket.IsCloseError(err, websocket.CloseNoStatusReceived) { // Normal close event
```

runner/websocket.go
```diff
@@ -0,0 +1,132 @@
...
+	headers := make(map[string][]string)
+	headers["Pganalyze-Api-Key"] = []string{server.Config.APIKey}
+	headers["Pganalyze-System-Id"] = []string{server.Config.SystemID}
+	headers["Pganalyze-System-Type"] = []string{server.Config.SystemType}
+	headers["Pganalyze-System-Scope"] = []string{server.Config.SystemScope}
+	headers["Pganalyze-System-Id-Fallback"] = []string{server.Config.SystemIDFallback}
+	headers["Pganalyze-System-Type-Fallback"] = []string{server.Config.SystemTypeFallback}
+	headers["Pganalyze-System-Scope-Fallback"] = []string{server.Config.SystemScopeFallback}
+	headers["User-Agent"] = []string{util.CollectorNameAndVersion}
+	conn, response, err := dialer.DialContext(connCtx, url.String(), headers)
```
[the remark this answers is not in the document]
seanlinsley: I got most of the way through this, but ran into a circular dependency error:
```
package github.com/pganalyze/collector
    imports github.com/pganalyze/collector/runner
    imports github.com/pganalyze/collector/grant
    imports github.com/pganalyze/collector/runner: import cycle not allowed
```

In general I think this should be a separate PR, since it's solidly a refactor that's unrelated to the changes in this PR.

<details><summary>Full diff</summary>

```sql

diff --git a/config/read.go b/config/read.go
index 0a6b2d60..733521c5 100644
--- a/config/read.go
+++ b/config/read.go
@@ -2,10 +2,8 @@ package config
 
 import (
 	"context"
-	"crypto/tls"
 	"fmt"
 	"io/ioutil"
-	"net"
 	"net/http"
 	"net/url"
 	"os"
@@ -21,11 +19,8 @@ import (
 	sdkresource "go.opentelemetry.io/otel/sdk/resource"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	semconv "go.opentelemetry.io/otel/semconv/v1.21.0"
-	"golang.org/x/net/http/httpproxy"
 
-
-	"github.com/hashicorp/go-retryablehttp"
 )
 
 const DefaultAPIBaseURL = "https://api.pganalyze.com"
@@ -372,70 +367,6 @@ func getDefaultConfig() *ServerConfig {
 	return config
 }
 
-func CreateHTTPClient(conf ServerConfig, logger *util.Logger, retry bool) *http.Client {
-	requireSSL := conf.APIBaseURL == DefaultAPIBaseURL
-	proxyConfig := httpproxy.Config{
-		HTTPProxy:  conf.HTTPProxy,
-		HTTPSProxy: conf.HTTPSProxy,
-		NoProxy:    conf.NoProxy,
-	}
-	transport := &http.Transport{
-		Proxy: func(req *http.Request) (*url.URL, error) {
-			return proxyConfig.ProxyFunc()(req.URL)
-		},
-		DialContext: (&net.Dialer{
-			Timeout:   30 * time.Second,
-			KeepAlive: 30 * time.Second,
-			DualStack: true,
-		}).DialContext,
-		MaxIdleConns:          100,
-		IdleConnTimeout:       90 * time.Second,
-		TLSHandshakeTimeout:   10 * time.Second,
-		ExpectContinueTimeout: 1 * time.Second,
-	}
-	if requireSSL {
-		transport.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
-			matchesProxyURL := false
-			if proxyConfig.HTTPProxy != "" {
-				proxyURL, err := url.Parse(proxyConfig.HTTPProxy)
-				if err == nil && proxyURL.Host == addr {
-					matchesProxyURL = true
-				}
-			}
-			if proxyConfig.HTTPSProxy != "" {
-				proxyURL, err := url.Parse(proxyConfig.HTTPSProxy)
-				if err == nil && proxyURL.Host == addr {
-					matchesProxyURL = true
-				}
-			}
-			// Require secure conection for everything except proxies
-			if !matchesProxyURL && !strings.HasSuffix(addr, ":443") {
-				return nil, fmt.Errorf("Unencrypted connection is not permitted by pganalyze configuration")
-			}
-			return (&net.Dialer{Timeout: 30 * time.Second, KeepAlive: 30 * time.Second, DualStack: true}).DialContext(ctx, network, addr)
-		}
-		transport.TLSClientConfig = &tls.Config{MinVersion: tls.VersionTLS12}
-	}
-
-	if retry {
-		client := retryablehttp.NewClient()
-		client.RetryWaitMin = 1 * time.Second
-		client.RetryWaitMax = 30 * time.Second
-		client.RetryMax = 4
-		client.Logger = nil
-		client.HTTPClient.Timeout = 120 * time.Second
-		client.HTTPClient.Transport = transport
-		return client.StandardClient()
-		// Note: StandardClient() only acts as a passthrough, handing the request to
-		// retryablehttp.Client whose nested HTTP client ends up using our custom transport.
-	} else {
-		return &http.Client{
-			Timeout:   120 * time.Second,
-			Transport: transport,
-		}
-	}
-}
-
 // CreateEC2IMDSHTTPClient - Create HTTP client for EC2 instance meta data service (IMDS)
 func CreateEC2IMDSHTTPClient(conf ServerConfig) *http.Client {
 	// Match https://github.com/aws/aws-sdk-go/pull/3066
diff --git a/grant/default.go b/grant/default.go
index 2d6eab66..534cfe76 100644
--- a/grant/default.go
+++ b/grant/default.go
@@ -10,6 +10,7 @@ import (
 	"github.com/pganalyze/collector/output/pganalyze_collector"
+	"github.com/pganalyze/collector/runner"
 )
 
 func GetDefaultGrant(ctx context.Context, server *state.Server, globalCollectionOpts state.CollectionOpts, logger *util.Logger) (state.Grant, error) {
@@ -20,6 +21,8 @@ func GetDefaultGrant(ctx context.Context, server *state.Server, globalCollection
 		return grant, err
 	}
 
+	req.Header = runner.AuthHeaders(server.Config, globalCollectionOpts)
+
 	req.Header.Set("Pganalyze-Api-Key", server.Config.APIKey)
 	req.Header.Set("Pganalyze-System-Id", server.Config.SystemID)
 	req.Header.Set("Pganalyze-System-Type", server.Config.SystemType)
diff --git a/logs/parse_test.go b/logs/parse_test.go
index f9a68566..fb72a7c4 100644
--- a/logs/parse_test.go
+++ b/logs/parse_test.go
@@ -129,7 +129,7 @@ var parseTests = []parseTestpair{
 		"2018-09-27 06:57:01.030 EST [20194][] : [1-1] [app=pganalyze_collector] LOG:  connection received: host=[local]",
 		nil,
 		state.LogLine{
-			OccurredAt:    time.Date(2018, time.September, 27, 6, 57, 1, 30*1000*1000, time.FixedZone("EST", -5*3600)),
+			OccurredAt:    time.Date(2018, time.September, 27, 6, 57, 1, 30*1000*1000, time.FixedZone("CDT", -5*3600)),
 			LogLevel:      pganalyze_collector.LogLineInformation_LOG,
 			BackendPid:    20194,
 			LogLineNumber: 1,
diff --git a/main.go b/main.go
index d50c7135..7e6976d8 100644
--- a/main.go
+++ b/main.go
@@ -60,8 +60,8 @@ func run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
 		prefixedLogger := logger.WithPrefix(cfg.SectionName)
 		prefixedLogger.PrintVerbose("Identified as api_system_type: %s, api_system_scope: %s, api_system_id: %s", cfg.SystemType, cfg.SystemScope, cfg.SystemID)
 
-		conf.Servers[idx].HTTPClient = config.CreateHTTPClient(cfg, prefixedLogger, false)
-		conf.Servers[idx].HTTPClientWithRetry = config.CreateHTTPClient(cfg, prefixedLogger, true)
+		conf.Servers[idx].HTTPClient = runner.CreateHTTPClient(cfg, prefixedLogger, false)
+		conf.Servers[idx].HTTPClientWithRetry = runner.CreateHTTPClient(cfg, prefixedLogger, true)
 		if cfg.OtelExporterOtlpEndpoint != "" {
 			conf.Servers[idx].OTelTracingProvider, conf.Servers[idx].OTelTracingProviderShutdownFunc, err = config.CreateOTelTracingProvider(ctx, cfg)
 			logger.PrintVerbose("Initializing OpenTelemetry tracing provider with endpoint: %s", cfg.OtelExporterOtlpEndpoint)
diff --git a/runner/util.go b/runner/util.go
new file mode 100644
index 00000000..b0c589d9
--- /dev/null
+++ b/runner/util.go
@@ -0,0 +1,125 @@
+package runner
+
+import (
+	"context"
+	"net"
+	"time"
+	"net/http"
+	"net/url"
+	"crypto/tls"
+	"fmt"
+	"strings"
+
+	"github.com/gorilla/websocket"
+	"golang.org/x/net/http/httpproxy"
+	"github.com/hashicorp/go-retryablehttp"
+)
+
+func CreateHTTPClient(conf config.ServerConfig, logger *util.Logger, retry bool) *http.Client {
+	requireSSL := conf.APIBaseURL == config.DefaultAPIBaseURL
+	proxyConfig := httpproxy.Config{
+		HTTPProxy:  conf.HTTPProxy,
+		HTTPSProxy: conf.HTTPSProxy,
+		NoProxy:    conf.NoProxy,
+	}
+	transport := &http.Transport{
+		Proxy: func(req *http.Request) (*url.URL, error) {
+			return proxyConfig.ProxyFunc()(req.URL)
+		},
+		DialContext: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+			DualStack: true,
+		}).DialContext,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+	}
+	if requireSSL {
+		transport.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
+			matchesProxyURL := false
+			if proxyConfig.HTTPProxy != "" {
+				proxyURL, err := url.Parse(proxyConfig.HTTPProxy)
+				if err == nil && proxyURL.Host == addr {
+					matchesProxyURL = true
+				}
+			}
+			if proxyConfig.HTTPSProxy != "" {
+				proxyURL, err := url.Parse(proxyConfig.HTTPSProxy)
+				if err == nil && proxyURL.Host == addr {
+					matchesProxyURL = true
+				}
+			}
+			// Require secure conection for everything except proxies
+			if !matchesProxyURL && !strings.HasSuffix(addr, ":443") {
+				return nil, fmt.Errorf("Unencrypted connection is not permitted by pganalyze configuration")
+			}
+			return (&net.Dialer{Timeout: 30 * time.Second, KeepAlive: 30 * time.Second, DualStack: true}).DialContext(ctx, network, addr)
+		}
+		transport.TLSClientConfig = &tls.Config{MinVersion: tls.VersionTLS12}
+	}
+
+	if retry {
+		client := retryablehttp.NewClient()
+		client.RetryWaitMin = 1 * time.Second
+		client.RetryWaitMax = 30 * time.Second
+		client.RetryMax = 4
+		client.Logger = nil
+		client.HTTPClient.Timeout = 120 * time.Second
+		client.HTTPClient.Transport = transport
+		return client.StandardClient()
+		// Note: StandardClient() only acts as a passthrough, handing the request to
+		// retryablehttp.Client whose nested HTTP client ends up using our custom transport.
+	} else {
+		return &http.Client{
+			Timeout:   120 * time.Second,
+			Transport: transport,
+		}
+	}
+}
+
+func CreateWebsocketConnection(ctx context.Context, conf config.ServerConfig, globalCollectionOpts state.CollectionOpts) (*websocket.Conn, *http.Response, error) {
+	proxyConfig := httpproxy.Config{
+		HTTPProxy:  conf.HTTPProxy,
+		HTTPSProxy: conf.HTTPSProxy,
+		NoProxy:    conf.NoProxy,
+	}
+	dialer := websocket.Dialer{
+		ReadBufferSize:  10240,
+		WriteBufferSize: 10240,
+		Proxy: func(req *http.Request) (*url.URL, error) {
+			return proxyConfig.ProxyFunc()(req.URL)
+		},
+	}
+	url, err := url.Parse(conf.APIBaseURL + "/v2/snapshots/websocket")
+	if err != nil {
+		return nil, nil, err
+	}
+	if url.Scheme == "http" {
+		url.Scheme = "ws"
+	} else {
+		url.Scheme = "wss"
+	}
+	headers := AuthHeaders(conf, globalCollectionOpts)
+	return dialer.DialContext(ctx, url.String(), headers)
+}
+
+func AuthHeaders(conf config.ServerConfig, globalCollectionOpts state.CollectionOpts) map[string][]string {
+	headers := make(map[string][]string)
+	headers["Pganalyze-Api-Key"] = []string{conf.APIKey}
+	headers["Pganalyze-System-Id"] = []string{conf.SystemID}
+	headers["Pganalyze-System-Type"] = []string{conf.SystemType}
+	headers["Pganalyze-System-Scope"] = []string{conf.SystemScope}
+	headers["Pganalyze-System-Id-Fallback"] = []string{conf.SystemIDFallback}
+	headers["Pganalyze-System-Type-Fallback"] = []string{conf.SystemTypeFallback}
+	headers["Pganalyze-System-Scope-Fallback"] = []string{conf.SystemScopeFallback}
+	if globalCollectionOpts.TestRun {
+		headers["Pganalyze-Test-Run"] = []string{"true"}
+	}
+	headers["User-Agent"] = []string{util.CollectorNameAndVersion}
+	return headers
+}
diff --git a/runner/websocket.go b/runner/websocket.go
index 5febe66b..b7f1c0fd 100644
--- a/runner/websocket.go
+++ b/runner/websocket.go
@@ -5,15 +5,12 @@ import (
 	"compress/zlib"
 	"context"
 	"io"
-	"net/http"
-	"net/url"
 	"time"
 
 	"github.com/gorilla/websocket"
 	"github.com/pganalyze/collector/output/pganalyze_collector"
-	"golang.org/x/net/http/httpproxy"
 	"google.golang.org/protobuf/proto"
 )
 
@@ -38,34 +35,7 @@ func SetupWebsocketForAllServers(ctx context.Context, servers []*state.Server, g
 
 func connect(ctx context.Context, server *state.Server, globalCollectionOpts state.CollectionOpts, logger *util.Logger) {
 	connCtx, cancelConn := context.WithCancel(ctx)
-	proxyConfig := httpproxy.Config{
-		HTTPProxy:  server.Config.HTTPProxy,
-		HTTPSProxy: server.Config.HTTPSProxy,
-		NoProxy:    server.Config.NoProxy,
-	}
-	dialer := websocket.Dialer{
-		ReadBufferSize:  10240,
-		WriteBufferSize: 10240,
-		Proxy: func(req *http.Request) (*url.URL, error) {
-			return proxyConfig.ProxyFunc()(req.URL)
-		},
-	}
-	url, _ := url.Parse(server.Config.APIBaseURL + "/v2/snapshots/websocket")
-	if url.Scheme == "http" {
-		url.Scheme = "ws"
-	} else {
-		url.Scheme = "wss"
-	}
-	headers := make(map[string][]string)
-	headers["Pganalyze-Api-Key"] = []string{server.Config.APIKey}
-	headers["Pganalyze-System-Id"] = []string{server.Config.SystemID}
-	headers["Pganalyze-System-Type"] = []string{server.Config.SystemType}
-	headers["Pganalyze-System-Scope"] = []string{server.Config.SystemScope}
-	headers["Pganalyze-System-Id-Fallback"] = []string{server.Config.SystemIDFallback}
-	headers["Pganalyze-System-Type-Fallback"] = []string{server.Config.SystemTypeFallback}
-	headers["Pganalyze-System-Scope-Fallback"] = []string{server.Config.SystemScopeFallback}
-	headers["User-Agent"] = []string{util.CollectorNameAndVersion}
-	conn, response, err := dialer.DialContext(connCtx, url.String(), headers)
+	conn, response, err := CreateWebsocketConnection(connCtx, server.Config, globalCollectionOpts)
 	if err != nil {
 		cancelConn()
 		logger.PrintWarning("Error starting websocket: %s %v", err, response)

**Comment by seanlinsley** [comment]
Okay, I believe all of the review comments have been addressed. I also fixed the issue where quitting the collector through Control-C showed a websocket warning.

File: runner/websocket.go (hunk 21 above)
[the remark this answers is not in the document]
Comment (lfittl) [reply runner/websocket.go]: Meh, that's annoying re: circular dependency. Agreed we can do this in a separate PR.

File: main.go [hunk 22]
```diff
@@ -397,7 +400,7 @@ func main() {
 		CollectSystemInformation: !noSystemInformation,
 		StateFilename:            stateFilename,
 		WriteStateUpdate:         (!dryRun && !dryRunLogs && !testRun) || forceStateUpdate,
-		ForceEmptyGrant:          dryRun || dryRunLogs || benchmark,
+		ForceEmptyGrant:          dryRun || dryRunLogs || testRunLogs || benchmark,
```
seanlinsley, line 403: This was needed for the test suite to pass. It's not clear from the `main.go` documentation what the difference is between `--test-logs` and `--dry-run-logs`. Maybe one of them should be dropped?

lfittl:
>     * If unable to start a WebSocket connection, the collector falls back to the legacy API.

Not sure where exactly, but it appears there is a problem with the fallback, at least in my local test right now (on the latest tip of this branch) I got the following error, running against the regular server without websocket support:

```
% ./pganalyze-collector --test
2024/10/01 09:42:57 I Running collector test with pganalyze-collector 0.58.0
2024/10/01 09:42:57 W [server2] Checking log_line_prefix: database (%d) and user (%u) not found: pganalyze will not be able to correctly classify some log lines
2024/10/01 09:42:57 I [server2] Testing statistics collection...
2024/10/01 09:42:57 I [server1] Testing statistics collection...
2024/10/01 09:42:57 W [server1] Error starting websocket: websocket: bad handshake &{404 Not Found 404 HTTP/1.1 1 1 map[Content-Length:[36] Content-Type:[text/plain; charset=utf-8] Date:[Tue, 01 Oct 2024 13:42:57 GMT]] {0x140005d40c0} 36 [] false false map[] 0x14000764d00 <nil>}
2024/10/01 09:42:57 W [server2] Error starting websocket: websocket: bad handshake &{404 Not Found 404 HTTP/1.1 1 1 map[Content-Length:[36] Content-Type:[text/plain; charset=utf-8] Date:[Tue, 01 Oct 2024 13:42:57 GMT]] {0x140004d2b70} 36 [] false false map[] 0x14000578400 <nil>}
2024/10/01 09:42:57 I [server2]   Test collection successful for PostgreSQL 16.3 on aarch64-darwin, compiled by clang-14.0.0, 64-bit
2024/10/01 09:42:57 E [server2] Error uploading snapshot: Error - can't upload without valid S3 grant
2024/10/01 09:42:57 E [server2] Could not process server: Error - can't upload without valid S3 grant
2024/10/01 09:42:58 I [server1]   Test collection successful for PostgreSQL 16.3 (Debian 16.3-1.pgdg120+1) on aarch64-unknown-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit
2024/10/01 09:42:58 E [server1] Error uploading snapshot: Error - can't upload without valid S3 grant
2024/10/01 09:42:58 E [server1] Could not process server: Error - can't upload without valid S3 grant
2024/10/01 09:42:58 I [server2] Testing activity snapshots...
2024/10/01 09:42:58 I [server1] Testing activity snapshots...
2024/10/01 09:42:58 I [server2]   Test snapshot successful
2024/10/01 09:42:58 I [server1]   Test snapshot successful
```

(note the "Error - can't upload without valid S3 grant", which I don't get when running on `main`)

main.go
lfittl: Hmm, so I think the different is:

* `--test-logs` is supposed to get a log grant, and then either do a log download or log tail
* `--dry-run-logs` is supposed to work without an API server, so skips the grant

However, I'm not sure if the distinction matters.

Really the main use case is `--test-logs`, which we document, to verify whether the log setup is correct. I suppose doing a log grant could be beneficial for Enterprise environments (since I think it would trigger the server side to check for correct LOG_ENCRYPTION_KEY / bucket setup?). Though that's soon to be removed anyway.

msakrejda: I'm not sure what the right course of action is here. Is there a blocker?

msakrejda (approved):
Assuming the `ForceEmptyGrant` change is okay (comment inline), this looks good.

main.go
lfittl: FWIW, I also don't think this is a blocker. I haven't seen much benefit of `--dry-run-logs`, so I'm good with dropping it.

That said, the regular `--dry-run` is useful though (it helps get clarity on what data the collector sent), so we shouldn't be messing with that, I think :)

seanlinsley: Sorry, I don't think I explained my reasoning here well. I'm thinking that there's little benefit to actually submitting test snapshots to the API, so in dropping `--dry-run-logs` my intention was to make `--test-logs` a dry run command. Then we only need to contact the API for a top-level `--test`.

PR #585, opened 2024-08-19
Add note about gcc to CONTRIBUTING.md

msakrejda:


CONTRIBUTING.md
```diff
@@ -1,11 +1,15 @@
 ## Contributing Instructions
 
-Pull the project with submodules.
+Pull the project.
```
msakrejda, line 3: As of #581, the project no longer uses submodules, so updated this in passing.

seanlinsley (approved):

PR #586, opened 2024-08-20
Support Query Parameters with auto_explain and support new lines better with text format

keiko713:
Starting from Postgres 16, auto_explain outputs the query parameters (bind parameters), used for prepared statements.
With this PR, the collector supports to extract these parameters and send them within query samples, with both JSON and TEXT format.

This PR also improves new lines support with text format. Previously, the text format is not parsing a query text and an EXPLAIN output well, and if the query text contains new lines, it was simply failing to extract them correctly.

With this PR, first we try to match the text format auto_explain with a newly introduced "Query Parameters:", which makes our life easier to actually separate a query text and an EXPLAIN output.
If that doesn't go well, we try to match with the cost output:
https://github.com/postgres/postgres/blob/4d93bbd4e0d414a33521f62d6249ac88486c866f/src/backend/commands/explain.c#L1815
While the most of auto_explain should have the cost output (it's turned on by default), you can also turn it off. In that case, use the existing logic as a fallback.
This way, most of the users should see the improved support even if text format + query text new lines are used.

### DIFF
--- a/logs/analyze_test.go
+++ b/logs/analyze_test.go
@@ -45,7 +45,7 @@ var tests = []testpair{
 			Content:  "duration: 4079.697 ms  execute <unnamed>: \nSELECT * FROM x WHERE y = $1 LIMIT $2",
 			LogLevel: pganalyze_collector.LogLineInformation_LOG,
 		}, {
-			Content:  "parameters: $1 = 'long string', $2 = '1'",
+			Content:  "parameters: $1 = 'o''clock', $2 = '1'",
 			LogLevel: pganalyze_collector.LogLineInformation_DETAIL,
 		}},
 		[]state.LogLine{{
@@ -63,11 +63,11 @@ var tests = []testpair{
 			ReviewedForSecrets: true,
 			SecretMarkers: []state.LogSecretMarker{{
 				ByteStart: 18,
-				ByteEnd:   29,
+				ByteEnd:   26,
 				Kind:      state.StatementParameterLogSecret,
 			}, {
-				ByteStart: 38,
-				ByteEnd:   39,
+				ByteStart: 35,
+				ByteEnd:   36,
 				Kind:      state.StatementParameterLogSecret,
 			}},
 		}},
@@ -75,7 +75,7 @@ var tests = []testpair{
 			Query:     "SELECT * FROM x WHERE y = $1 LIMIT $2",
 			RuntimeMs: 4079.697,
 			Parameters: []null.String{
-				null.StringFrom("long string"),
+				null.StringFrom("o''clock"),
 				null.StringFrom("1"),
 			},
 		}},
--- a/logs/analyze_test.go
+++ b/logs/analyze_test.go
@@ -4145,6 +4145,54 @@ var tests = []testpair{
 				"          Index Cond: (pgbench_branches.bid = 59)",
 		}},
 	},
+	{
+		[]state.LogLine{{
+			Content: "duration: 0.006 ms  plan:\n" +
+				"\t{\n" +
+				"\t  \"Query Text\": \"PREPARE show_values (text, uuid, integer, text) AS SELECT $1, $2, $3, $4;\",\n" +
+				"\t  \"Query Parameters\": \"$1 = 'o''clock', $2 = '13f82cb8-a0cf-424e-903a-6f8d7db83191', $3 = '1234', $4 = NULL\",\n" +
+				"\t  \"Plan\": {\n" +
+				"\t    \"Node Type\": \"Result\",\n" +
+				"\t    \"Parallel Aware\": false,\n" +
+				"\t    \"Async Capable\": false,\n" +
+				"\t    \"Startup Cost\": 0.00,\n" +
+				"\t    \"Total Cost\": 0.01,\n" +
+				"\t    \"Plan Rows\": 1,\n" +
+				"\t    \"Plan Width\": 84\n" +
+				"\t  }\n" +
+				"\t}\n",
+		}},
+		[]state.LogLine{{
+			Query:              "PREPARE show_values (text, uuid, integer, text) AS SELECT $1, $2, $3, $4;",
+			Classification:     pganalyze_collector.LogLineInformation_STATEMENT_AUTO_EXPLAIN,
+			ReviewedForSecrets: true,
+			SecretMarkers: []state.LogSecretMarker{{
+				ByteStart: 27,
+				ByteEnd:   436,
+				Kind:      state.StatementTextLogSecret,
+			}},
+		}},
+		[]state.PostgresQuerySample{{
+			Query:         "PREPARE show_values (text, uuid, integer, text) AS SELECT $1, $2, $3, $4;",
+			RuntimeMs:     0.006,
+			HasExplain:    true,
+			ExplainSource: pganalyze_collector.QuerySample_AUTO_EXPLAIN_EXPLAIN_SOURCE,
+			ExplainFormat: pganalyze_collector.QuerySample_JSON_EXPLAIN_FORMAT,
+			Parameters:    []null.String{null.StringFrom("o''clock"), null.StringFrom("13f82cb8-a0cf-424e-903a-6f8d7db83191"), null.StringFrom("1234"), null.NewString("", false)},
+			ExplainOutputJSON: &state.ExplainPlanContainer{
+				Plan: []byte("{\n" +
+					"\t    \"Node Type\": \"Result\",\n" +
+					"\t    \"Parallel Aware\": false,\n" +
+					"\t    \"Async Capable\": false,\n" +
+					"\t    \"Startup Cost\": 0.00,\n" +
+					"\t    \"Total Cost\": 0.01,\n" +
+					"\t    \"Plan Rows\": 1,\n" +
+					"\t    \"Plan Width\": 84\n" +
+					"\t  }"),
+				QueryParameters: "$1 = 'o''clock', $2 = '13f82cb8-a0cf-424e-903a-6f8d7db83191', $3 = '1234', $4 = NULL",
+			},
+		}},
+	},
 	// pganalyze-collector-identify
 	{
 		[]state.LogLine{{
--- a/logs/analyze_test.go
+++ b/logs/analyze_test.go
@@ -4193,6 +4193,68 @@ var tests = []testpair{
 			},
 		}},
 	},
+	// new lines
+	{
+		[]state.LogLine{{
+			Content: "duration: 1681.452 ms  plan:\n" +
+				"\tQuery Text: PREPARE pga_keys (text) AS\n" +
+				"\t    SELECT * FROM api_keys WHERE access_scope_type = $1\n" +
+				"\t    AND organization_id = '13f82cb8-a0cf-424e-903a-6f8d7db83191';\n" +
+				"\tSeq Scan on api_keys  (cost=0.00..11.93 rows=1 width=132)\n" +
+				"\t  Filter: ((access_scope_type = 'organization'::text) AND (organization_id = '13f82cb8-a0cf-424e-903a-6f8d7db83191'::uuid))",
+		}},
+		[]state.LogLine{{
+			Query:              "PREPARE pga_keys (text) AS\n\t    SELECT * FROM api_keys WHERE access_scope_type = $1\n\t    AND organization_id = '13f82cb8-a0cf-424e-903a-6f8d7db83191';",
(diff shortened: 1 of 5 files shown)


logs/querysample/querysample.go
```diff
@@ -116,3 +133,19 @@ func TransformLogMinDurationStatementToQuerySample(logLine state.LogLine, queryT
 	}
 	return sample, true
 }
+
+func findQueryParameters(paramText string) []null.String {
+	// Handle Query Parameters (available from Postgres 16+)
+	var parameters []null.String
+	// Regular expression to find all values in single quotes or NULL
+	// Query Parameters example: $1 = 'foo', $2 = '123', $3 = NULL, $4 = 'bo''o'
```
keiko713, line 141: note: params are always quoted, regardless of the type (e.g. $2 is passed as integer but the output is quoted)

https://github.com/postgres/postgres/blob/4d93bbd4e0d414a33521f62d6249ac88486c866f/src/backend/nodes/params.c#L379

logs/querysample/querysample.go
```diff
@@ -51,38 +51,55 @@ func transformExplainJSONToQuerySample(logLine state.LogLine, explainText string
...
 	}, nil
 }
 
 var autoExplainTextPlanDetailsRegexp = regexp.MustCompile(`^Query Text: (.+)\s+([\s\S]+)`)
 var herokuAutoExplainWithTabRegexp = regexp.MustCompile(`^Query Text: ([^\t]+)\t([\s\S]+)`)
+var autoExplainTextWithQueryParametersRegexp = regexp.MustCompile(`^Query Text: ([\s\S]+)\r?\n\s*Query Parameters: (.+)\r?\n\s*([\s\S]+)`)
+var autoExplainTextWithCostsRegexp = regexp.MustCompile(`^Query Text: ([\s\S]+?)\r?\n\s*([\S ]+  \(cost=\d+\.\d{2}\.\.\d+\.\d{2} rows=\d+ width=\d+\)[\s\S]+)`)
 
 func transformExplainTextToQuerySample(logLine state.LogLine, explainText string, queryRuntimeMs float64) (state.PostgresQuerySample, error) {
-	explainParts := autoExplainTextPlanDetailsRegexp.FindStringSubmatch(explainText)
```
keiko713, line 61: note: might be better to review with hide whitespace here

logs/analyze_test.go
```diff
@@ -4145,6 +4145,116 @@ var tests = []testpair{
...
+				Plan: []byte("{\n" +
+					"\t    \"Node Type\": \"Result\",\n" +
+					"\t    \"Parallel Aware\": false,\n" +
+					"\t    \"Async Capable\": false,\n" +
+					"\t    \"Startup Cost\": 0.00,\n" +
+					"\t    \"Total Cost\": 0.01,\n" +
+					"\t    \"Plan Rows\": 1,\n" +
+					"\t    \"Plan Width\": 84\n" +
+					"\t  }"),
+				QueryParameters: "$1 = 'o''clock', $2 = '13f82cb8-a0cf-424e-903a-6f8d7db83191', $3 = '1234', $4 = NULL",
```
keiko713, line 4192: if normalized is set, remove the QueryParameters key

keiko713: https://github.com/pganalyze/collector/blob/768c8019ab80d9596326b8e976b7371c7f91fc45/runner/logs.go#L324

logs/querysample/querysample.go
```diff
@@ -116,3 +133,19 @@ func TransformLogMinDurationStatementToQuerySample(logLine state.LogLine, queryT
...
 	return sample, true
 }
+
+func findQueryParameters(paramText string) []null.String {
+	// Handle Query Parameters (available from Postgres 16+)
+	var parameters []null.String
+	// Regular expression to find all values in single quotes or NULL
+	// Query Parameters example: $1 = 'foo', $2 = '123', $3 = NULL, $4 = 'bo''o'
+	re := regexp.MustCompile(`'((?:[^']|'')*)'|NULL`)
+	for _, part := range re.FindAllString(paramText, -1) {
```
keiko713, line 143: check if this can be somehow combined with https://github.com/pganalyze/collector/blob/768c8019ab80d9596326b8e976b7371c7f91fc45/logs/querysample/querysample.go#L108

keiko713: well, not really, so I'll keep this as is

logs/querysample/querysample.go
```diff
@@ -116,3 +133,19 @@ func TransformLogMinDurationStatementToQuerySample(logLine state.LogLine, queryT
...
+		if part == "NULL" {
+			parameters = append(parameters, null.NewString("", false))
+		} else {
+			parameters = append(parameters, null.StringFrom(strings.Trim(part, "'")))
```
msakrejda, line 147: Does this mean that for parameters with escaped quotes (e.g., `o''clock`), we store them with those escapes (i.e., the string that ends up in the snapshot is `o''clock` rather than the actual parameter value `o'clock`? Is that what we are doing for `log_min_duration_statement` parameters? I know that the code here is similar to the code you linked above, but I tried to trace back where the parameter values come from there, checking to confirm how escaped quotes are handled there, and I got lost. Are we consistent?

msakrejda (approved):

logs/querysample/querysample.go
```diff
@@ -51,38 +51,55 @@ func transformExplainJSONToQuerySample(logLine state.LogLine, explainText string
...
 		ExplainFormat:     pganalyze_collector.QuerySample_JSON_EXPLAIN_FORMAT,
 		ExplainOutputJSON: &explainJSONOutput,
+		Parameters:        findQueryParameters(explainJSONOutput.QueryParameters),
 	}, nil
 }
 
 var autoExplainTextPlanDetailsRegexp = regexp.MustCompile(`^Query Text: (.+)\s+([\s\S]+)`)
 var herokuAutoExplainWithTabRegexp = regexp.MustCompile(`^Query Text: ([^\t]+)\t([\s\S]+)`)
+var autoExplainTextWithQueryParametersRegexp = regexp.MustCompile(`^Query Text: ([\s\S]+)\r?\n\s*Query Parameters: (.+)\r?\n\s*([\s\S]+)`)
+var autoExplainTextWithCostsRegexp = regexp.MustCompile(`^Query Text: ([\s\S]+?)\r?\n\s*([\S ]+  \(cost=\d+\.\d{2}\.\.\d+\.\d{2} rows=\d+ width=\d+\)[\s\S]+)`)
```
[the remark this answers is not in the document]
keiko713: Agree that it's easier to follow, I'll update 👍 Just a background, I did `{2}` since Postgres is using  `(cost=%.2f..%.2f rows=%.0f width=%d)`.

https://github.com/postgres/postgres/blob/4d93bbd4e0d414a33521f62d6249ac88486c866f/src/backend/commands/explain.c#L1815

logs/querysample/querysample.go
```diff
@@ -51,38 +51,55 @@ func transformExplainJSONToQuerySample(logLine state.LogLine, explainText string
...
+			explainParts = autoExplainTextPlanDetailsRegexp.FindStringSubmatch(explainText)
+			if len(explainParts) != 3 {
+				return state.PostgresQuerySample{}, fmt.Errorf("auto_explain output doesn't match expected format")
+			}
+		}
+
+		// If EXPLAIN output's first char is not a capital letter (e.g. not something like "Update on" or "Index Scan"),
+		// likely it's hitting the Heroku's newline break in "Query Text:" chunk
+		// Handle the separation of the query and the explain output text with the tab for these cases
+		// (this can be retired with the autoExplainTextWithCostsRegexp, but leaving here for the EXPLAIN without costs case)
```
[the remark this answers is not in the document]
keiko713: Actually, looking into more, auto_explain simply doesn't allow turning off the costs, so I don't need to worry about it.
Honestly I'm still not super confident that a new way works 100% but maybe we can retire the old way. Will update.

keiko713:
@msakrejda Thanks for reviewing! Given that auto_explain will always have costs, I retired the previous logic entirely.
I'm really not trusting myself enough to feel comfortable with this, but also I don't have enough counterarguments to keep the old way, so I think we can retire these logic.
Can you please check again if this is looking good?

msakrejda:
Sorry, this slipped off my radar. Will check again today.

logs/querysample/querysample.go
```diff
@@ -51,38 +50,39 @@ func transformExplainJSONToQuerySample(logLine state.LogLine, explainText string
 		ExplainSource:     pganalyze_collector.QuerySample_AUTO_EXPLAIN_EXPLAIN_SOURCE,
 		ExplainFormat:     pganalyze_collector.QuerySample_JSON_EXPLAIN_FORMAT,
 		ExplainOutputJSON: &explainJSONOutput,
+		Parameters:        findQueryParameters(explainJSONOutput.QueryParameters),
 	}, nil
 }
 
-var autoExplainTextPlanDetailsRegexp = regexp.MustCompile(`^Query Text: (.+)\s+([\s\S]+)`)
-var herokuAutoExplainWithTabRegexp = regexp.MustCompile(`^Query Text: ([^\t]+)\t([\s\S]+)`)
+var autoExplainTextWithQueryParametersRegexp = regexp.MustCompile(`^Query Text: ([\s\S]+)\r?\n\s*Query Parameters: (.+)\r?\n\s*([\s\S]+)`)
```
msakrejda, line 57: How is `[\s\S]` different from `.`?

logs/querysample/querysample.go
msakrejda: Still wondering about this.

msakrejda (approved):
Re-reviewed, and I still have the outstanding question about quote handling, but otherwise looks good.

logs/querysample/querysample.go
keiko713: With golang, `.` doesn't match with a new line. As "Query Parameters" line is known to be one line (even if the param contains new line, it'll be `\n` instead of being an actual new line there), so I'm matching with `(.+)` there instead of doing `([\s\S]+)`, to make sure that it'll match just a Query Parameters line, and not the other lines.
https://pkg.go.dev/regexp/syntax

logs/querysample/querysample.go
keiko713: You don't realize that you have 2 weeks old pending comment until someone else wonders that you haven't said anything 🙃

keiko713: Interestingly, it does look like I commented before Maciek commented, but for the record, Maciek did wonder before I commented.

msakrejda: Oh whoops, I'm not sure how I missed that. I think GH did not refresh the comments when I was doing my review somehow.

> Regarding storing param data as is (o''clock) or not (o'clock), I prefer to keep it as is. Honestly, nothing is going to be perfect around here (e.g. we don't have types, some binary params will act weirdly, etc.), so I'd like to start from somewhere. If we found some inconvenience, we can always come back here to improve things.

Okay, but I do think we need to be really careful here. This is the kind of thing that leads to SQL injection, and while that's not a problem in this situation, I think giving the impression that we don't know how to handle SQL injection could make our customers lose trust.

msakrejda (approved):

logs/querysample/querysample.go
keiko713: I see, that's a good point. Basically, we'll need to do the opposite of this here to get the actual value:
https://github.com/postgres/postgres/blob/b4db64270e0c18f72c9c7bf214c809949875a85a/src/backend/nodes/params.c#L321-L336

Though, given that we are only going to pass this value as a string to pganalyze side, how much we're going to mutate this value in the collector side would be question. IOW, if we pass this as is, we could treat these parameters as "param log string" (not a param) and do all the potential parsing in pganalyze app side. With that, I think we can say that we're not neglecting the SQL injection aspect.

-----

Regarding "unescape", here is some example of complex parameter value:

```
$4 = '{\"(12345,-56789,111111,\\\"\\\\\\\\x041d44a78f95610f\\\",c15c4fd4-f643-42db-ac03-9335a6bd78fc)\"}'
```

It's hard to read, but it's one element array of record. The record is with the first 3 values being integer, 4th value is bytea, 5th value is uuid. This is an extreme example (note: this is an argument of function, FYI), but how should we treat the case like this? I think we need to do some dirty thing case by case, and if we want to reuse this as a parameter to be used for the EXPLAIN command, more manipulations are needed, which I feel like good to have it in pganalyze side, not in the collector side.

msakrejda: Great point. I think doing any filtering on the receiving side is a much better idea. That way, any updates or fixes to the parameter filtering logic can be deployed immediately, without having to wait for users to upgrade their collectors.

PR #587, opened 2024-08-21
Improve Azure log handling to support Flexible Server and Cosmos DB better

keiko713:
This PR fixes two things of Azure log handling (Log Insights):

1. Handle the case Flexible Server returns `LogicalServerName` in record
   * This bug was found in https://github.com/pganalyze/collector/pull/494#issuecomment-2297975788
   * Previously, it was assumed that it won't be returned with Flexible Server
   * Instead, use `resourceId`, which contains resource type, to identify if it's single server or not
2. Fix parse errors with Cosmos DB logs
   * Lukas mentioned that this was reported by the customer that the current logic doesn't work with Cosmos DB
   * Confirmed that it doesn't work due to how Cosmos DB doesn't add log level after the prefix
   * Applied some workaround as a fix (similar to Flexible Server but a bit more complex)

### DIFF
--- a/input/system/azure/logs.go
+++ b/input/system/azure/logs.go
@@ -44,6 +44,17 @@ type AzurePostgresLogRecord struct {
 	Properties        AzurePostgresLogMessage `json:"properties"`
 }
 
+func (record *AzurePostgresLogRecord) IsSingleServer() bool {
+	resourceParts := strings.Split(record.ResourceID, "/")
+	// ResourceID's second to last element is a resource type
+	return strings.ToLower(resourceParts[len(resourceParts)-2]) == "servers"
+}
+
+func (record *AzurePostgresLogRecord) IsCosmosDB() bool {
+	resourceParts := strings.Split(record.ResourceID, "/")
+	return strings.ToLower(resourceParts[len(resourceParts)-2]) == "servergroupsv2"
+}
+
 type AzureEventHubData struct {
 	Records []AzurePostgresLogRecord `json:"records"`
 }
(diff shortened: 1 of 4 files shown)


input/system/azure/logs.go
```diff
@@ -256,6 +267,13 @@ func ParseRecordToLogLines(in AzurePostgresLogRecord, parser state.LogParser) ([
...
 		// Single Server (but our parser expects them together)
 		logLineContent = fmt.Sprintf("%s%s:  %s", in.Properties.Prefix, in.Properties.ErrorLevel, logLineContent)
+	} else if in.IsCosmosDB() { // Cosmos DB
+		prefix, content, ok := parser.GetPrefixAndContent(logLineContent)
+		if ok {
+			// Cosmos DB doesn't output the log level after the log_line_prefix and before the content
+			// Manually adding it between them so the parser can parse the log line
+			logLineContent = fmt.Sprintf("%s%s:  %s", prefix, in.Properties.ErrorLevel, content)
+		}
 	}
```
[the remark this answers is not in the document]
keiko713: Yeah for flexible server, there is no need for a tweak. I will add a comment 👍

input/system/azure/logs.go
```diff
@@ -229,20 +240,20 @@ func SetupLogSubscriber(ctx context.Context, wg *sync.WaitGroup, globalCollectio
...
-		return in.LogicalServerName
 	}
 }
 
 // Parses one Azure Event Hub log record into one or two log lines (main + DETAIL)
 func ParseRecordToLogLines(in AzurePostgresLogRecord, parser state.LogParser) ([]state.LogLine, error) {
 	logLineContent := in.Properties.Message
 
-	if in.LogicalServerName != "" { // Single Server
+	if in.IsSingleServer() { // Single Server
```
msakrejda, line 256: Now that you have helper methods here, the comments are superfluous.
```suggestion
	if in.IsSingleServer() {
```

input/system/azure/logs.go
```diff
@@ -256,6 +267,15 @@ func ParseRecordToLogLines(in AzurePostgresLogRecord, parser state.LogParser) ([
 		// Add prefix and error level, which are separated from the content on
 		// Single Server (but our parser expects them together)
 		logLineContent = fmt.Sprintf("%s%s:  %s", in.Properties.Prefix, in.Properties.ErrorLevel, logLineContent)
+	} else if in.IsCosmosDB() { // Cosmos DB
```
msakrejda, line 270: As above
```suggestion
	} else if in.IsCosmosDB() {
```

msakrejda (approved):

PR #589, opened 2024-08-27
Error out on incomplete config sections

msakrejda:
A customer ran into this because they specified some configuration settings in uppercase (`DB_NAME`). We currently silently skip incomplete sections, which is pretty confusing.

Separately, I considered parsing config keys in a case-insensitive manner, but since some but not all of our environment variable keys are identical to our config file keys, I thought this would be more confusing than a clear error. I'm open to revisiting that decision.

### DIFF
--- a/config/read.go
+++ b/config/read.go
@@ -782,15 +782,11 @@ func Read(logger *util.Logger, filename string) (Config, error) {
 			}
 
 			// Ensure we have no duplicate identifiers within one collector
-			skip := false
 			for _, server := range conf.Servers {
 				if config.Identifier == server.Identifier {
-					skip = true
+					logger.PrintError("Skipping config section %s, detected as duplicate. Note: To monitor multiple databases on the same server, db_name accepts a comma-separated list.", config.SectionName)
+					continue
 				}
-			}
-			if skip {
-				logger.PrintError("Skipping config section %s, detected as duplicate. Note: To monitor multiple databases on the same server, db_name accepts a comma-separated list.", config.SectionName)
-				continue
 			}
 
 			conf.Servers = append(conf.Servers, *config)
--- a/config/read.go
+++ b/config/read.go
@@ -766,8 +766,16 @@ func Read(logger *util.Logger, filename string) (Config, error) {
 				return conf, err
 			}
 
+			if config.DbURL != "" {
+				_, err := url.Parse(config.DbURL)
+				if err != nil {
+					logger.PrintError("Could not parse db_url in section %s; check URL format and note that any special characters must be percent-encoded", config.SectionName)
+				}
+			}
+
 			if config.GetDbName() == "" {
-				return conf, fmt.Errorf("No connection info found for section %s; see https://pganalyze.com/docs/collector/settings", sectionName)
+				logger.PrintError("No connection info found for section %s; see https://pganalyze.com/docs/collector/settings", sectionName)
+				continue
 			}
 
 			config.SectionName = sectionName
@@ -786,13 +794,6 @@ func Read(logger *util.Logger, filename string) (Config, error) {
 				if config.Identifier == server.Identifier {
 					logger.PrintError("Skipping config section %s, detected as duplicate. Note: To monitor multiple databases on the same server, db_name accepts a comma-separated list.", config.SectionName)
 					continue
-				}
-			}
-
-			if config.DbURL != "" {
-				_, err := url.Parse(config.DbURL)
-				if err != nil {
-					return conf, fmt.Errorf("Could not parse db_url in section %s; check URL format and note that any special characters must be percent-encoded", config.SectionName)
 				}
 			}
 
--- a/config/read.go
+++ b/config/read.go
@@ -800,7 +800,7 @@ func Read(logger *util.Logger, filename string) (Config, error) {
 		}
 
 		if len(conf.Servers) == 0 {
-			return conf, fmt.Errorf("Configuration file is empty, please edit %s and reload the collector", filename)
+			return conf, fmt.Errorf("Configuration contains no valid servers, please edit %s and reload the collector", filename)
 		}
 	} else {
 		if util.IsHeroku() {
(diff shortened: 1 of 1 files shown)


config/read.go
```diff
@@ -771,20 +781,19 @@ func Read(logger *util.Logger, filename string) (Config, error) {
...
 				}
 			}
+			if skip {
+				logger.PrintError("Skipping config section %s, detected as duplicate. Note: To monitor multiple databases on the same server, db_name accepts a comma-separated list.", config.SectionName)
+				continue
+			}
+
+			conf.Servers = append(conf.Servers, *config)
 
 			if config.DbURL != "" {
```
msakrejda, line 798: It's a little weird that we do this check _after_ we already process and append the server to the server list. I didn't touch it, but I was tempted to. Thoughts?

msakrejda: Oh also, we only log an error there, instead of returning an error. I don't understand why.

msakrejda:
The `reload` integation test is failing; I'm not sure why, but I can reproduce it. Investigating.

config/read.go
keiko713: Ask Maciek 4 years ago?
https://github.com/pganalyze/collector/commit/ac45be47c7b525a9fad9928e3398dba71f332319

> Validate at startup, but allow to continue to avoid retrofitting error
handling across helpers that right now cannot return errors. The helpers
return dummy/default values which will fail authentication.

Though honestly reading these, I still don't understand well for why we can't return the error here.

config/read.go
```diff
@@ -748,6 +748,11 @@ func Read(logger *util.Logger, filename string) (Config, error) {
 
 		sections := configFile.Sections()
 		for _, section := range sections {
+			sectionName := section.Name()
+			if sectionName == "pganalyze" || sectionName == "DEFAULT" {
```
keiko713, line 752:
```suggestion
			if sectionName == "pganalyze" || sectionName == ini.DefaultSection {
```

Is default section really handled already? In which part it's handled?

keiko713 (commented):
I think it's a good direction, but have some questions.

config/read.go
msakrejda: Oh man, that guy. He writes the worst code. And the guy who approved that: he'll approve anything!

I'm pretty sure the reason it ended up being a `PrintError` message originally is that the error above, about duplicate config sections, is handled with `PrintError`. On reflection, (1) that case is a little different, so not returning an error makes a little more sense and (2) it still does not make enough sense to do it that way, and we should change the duplicate section handling to return an error.

But I think changing duplicate section handling could break some currently-working configurations, so I'm not super excited about doing that. Whereas in this case, if you don't have a valid DB URL, you're not going to get very far with the collector, so I think it's safe to error out earlier. I'll update the code to do that.

config/read.go
msakrejda: Ah, good catch that a constant exists for this. It's handled in the sense that we don't use it. It will go through this loop but not really do anything. I think it's clearer to skip it explicitly. I'll update the comment to reflect that more clearly.

config/read.go
msakrejda: Okay, updated.

msakrejda:
This now include some other minor cleanup, in separate commits.

keiko713 (approved):
Nice, looking good 👍

config/read.go
```diff
@@ -771,32 +781,26 @@ func Read(logger *util.Logger, filename string) (Config, error) {
...
+					continue
 				}
 			}
 
 			if config.DbURL != "" {
 				_, err := url.Parse(config.DbURL)
 				if err != nil {
-					prefixedLogger := logger.WithPrefix(config.SectionName)
-					prefixedLogger.PrintError("Could not parse db_url; check URL format and note that any special characters must be percent-encoded")
+					return conf, fmt.Errorf("Could not parse db_url in section %s; check URL format and note that any special characters must be percent-encoded", config.SectionName)
```
lfittl, line 795: I think in practice this would never be hit, since `config.GetDbName()` ends up being empty in such cases ( (https://github.com/pganalyze/collector/blob/main/config/config.go#L459), leading to an early return a few lines up.

config/read.go
```diff
@@ -760,7 +765,12 @@ func Read(logger *util.Logger, filename string) (Config, error) {
 			if err != nil {
 				return conf, err
 			}
-			config.SectionName = section.Name()
+
+			if config.GetDbName() == "" {
+				return conf, fmt.Errorf("No connection info found for section %s; see https://pganalyze.com/docs/collector/settings", sectionName)
```
lfittl, line 770: Is an error for the whole file really the best way to deal with this?

I agree that we shouldn't silent when there are invalid URLs, but I think generally our approach is that if one section is invalid (e.g. we can't connect to the server), we still handle the other ones. I'm not sure I see the need to make this more than a `PrintError` (though I don't think this is a problem per se, just find the behavior inconsistent).

msakrejda: I think failing when reading the configuration file is the best time to fail, since someone is more likely to be interacting with the collector and likely to notice the error. I think something like "no connection info at all" is different than "could not connect to this server right now". But I don't feel strongly about this; I'll change this back to a PrintError.

config/read.go
msakrejda: Oh good catch. I can put it before the `GetDbName` check, right? I think it's still a useful error to have, since we can give more concrete guidance in this case.

msakrejda: And I guess similar to the other comment, I'll make this a `PrintError`.

config/read.go
lfittl: Yeah, I see your point - my main concern is that auto-generated config files might not have much interactivity, and one broken section shouldn't cause the whole file to fail (e.g. thinking someone didn't escape the password well, or something). I think having a PrintError is a clear improvement to what we have currently :)

config/read.go
lfittl: Yeah, I think you should be able to do that check early, and agreed, its useful to have.

PR #590, opened 2024-08-27
Release 0.58.0

msakrejda:
Other pending PRs:
 - #589
 - #588 ?
 - ~one coming up as a minor follow-up to #494~ (this was actually done in #494; I missed it when looking at the code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## 0.58.0      2024-08-27
+
+* Log Insights:
+  - New log parsing mechanism that is more performant and allows
+    for arbitrary log_line_prefix settings. The new parsing mechanism
+    is the default, but you can set `db_log_line_prefix = legacy` in
+    the config file or `DB_LOG_LINE_PREFIX=legacy` in the environment
+    to revert to the old mechanism.
+* RDS:
+  - Update AWS SDK to v1.55.3
+* Azure:
+  - Improve log handling for Flexible Server and Cosmos DB
+* Crunchy Bridge:
+  - fix error handling for error responses from Crunchy Bridge API
+
+
 ## 0.57.1      2024-07-17
 
 * Log Insights:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 0.58.0      2024-08-29
+## 0.58.0      2024-08-30
 
 * Log Insights: Revamp log parsing mechanism
   - The new mechanism is more performant and allows for arbitrary
@@ -19,6 +19,7 @@
   - Support log parsing for Azure Database for Cosmos DB Postgres
 * Crunchy Bridge:
   - Fix error handling for error responses from Crunchy Bridge API
+* Fix hang on exit with the `--discover-log-location` flag
 
 
 ## 0.57.1      2024-07-17
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
   - New log parsing mechanism that is more performant and allows
     for arbitrary log_line_prefix settings. The new parsing mechanism
     is the default, but you can set `db_log_line_prefix = legacy` in
-    the config file or `DB_LOG_LINE_PREFIX=legacy` in the environment
+    the config file or `LOG_LINE_PREFIX=legacy` in the environment
     to revert to the old mechanism.
 * RDS:
   - Update AWS SDK to v1.55.3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,18 +2,17 @@
 
 ## 0.58.0      2024-08-27
 
-* Log Insights:
-  - New log parsing mechanism that is more performant and allows
-    for arbitrary log_line_prefix settings. The new parsing mechanism
-    is the default, but you can set `db_log_line_prefix = legacy` in
-    the config file or `LOG_LINE_PREFIX=legacy` in the environment
-    to revert to the old mechanism.
-* RDS:
-  - Update AWS SDK to v1.55.3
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+## 0.58.0      2024-08-27
+
+* Log Insights:
+  - New log parsing mechanism that is more performant and allows
+    for arbitrary log_line_prefix settings. The new parsing mechanism
+    is the default, but you can set `db_log_line_prefix = legacy` in
+    the config file or `DB_LOG_LINE_PREFIX=legacy` in the environment
+    to revert to the old mechanism.
+* RDS:
+  - Update AWS SDK to v1.55.3
```
msakrejda, line 12: This affects https://github.com/pganalyze/pganalyze-docs/pull/250#discussion_r1578972209 , but is that something we want to mention here before we finish those docs?

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 # Changelog
 
+## 0.58.0      2024-08-27
+
+* Log Insights:
+  - New log parsing mechanism that is more performant and allows
+    for arbitrary log_line_prefix settings. The new parsing mechanism
+    is the default, but you can set `db_log_line_prefix = legacy` in
+    the config file or `DB_LOG_LINE_PREFIX=legacy` in the environment
```
msakrejda, line 9: Turns out the env part of this does not work yet: I missed that in #494. PR coming up.

CHANGELOG.md
keiko713: I don't think we need to advertise that we started supporting it. It's still nice to update docs (still in my TODO list but a little bit in the bottom), but yeah I'd say no need to mention.

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+    for arbitrary log_line_prefix settings. The new parsing mechanism
+    is the default, but you can set `db_log_line_prefix = legacy` in
+    the config file or `LOG_LINE_PREFIX=legacy` in the environment
+    to revert to the old mechanism.
+* RDS:
+  - Update AWS SDK to v1.55.3
+* Azure:
+  - Improve log handling for Flexible Server and Cosmos DB
+* Crunchy Bridge:
+  - fix error handling for error responses from Crunchy Bridge API
```
keiko713, line 16:
```suggestion
  - Fix error handling for error responses from Crunchy Bridge API
```

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 # Changelog
 
+## 0.58.0      2024-08-27
+
+* Log Insights:
```
keiko713, line 5: Might be nice to say something as a title, like "Log Insights: Update log parsing mechanism" or "Log Insights: Introduce a new log parsing mechanism". Uhmm both don't sound so good, but it'd be nice to have a title.

keiko713 (commented):
Pending review as we may update changelog more

CHANGELOG.md
lfittl: Maybe we can just note that EKS pod identity is now supported? I think for those that care about it they'd appreciate knowing its supported, and would if needed reach out if unclear how to utilize it.

keiko713: I wasn't sure if the setup flow goes well as I wasn't able to complete the test back then, hence the hesitation.
That said, it'll likely go well and the fact that the updated SDK version does support EKS Pod Identity credentials so mentioning it will not going to be a lie, so maybe it's fine to mention it.
FYI, I plan to test and update the docs after the collector release, as testing before the release is going to be a PITA.

msakrejda: I've changed this to
```
* RDS: Update AWS SDK to v1.55.3
  - This allows using EKS pod identity; documentation coming soon
```

CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
 # Changelog
 
+## 0.58.0      2024-08-27
```
seanlinsley, line 3: The release date should probably be updated

CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
...
+* Log Insights: Revamp log parsing mechanism
+  - The new mechanism is more performant and allows for arbitrary
+    log_line_prefix settings. The new parsing mechanism is the default, but you
+    can set `db_log_line_prefix = legacy` in the config file or
+    `LOG_LINE_PREFIX=legacy` in the environment to revert to the old mechanism.
+* Log Insights: Redact parameters from utility statements by default
+  - Statements like `CREATE USER u WITH PASSWORD 'passw0rd'` can leak sensitive
+    data into Log Insights, so they are now redacted by default. Note that these
+    statements are usually very fast, and are normally only logged with Postgres
+    configurations like `log_min_duration_statement = 0` that log every statement.
```
lfittl, line 14: I think that's somewhat confusing, since we actually skip over slow query logs if `log_min_duration_statement = 0` is set.

Maybe we can instead state that such statements are only likely to be logged in edge cases, like a lock wait problem relating to the statement?

CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
...
+* RDS: Update AWS SDK to v1.55.3
+  - This allows using EKS pod identity; documentation coming soon
+* Azure:
+  - Improve log handling for Flexible Server and Cosmos DB
```
lfittl, line 18: It may make sense to split them, since the CosmosDB for Postgres support is actually more significant here (it appears that just didn't work in the past).

Maybe:

```suggestion
  - Ensure correct log handling for all Flexible Server events (don't accidentally treat them as Single Server events)
  - Support log parsing for Azure Database for Cosmos DB Postgres
```

msakrejda:
We should probably also include https://github.com/pganalyze/collector/pull/592

PR #591, opened 2024-08-29
Collect system info and metrics for Azure

keiko713:
Collect system info and metrics for Azure Flexible Server and Cosmos DB for PostgreSQL. Note that we won't support this for Single Server as it's close to EOL.

This uses 3 different SDK to collect information:
* Flexible Server SDK: collect database server information, such as location
* Cosmos DB for PostgreSQL SDK: collect database server and cluster information
* Azure Monitor Query client module: collect system metrics, like CPU usage

To start using this, the customer needs to supply a new config variable `AZURE_SUBSCRIPTION_ID` to the collector. Also, they'll need to assign the managed identity to the collector VM (if not already with Log Insights), then grant the Read permission of the database to that managed identity. This requires docs update, likely including some setup process change.

When `AZURE_SUBSCRIPTION_ID` is not supplied, the collector doesn't error out and simply won't collect the system info.


input/system/azure/system.go
```diff
@@ -0,0 +1,302 @@
...
+	if err != nil {
+		server.SelfTest.MarkCollectionAspectError(state.CollectionAspectSystemStats, "error getting credential: %v", err)
+		server.SelfTest.HintCollectionAspect(state.CollectionAspectSystemStats, "Make sure the managed identity is assigned to the collector VM.")
+		logger.PrintError("Azure/System: Encountered error getting credential: %v\n", err)
+		return
+	}
+
+	var resourceID string
+
+	// Server info: Flexible Server
```
keiko713, line 35: Since we don't know the resource type (if Flexible server or Cosmos DB) at this point, we search in Flexible Servers first, then search in Cosmos DB.

Also, this is not getting the specific server info, but getting the all servers in that subscription. It's a bit unnecessary step (e.g. we could have asked all the needed information as config vars to avoid this), but we also want to use this pattern so that we could potentially do this at the collector level v.s. doing this over and over in each server. For example, I can make one list API request and obtain all servers info that the collector is monitoring.
(Lukas mentioned that it's the way we want to do in the future)

input/system/azure/system.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+	}
+
+	if resourceID == "" {
+		server.SelfTest.MarkCollectionAspectWarning(state.CollectionAspectSystemStats, "unable to find the database server info")
+		server.SelfTest.HintCollectionAspect(state.CollectionAspectSystemStats, "Make sure the Reader permission of the database is granted to the managed identity.")
+		return
+	}
+
+	// Server metrics
```
keiko713, line 136: Starting from here, it's getting the metrics. Available metrics are different between Cosmos DB and Flexible Server, but we can use the same logic here.

input/system/azure/system.go
```diff
@@ -0,0 +1,302 @@
...
+	client, err := azquery.NewMetricsClient(credential, nil)
+	if err != nil {
+		server.SelfTest.MarkCollectionAspectError(state.CollectionAspectSystemStats, "error making a metrics client: %v", err)
+		logger.PrintError("Azure/System: Failed to make a metrics client: %v\n", err)
+		return
+	}
+
+	// Query metrics data with 1 min interval, for the last 1 min (should return 1 value)
```
keiko713, line 144: I noticed that this API actually returns null values often for certain metrics. Read/Write throughputs with Flexible Server is a bad one I've seen, the following graph is from a test database but you can see that it's lacking values often (the dotted line is missing data):

image

I thought about changing the interval and timespan a bit to potentially make the data better quality (less null data), but I think expanding this is not going to help much. Also, the next available interval is 1m, 5m, 15m and no 10m, so I figured that I'll simply use 1m here.

input/system/azure/system.go
```diff
@@ -0,0 +1,302 @@
...
+		Aggregation: to.SliceOfPtrs(azquery.AggregationTypeAverage),
+		Interval:    to.Ptr("PT1M"),
+		Timespan:    to.Ptr(azquery.TimeInterval("PT1M")),
+	}
+
+	metricsRes, err := client.QueryResource(ctx, resourceID, option)
+	if err != nil {
+		server.SelfTest.MarkCollectionAspectError(state.CollectionAspectSystemStats, "error getting server metrics: %v", err)
+		logger.PrintError("Azure/System: Encountered error getting server metrics: %v\n", err)
+		server.SelfTest.HintCollectionAspect(state.CollectionAspectSystemStats, "Make sure the Monitoring Reader permission of the database is granted to the managed identity.")
```
keiko713, line 161: I actually saw that the only Reader role is enough to query this API too. Likely the Reader role is a parent role of the Monitoring Reader, so maybe this error message is incorrect and I should be simply saying "Reader permission" 🤔

input/system/azure/system.go
```diff
@@ -0,0 +1,302 @@
...
+								UserPercent: *metricValue.Average,
+							},
+						}
+					}
+				}
+			}
+		} else if *metric.Name.Value == "memory_percent" {
+			// Currently, we can only retrieve memory percent
+			// Since total memory size is not listed in the server info,
+			// we are unable to pass this value to pganalyze at the moment
```
keiko713, line 186: While the `memory_percent` metrics is available for both Flexible Server and Cosmos DB, it's super hard to get the memory size information of the server. Since pganalyze doesn't accept the memory percentage at the moment, I simply skipped recording this for now.

In the future, we could try obtaining the memory information of the server somehow. One option is to have our own SKU and memory size mapping data here (the list of sku can be obtained via `az postgres flexible-server list-skus --location eastus -o table` command), but looks like there is no easy place to find that information with Cosmos DB (the same CLI command is not available for Cosmos DB), so I decided to skip it.

input/system/azure/system.go
```diff
@@ -0,0 +1,302 @@
...
+							system.DiskPartitions = make(state.DiskPartitionMap)
+							system.DiskPartitions["/"] = state.DiskPartition{
+								DiskName:   "default",
+								UsedBytes:  uint64(*metricValue.Average),
+								TotalBytes: totalGB * 1024 * 1024 * 1024,
+							}
+						} else if system.Info.Azure.CoordinatorStorageMB != 0 {
+							// Cosmos DB
+							// TODO: check if we need to sum up the node storage MB too
+							// (e.g. system.Info.Azure.NodeStorageMB * system.Info.Azure.NodeCount)
```
keiko713, line 269: @lfittl Do you know the answer for this, by any chance? I haven't tried it since making a node with Cosmos DB is really expensive 🙈

input/system/azure/system.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+		for _, v := range page.Value {
+			if v.ID != nil {
+				resourceIDParts := strings.Split(*v.ID, "/")
+				resourceGroup := resourceIDParts[4]
+				resourceType := resourceIDParts[7]
+				resourceName := resourceIDParts[8]
+
+				if config.AzureDbServerName == resourceName {
+					system.Info.Azure = &state.SystemInfoAzure{
```
keiko713, line 58: FYI: state/system.go file is a good place to check and discuss if we're collecting good enough info here

util/null.go
```diff
@@ -37,9 +37,16 @@ func IntPtrToInt(ptr *int64) int64 {
 	return *ptr
 }
 
-func IntPtrToString(ptr *int64) int64 {
+func Int32PtrToInt(ptr *int32) int32 {
```
keiko713, line 40: This function wasn't used, and also not doing what it's advertising, so I reused and converted the one we actually use

input/system/azure/system.go
```diff
@@ -0,0 +1,302 @@
...
+	pager := clientFactory.NewServersClient().NewListPager(nil)
+	for pager.More() {
+		page, err := pager.NextPage(ctx)
+		if err != nil {
+			logger.PrintError("Azure/System: Failed to advance page of Flexible Server list: %v\n", err)
+			break
+		}
+		for _, v := range page.Value {
+			if v.ID != nil {
+				resourceIDParts := strings.Split(*v.ID, "/")
```
msakrejda, line 52: Might be nice to give an example of a resource id here, or link to Azure docs about the format.

input/system/azure/system.go
```diff
@@ -0,0 +1,302 @@
...
+							NodeServerEdition:        util.StringPtrToString(v.Properties.NodeServerEdition),
+							NodeCount:                util.Int32PtrToInt(v.Properties.NodeCount),
+						}
+						resourceID = *v.ID
+						break
+					}
+				}
+			}
+		}
+	}
```
msakrejda, line 128: Hmm. There's no way to get info about these directly by resource name? We have to iterate? It's probably not a big deal in most situations, but if someone is monitoring 100 servers in Azure, we'll make 100 (at least, subject to pagination) API calls every time we're collecting system stats, no? Due to our snapshot scheduling, these are also going to be happening at basically the same time, so if Azure has burst-unfriendly API limits, that could cause problems.

input/system/azure/system.go
msakrejda: Hmm. Might be worth opening a ticket with Azure support, just to understand what's going on?

input/system/azure/system.go
```diff
@@ -0,0 +1,302 @@
...
+	client, err := azquery.NewMetricsClient(credential, nil)
+	if err != nil {
+		server.SelfTest.MarkCollectionAspectError(state.CollectionAspectSystemStats, "error making a metrics client: %v", err)
+		logger.PrintError("Azure/System: Failed to make a metrics client: %v\n", err)
+		return
+	}
+
+	// Query metrics data with 1 min interval, for the last 1 min (should return 1 value)
+	metricNames := "cpu_percent,memory_percent,network_bytes_egress,network_bytes_ingress,storage_used"
+	if strings.ToLower(system.Info.Azure.ResourceType) == "flexibleservers" {
```
msakrejda, line 146: It's "flexibleservers", not "flexibleserver"? Just checking.

input/system/azure/system.go
msakrejda: :man_shrugging:

msakrejda (commented):
Some questions, and did not actually try this out, but looks good.

input/system/azure/system.go
keiko713: I think we don't have a support line in Azure. Given that it's like this in Azure console too (above screenshot is from Azure console), I think they simply don't have the data, so I'd say it's okay to leave it as is.

input/system/azure/system.go
keiko713: It is. It was "servers" with a single server.
https://learn.microsoft.com/en-us/azure/templates/microsoft.dbforpostgresql/allversions
And it is "servergroupsv2" with a Cosmos DB.
https://learn.microsoft.com/en-us/azure/templates/microsoft.dbforpostgresql/hsc-allversions

input/system/azure/system.go
keiko713: > if someone is monitoring 100 servers in Azure, we'll make 100 (at least, subject to pagination) API calls every time we're collecting system stats, no?

Do you mean if someone has 100 servers in Azure and even if they're monitoring one server in pganalyze, it'll make 100 API calls? Or, do you mean that if someone has 100 servers in Azure and they monitor 100 servers also in pganalyze, it'll make 100 API calls?

For the former, no, it won't, this part of API call should be one per the server that the collector is monitoring. Even with the pager, it'll get all 100 servers info with one API request (fun fact, we can't even specify the pager detail anyways, so we can only get all of them at once). The latter, this is what I covered in https://github.com/pganalyze/collector/pull/591/files#r1735845140. We want to reduce the number of API calls for this indeed, and it would be nice to actually do such thing.
Summarizing, the number of API calls that we will make per server will be up to 3:

1. list all Flexible servers
2. list all Cosmos DB for Postgres servers (only if the server can't be found in 1)
3. get metrics of the server

There is another problem of getting all servers info with one API request, like this: https://github.com/Azure/azure-sdk-for-go/issues/20470
If they indeed have 100 servers, while it's smaller than 30k+ resources like mentioned above, it is still not a small amount of data, so it could cause some memory spike (and compute power to go over that).

seanlinsley: Soon we'll start tracking system stats every 1 minute, so it's worth thinking about how to reduce API requests. One option might be to cache the API response in a separate data structure and only refresh it every 10 minutes.

keiko713: Agreed, when we increase a system stats collection frequency, we will need to rethink this, and also for other systems too (I believe we have the same problem for all other systems too).
For reducing step 1 and 2 requests, as Sean mentioned, we could only do this once 10 mins and do only 3 every 1 min as a starter. Doing 1 and 2 at the collector level as mentioned in the other comment is also good when the collector is monitoring many servers (likely 3 can't be done at the collector level though).

input/system/azure/system.go
keiko713: Just confirmed that "Monitoring Reader" is actually the required permission here, and "Reader" is too wide. I'll update and just mention "Monitoring Reader" here.

input/system/azure/system.go
lfittl: My assumption is no, i.e. we don't need to consider the worker node storage as part of the coordinator's storage (they can be considered separate servers)

keiko713:
Redeployed this version of collector to test instances, looking good including tags:
image

This is ready for re-review!
cc/ @lfittl

input/system/azure/system.go
```diff
@@ -0,0 +1,320 @@
...
+	"github.com/pganalyze/collector/util"
+)
+
+// GetSystemState - Gets system information about an Azure instance
+func GetSystemState(ctx context.Context, server *state.Server, logger *util.Logger) (system state.SystemState) {
+	config := server.Config
+	system.Info.Type = state.AzureDatabaseSystem
+	if config.AzureSubscriptionID == "" {
+		server.SelfTest.MarkCollectionAspectWarning(state.CollectionAspectSystemStats, "unable to collect system stats")
+		server.SelfTest.HintCollectionAspect(state.CollectionAspectSystemStats, "Config value azure_subscription_id is required to collect system stats.")
```
lfittl, line 21: Not sure if worth it, but we could reference the environment variable style here as well (i.e. `azure_subscription_id / AZURE_SUBSCRIPTION_ID` or something)

input/system/azure/system.go
```diff
@@ -0,0 +1,320 @@
...
+						break
+					}
+				}
+			}
+		}
+	}
+
+	if resourceID == "" {
+		server.SelfTest.MarkCollectionAspectWarning(state.CollectionAspectSystemStats, "unable to find the database server info")
+		server.SelfTest.HintCollectionAspect(state.CollectionAspectSystemStats, "Make sure the Monitoring Reader permission of the database is granted to the managed identity.")
```
lfittl, line 152: Does this need to be moved to the earlier parts where we actually make the list API calls? (since we return early on error) -- or do we just not get any results if the permissions are insufficient?

input/system/azure/system.go
```diff
@@ -0,0 +1,320 @@
...
+			// Since total memory size is not listed in the server info,
+			// we are unable to pass this value to pganalyze at the moment
+		} else if *metric.Name.Value == "txlogs_storage_used" {
+			for _, timeSeriesElement := range metric.TimeSeries {
+				for _, metricValue := range timeSeriesElement.Data {
+					if metricValue.Average != nil {
+						system.XlogUsedBytes = uint64(*metricValue.Average)
+					}
+				}
+			}
```
lfittl, line 214: We could refactor the duplication here a bit, similar to how we have this for RDS https://github.com/pganalyze/collector/blob/main/util/awsutil/rds.go#L150

protobuf/full_snapshot.proto
```diff
@@ -319,6 +319,7 @@ message CollectorConfig {
   string azure_ad_tenant_id = 57;
   string azure_ad_client_id = 58;
   bool azure_has_ad_certificate = 59;
+  string azure_resource_id = 60;
```
lfittl, line 322: Shouldn't this be `azure_subscription_id`?

input/collector.go
```diff
@@ -102,6 +102,7 @@ func getCollectorConfig(c config.ServerConfig) state.CollectorConfig {
 		AzureAdTenantId:            c.AzureADTenantID,
 		AzureAdClientId:            c.AzureADClientID,
 		AzureHasAdCertificate:      c.AzureADCertificatePath != "",
+		AzureSubscriptionID:        c.AzureSubscriptionID,
```
lfittl, line 105: This also needs to be emitted on output

lfittl (approved):

input/system/azure/system.go
keiko713: Just to close the loop: I added the comment above, but this can be reached when the managed identity is assigned to some db but not this db, so I think the wording is still accurate. When it's not assigned to any db resources, it'll raise an error when we do `page, err := pager.NextPage(ctx)` earlier in line 46.

I tweaked that part of code, to make sure that it'll early return when it receives errors (and show the same hint). It'll look like the following. It's not super good looking sadly (I think %v is just showing too much), but I think it does show good info for debugging so I'd say it's a good start.

image

input/system/azure/system.go
```diff
@@ -0,0 +1,282 @@
...
+		case "txlogs_storage_used":
+			system.XlogUsedBytes = uint64(*metricValue.Average)
+		case "network_bytes_egress":
+			// value is total of 1min
+			diffedNetworkStats.TransmitThroughputBytesPerSecond = uint64(*metricValue.Average / 60)
+		case "network_bytes_ingress":
+			// value is total of 1min
+			diffedNetworkStats.ReceiveThroughputBytesPerSecond = uint64(*metricValue.Average / 60)
+		case "read_iops":
+			diffedDiskStats.ReadOperationsPerSecond = *metricValue.Average
```
seanlinsley, line 222: Sentry reported a nil pointer dereference here https://sentry.io/issues/5910271246

keiko713: Yeah it's already fixed actually, but thanks for the ping!

input/system/azure/system.go
```diff
@@ -0,0 +1,282 @@
...
+
+// getFirstMetricValue gets the first data from the time series metric and returns the value
+func getFirstMetricValue(metric *azquery.Metric) (metricValue *azquery.MetricValue) {
+	for _, timeSeriesElement := range metric.TimeSeries {
+		for _, mValue := range timeSeriesElement.Data {
+			metricValue = mValue
+		}
+	}
+	return
+}
```
msakrejda, line 282: Reviewing #645, this definition jumped out at me as a little unclear: this seems to look for the last Data element in the last TimeSeries, right? That doesn't seem to align with the comment.

keiko713: Very good point, if I wanna make the comment right, I should be adding return after line 278 - otherwise, I should be updating the comment to say "the last data". I think the comment and function name is pretty clear that Keiko in September wanted to pick the first value, so I'll update the logic to do so.

PR #595, opened 2024-09-08
Add DB_URL_FILE and DB_PASSWORD_FILE

### LINKED ISSUE #540: Feature request: Add DB_PASSWORD_FILE or similar
I am looking to set up pganalyze on my Nix-based system. Unfortunately, due to the way I am handling secrets, I cannot directly provide the password in an environment variable (that would require storing the value in the nix store, which is readable by anyone with access to the system), so I was wondering if you might consider adding support for a DB_PASSWORD_FILE environment variable or similar. This would let me put the password file in a secure place (only readable by the pganalyze user) and then pass the file location to the program.

Thanks for your work on pganalyze, I'm really looking forward to trying it out on my system!

Munksgaard:
This allows passing sensitive DB passwords through files instead of environment variables. This makes collector work better with [systemd credentials](https://systemd.io/CREDENTIALS/) and NixOS flakes.

Fixes #540

### DIFF
--- a/config/read.go
+++ b/config/read.go
@@ -109,6 +109,15 @@ func getDefaultConfig() *ServerConfig {
 	if dbURL := os.Getenv("DB_URL"); dbURL != "" {
 		config.DbURL = dbURL
 	}
+	if dbURLFile := os.Getenv("DB_URL_FILE"); dbURLFile != "" {
+		dbURL, err := os.ReadFile(dbURLFile)
+
+		if err != nil {
+			panic(err)
+		}
+
+		config.DbURL = strings.TrimSpace(string(dbURL))
+	}
 	if dbName := os.Getenv("DB_NAME"); dbName != "" {
 		config.DbName = dbName
 	}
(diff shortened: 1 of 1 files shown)


Munksgaard:
I am unsure if I should add similar functionality for PGA_API_KEY. I'm guessing this is considered sensitive information as well, but the logic around loading PGA_API_KEY seems a bit more complicated, so I would probably need some guidance.

msakrejda (commented):
Nice, thanks for the contribution. Thoughts:
 - Should the `_FILE` versions silently override the explicitly specified strings like this?
 - Should we be using `TrimSpace`? Technically your password could start or end with a space
 - Should we support these from a file-based config, in addition to env-based config? 
 - There's currently no way to return an error from `getDefaultConfig`, but is `panic` the right error-handling strategy here?

Munksgaard:
Hi @msakrejda!

These are all good questions, which I don't think I'm necessarily the right person to answer, but I'll give it a shot. To preface, I'll say that I don't have much experience with go, so forgive my ignorance.

1. I _think_ that is how these things usually go, but I wouldn't mind adding a warning of some sort. How would you suggest I do that?
2. Great question. I would think that these values are usually trimmed, to remove any newlines at the end of the file, but you've got a great point that the password might contain spaces. I'm fine with removing the trimming if desired.
3. Sure, I don't see why not. How/where would I add that?
4. Great question. I'll leave that up to the maintainers to decide.

msakrejda:
Thanks. I'll check in with my colleagues and we'll follow up.

msakrejda:
Closing in favor of #629

PR #596, opened 2024-09-09
Reuse existing database connections

seanlinsley:
- Update `postgres.GetPostgresSetting` to accept a database connection as an argumentf
- Update `runner/full.go` to use a single database connection while collecting a full snapshot


lfittl:
As of 3/18/25 we still think this is worth considering, but this no longer applies after recent refactorings - need to rework the PR to make it work again.

PR #597, opened 2024-09-09
Include log text in snapshots

seanlinsley:
Log text is now included directly in snapshots instead of uploading the logs separately to S3.

In passing, filtered logs are now replaced with `[redacted]`. Previously an exact number of `X` characters would be used to keep the length of the log text from changing. This resulted in hard-to-read logs when the filtered content was very long.

Possible followup PRs:

- [ ] [remove `--dry-run-logs`](https://github.com/pganalyze/collector/pull/583#discussion_r1781870418)
- [ ] remove temp file usage


logs/analyze.go
```diff
@@ -2266,8 +2266,7 @@ func AnalyzeBackendLogLines(logLines []state.LogLine) (logLinesOut []state.LogLi
 		logLinesOut = append(logLinesOut, logLine)
 	}
 
-	// Ensure no other part of the system accidentally sends log line contents, as
-	// they should be considered opaque from here on
+	// Remove log line content. Note that ReplaceSecrets adds it back after secrets have been removed.
 	for idx := range logLinesOut {
 		logLinesOut[idx].Content = ""
```
seanlinsley, line 2271: A future PR might remove the tempfile being written since I think it's not needed now. At that point we wouldn't need to zero-out the log text here. Though that would also require refactoring the log filter code so the byte offsets are relative to that string, instead of being relative to the entire file.

logs/replace.go
```diff
@@ -1,74 +1,41 @@
...
+			offset := 0
 			for _, m := range logLine.SecretMarkers {
-				filter := false
 				for _, k := range filterLogSecret {
-					if m.Kind == k {
-						filter = true
+					if m.Kind == k && m.ByteStart > bytesChecked {
+						content = slices.Replace(content, m.ByteStart-offset, m.ByteEnd-offset, []byte(replacement)...)
+						bytesChecked = m.ByteEnd
+						offset += max(1, len(replacement)-m.ByteEnd-m.ByteStart)
```
msakrejda, line 34: I don't quite follow this. Should this be `len(replacement)-(m.ByteEnd-m.ByteStart)`? Why does the offset have a minimum of 1? Shouldn't a negative offset be possible if we're replacing very long content with a short marker?

seanlinsley: Looks like it actually needed to be `(m.ByteEnd - m.ByteStart) - len(replacement)`. I've updated the test to cover the case where long content is after a short marker.

msakrejda: Ah yes, right. That looks good. Thanks for updating the test.

msakrejda (approved):

PR #598, opened 2024-09-10
Upgrade to Go 1.23.1

seanlinsley:
- https://tip.golang.org/doc/go1.22
- https://tip.golang.org/doc/go1.23


lfittl:
I wonder if we need to check for impact of the for loop change in 1.22?

see https://tip.golang.org/doc/go1.22#language and https://tip.golang.org/wiki/LoopvarExperiment#my-test-fails-with-the-change-how-can-i-debug-it

keiko713:
Closing in favor of https://github.com/pganalyze/collector/pull/669

PR #601, opened 2024-10-02
Improve support for EDB Postgres Advanced Server (EPAS)

lfittl:
Previously the collector would emit system catalogs that are only existing on EPAS ("sys" schema and child schemas), causing unnecessary schema information to show in the pganalyze dashboard.

Here we fix this by introducing EPAS-specific filtering of system catalogs, and share the system catalog filter logic across all files that use it.

Further, EPAS allows function signatures to be re-used between functions and procedures (presumably this was allowed before upstream Postgres added CREATE PROCEDURE), like this:

```
BEGIN;
CREATE FUNCTION test() RETURNS void AS $$
BEGIN
    RAISE NOTICE 'test function';
END;
$$ LANGUAGE plpgsql;
CREATE PROCEDURE test() AS $$
BEGIN
    RAISE NOTICE 'test procedure';
END;
$$ LANGUAGE plpgsql;
COMMIT;
 
SELECT prokind, pronamespace, pg_catalog.pg_get_function_arguments(pp.oid), pg_catalog.pg_get_function_result(pp.oid) FROM pg_proc pp WHERE proname = 'test';

 prokind | pronamespace | pg_get_function_arguments | pg_get_function_result 
---------+--------------+---------------------------+------------------------
 f       |         2200 |                           | integer
 p       |         2200 |                           | 
(2 rows)
```

The same SQL causes an error on regular Postgres. Unfortunately this unexpected duplication causes a server-side error with pganalyze due to a unique constraint violation.

Whilst we could track this properly on the server side, since this is an edge case and there are no intents to support this in upstream Postgres (see note on https://www.postgresql.org/docs/current/sql-createprocedure.html), ignore duplicate functions/procedures that have the same signature for now, but only if we're on EPAS (to avoid overhead on regular Postgres).

### DIFF
--- a/input/postgres/relations.go
+++ b/input/postgres/relations.go
@@ -11,6 +11,9 @@ import (
 	"github.com/pganalyze/collector/state"
 )
 
+const relationSQLdefaultSystemCatalogFilter = "n.nspname NOT IN ('pg_catalog', 'pg_toast', 'information_schema')"
+const relationSQLepasSystemCatalogFilter = "n.nspname NOT IN ('pg_catalog', 'pg_toast', 'information_schema', 'sys') AND n.nspparent <> 'sys'::regnamespace"
+
 const relationsSQLOidField = "c.relhasoids AS relation_has_oids"
 const relationsSQLpg12OidField = "false AS relation_has_oids"
 
@@ -41,7 +44,7 @@ const relationsSQL string = `
 	WHERE c.relkind IN ('r','v','m','p')
 				AND c.relpersistence <> 't'
 				AND c.oid NOT IN (SELECT pd.objid FROM pg_catalog.pg_depend pd WHERE pd.deptype = 'e' AND pd.classid = 'pg_catalog.pg_class'::regclass)
-				AND n.nspname NOT IN ('pg_catalog','pg_toast','information_schema')
+				AND %s
 				AND ($1 = '' OR (n.nspname || '.' || c.relname) !~* $1)`
 
 const columnsSQL string = `
@@ -64,7 +67,7 @@ const columnsSQL string = `
  WHERE c.relkind IN ('r','v','m','p')
 			 AND c.relpersistence <> 't'
 			 AND c.oid NOT IN (SELECT pd.objid FROM pg_catalog.pg_depend pd WHERE pd.deptype = 'e' AND pd.classid = 'pg_catalog.pg_class'::regclass)
-			 AND n.nspname NOT IN ('pg_catalog','pg_toast','information_schema')
+			 AND %s
 			 AND a.attnum > 0
 			 AND NOT a.attisdropped
 			 AND c.oid NOT IN (SELECT relid FROM locked_relids)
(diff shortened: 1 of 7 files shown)


state/postgres_version.go
```diff
@@ -22,6 +22,7 @@ type PostgresVersion struct {
 	Numeric int    `json:"numeric"` // e.g. 90501
 
 	// For collector use only, to avoid calling functions that don't work
-	IsAwsAurora bool
-	IsCitus     bool
+	IsAwsAurora bool // Amazon Aurora
+	IsCitus     bool // Citus extension (e.g. with Azure CosmosDB for PostgreSQL)
+	IsEPAS      bool // EnterpriseDB Advanced Server
```
lfittl, line 27: FWIW, the reason this is not just "IsEDB" is because they actually have a few different options (e.g. "Extended Server"), but only Advanced Server is the one that has the additional catalog things (and lots of Oracle compatibility features), in my understanding.

The abbreviation "EPAS" is also used in their official docs, e.g. here: https://www.enterprisedb.com/docs/epas/latest/

input/postgres/relations.go
```diff
@@ -11,6 +11,9 @@ import (
 	"github.com/pganalyze/collector/state"
 )
 
+const relationSQLdefaultSystemCatalogFilter = "n.nspname NOT IN ('pg_catalog', 'pg_toast', 'information_schema')"
+const relationSQLepasSystemCatalogFilter = "n.nspname NOT IN ('pg_catalog', 'pg_toast', 'information_schema', 'sys') AND n.nspparent <> 'sys'::regnamespace"
```
lfittl, line 15: Note that "nspparent" is not a feature in regular Postgres (in EPAS there are a handful of internal schemas that all are children of "sys").

input/postgres/relations.go
```diff
@@ -11,6 +11,9 @@ import (
 	"github.com/pganalyze/collector/state"
 )
 
+const relationSQLdefaultSystemCatalogFilter = "n.nspname NOT IN ('pg_catalog', 'pg_toast', 'information_schema')"
```
lfittl, line 14: Note that "pg_toast" wasn't in all the previous queries (some only excluded pg_catalog and information_schema), but I don't see harm in excluding this always.

The one exception I can think of today where we actively look at TOAST data is for vacuums, and this change shouldn't have any effect on that.

input/postgres/relations.go
keiko713: not a strong opinion but it can be all capital letters as it's an acronym:

```suggestion
const relationSQLEPASSystemCatalogFilter = "n.nspname NOT IN ('pg_catalog', 'pg_toast', 'information_schema', 'sys') AND n.nspparent <> 'sys'::regnamespace"
```

input/postgres/types.go
```diff
@@ -33,13 +34,18 @@ SELECT t.oid,
     AND (t.typrelid = 0 OR (SELECT c.relkind = 'c' FROM pg_catalog.pg_class c WHERE c.oid = t.typrelid))
     AND NOT EXISTS (SELECT 1 FROM pg_catalog.pg_type el WHERE el.oid = t.typelem AND el.typarray = t.oid)
 	AND t.oid NOT IN (SELECT pd.objid FROM pg_catalog.pg_depend pd WHERE pd.deptype = 'e' AND pd.classid = 'pg_catalog.pg_type'::regclass)
-    AND n.nspname <> 'pg_catalog'
-    AND n.nspname <> 'information_schema'
-    AND n.nspname !~ '^pg_toast'
```
keiko713, line 38: note from Lukas: there can be `pg_toast_something` but we're only doing this here, and there is nothing special about type specific, so he decided to drop this and consolidate

keiko713 (approved):
Small comments, but overall looking good (also confirmed that Lukas tested this)

PR #602, opened 2024-10-02
Fix --analyze-logfile flag

msakrejda:
In #494, I missed updating the --analyze-logfile flag. The flag is
intended to analyze a logfile "offline", which does not work with the
new mechanism, since that relies on being able to read the server's
log_timezone and log_line_prefix.

This adds additional flags to pass that information to the
--analyze-logfile path. The timezone defaults to UTC; the prefix is
required.

### DIFF
--- a/main.go
+++ b/main.go
@@ -316,7 +316,7 @@ func main() {
 	flag.BoolVar(&dryRunLogs, "dry-run-logs", false, "Print JSON data for log snapshot (without actually sending) and exit afterwards")
 	flag.StringVar(&analyzeLogfile, "analyze-logfile", "", "Analyzes the content of the given log file and returns debug output about it")
 	flag.StringVar(&analyzeLogfilePrefix, "analyze-logfile-prefix", "", "The log_line_prefix to use with --analyze-logfile")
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -313,6 +315,8 @@ func main() {
 	flag.BoolVar(&dryRun, "dry-run", false, "Print JSON data that would get sent to web service (without actually sending) and exit afterwards")
 	flag.BoolVar(&dryRunLogs, "dry-run-logs", false, "Print JSON data for log snapshot (without actually sending) and exit afterwards")
 	flag.StringVar(&analyzeLogfile, "analyze-logfile", "", "Analyzes the content of the given log file and returns debug output about it")
+	flag.StringVar(&analyzeLogfilePrefix, "analyze-logfile-prefix", "", "The log_line_prefix to use with --analyze-logfile")
+	flag.StringVar(&analyzeLogfileTz, "analyze-logfile-tz", "", "The timezone to use with --analyze-logfile (default UTC)")
```
keiko713, line 319: Does `time.LoadLocation("")` become UTC?

msakrejda: Yeah, exactly.

keiko713 (approved):

PR #606, opened 2024-10-05
Handle the database / user being empty

JamesMcNee:
Raised as part of open case (5773) with support regarding a memory leak. Will link PR in ticket. 

It is possible for the username and the database to be empty for some log messages. For example with CloudSQL the vacuum analysis logs are output as

> 2024-10-04 06:38:28.316 UTC [2569017]: [1-1] db=,user= LOG:  automatic vacuum of table \\\"database_name.pg_catalog.pg_database\\\": etc

This previously caused a bug where these lines were being treated as incomplete, resulting in a memory leak as "the rest" of the line was awaited (to stitch together logs that appear over multiple lines).

The ultimate result of this is that vacuum logs on CloudSQL never make it to PgAnalyze and therefore are not available in log insights or used as part of the vacuum advisor.

This commit fixes this by changing the regex matching for username and database to match anything that is not a single space, rather than just anything.

### DIFF
--- a/logs/parse.go
+++ b/logs/parse.go
@@ -69,7 +69,7 @@ var EscapeMatchers = map[rune]PrefixEscape{
 	},
 	// User name
 	'u': {
-		Regexp: `.*?`,
+		Regexp: `.+?`,
 		ApplyValue: func(value string, logLine *state.LogLine, parser *LogParser) {
 			if value == "[unknown]" {
 				return
@@ -80,7 +80,7 @@ var EscapeMatchers = map[rune]PrefixEscape{
 	},
 	// Database name
 	'd': {
-		Regexp: `.*?`,
+		Regexp: `.+?`,
 		ApplyValue: func(value string, logLine *state.LogLine, parser *LogParser) {
 			if value == "[unknown]" {
 				return
--- a/logs/parse.go
+++ b/logs/parse.go
@@ -69,7 +69,7 @@ var EscapeMatchers = map[rune]PrefixEscape{
 	},
 	// User name
 	'u': {
-		Regexp: `.+?`,
+		Regexp: `[^ ].+?`,
 		ApplyValue: func(value string, logLine *state.LogLine, parser *LogParser) {
 			if value == "[unknown]" {
 				return
@@ -80,7 +80,7 @@ var EscapeMatchers = map[rune]PrefixEscape{
 	},
 	// Database name
 	'd': {
-		Regexp: `.+?`,
+		Regexp: `[^ ].+?`,
 		ApplyValue: func(value string, logLine *state.LogLine, parser *LogParser) {
 			if value == "[unknown]" {
 				return
--- a/logs/parse.go
+++ b/logs/parse.go
@@ -69,7 +69,7 @@ var EscapeMatchers = map[rune]PrefixEscape{
 	},
 	// User name
 	'u': {
-		Regexp: `[^ ].+?`,
+		Regexp: `(?:[^ ]|[^ ]{2,})?`,
 		ApplyValue: func(value string, logLine *state.LogLine, parser *LogParser) {
 			if value == "[unknown]" {
 				return
@@ -80,7 +80,7 @@ var EscapeMatchers = map[rune]PrefixEscape{
 	},
 	// Database name
 	'd': {
-		Regexp: `[^ ].+?`,
+		Regexp: `(?:[^ ]|[^ ]{2,})?`,
 		ApplyValue: func(value string, logLine *state.LogLine, parser *LogParser) {
 			if value == "[unknown]" {
 				return
--- a/logs/parse.go
+++ b/logs/parse.go
@@ -69,7 +69,7 @@ var EscapeMatchers = map[rune]PrefixEscape{
 	},
 	// User name
 	'u': {
-		Regexp: `(?:[^ ]|[^ ]{2,})?`,
+		Regexp: `.*?`,
 		ApplyValue: func(value string, logLine *state.LogLine, parser *LogParser) {
 			if value == "[unknown]" {
 				return
@@ -80,7 +80,7 @@ var EscapeMatchers = map[rune]PrefixEscape{
 	},
 	// Database name
 	'd': {
-		Regexp: `(?:[^ ]|[^ ]{2,})?`,
+		Regexp: `.*?`,
 		ApplyValue: func(value string, logLine *state.LogLine, parser *LogParser) {
 			if value == "[unknown]" {
 				return
--- a/logs/parse.go
+++ b/logs/parse.go
@@ -195,7 +195,7 @@ type LogParser struct {
 
 func NewLogParser(prefix string, tz *time.Location, isSyslog bool) *LogParser {
 	prefixRegexp, prefixElements := parsePrefix(prefix)
-	lineRegexp := regexp.MustCompile("(?ms)^" + prefixRegexp + `(\w+):\s+(.*\n?)$`)
+	lineRegexp := regexp.MustCompile("(?ms)^" + prefixRegexp + `(DEBUG|INFO|NOTICE|WARNING|ERROR|LOG|FATAL|PANIC|DETAIL|HINT|CONTEXT|STATEMENT|QUERY):\s+(.*\n?)$`)
 	lineRegexpWithoutLogLevel := regexp.MustCompile("(?ms)^" + prefixRegexp + `(.*\n?)$`)
 	return &LogParser{
 		prefix:   prefix,
--- a/logs/parse_test.go
+++ b/logs/parse_test.go
@@ -458,6 +458,32 @@ var parseTests = []parseTestpair{
 		},
 		true,
 	},
+	// Custom 11 format
+	{
+		logs.LogPrefixCustom11,
+		"pid=8284,user=[unknown],db=[unknown],app=[unknown],client=[local] LOG: connection received: host=[local]",
+		nil,
+		state.LogLine{
+			BackendPid: 8284,
+			LogLevel:   pganalyze_collector.LogLineInformation_LOG,
+			Content:    "connection received: host=[local]",
+		},
+		true,
+	},
+	{
+		// Updating this to reflect that we can now capture the unusual application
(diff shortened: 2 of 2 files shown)


seanlinsley:
Thanks for the contribution! I think it generally makes sense, but will defer to @msakrejda or @lfittl on whether `[^ ]` is the right pattern to match.

JamesMcNee:
Hi @seanlinsley, thanks for picking up. I noticed some tests failed with the regex in the original commit (apologies I only ran the parser tests originally 😅). I have updated the regex now and it seems to do the trick.

msakrejda (commented):
Thanks for tracking down the issue, @JamesMcNee, and for the patch!

I think these may be empty because they're running in a non-session context, so there may not be a database or user associated with the log line. Normally the [%q](https://www.postgresql.org/docs/17/runtime-config-logging.html#GUC-LOG-LINE-PREFIX) format is used to avoid including those fields entirely for non-session processes, but we should definitely handle cases where that's not done, and the fields are logged as empty.

I'm leaving some comments about the regexp, but broadly this looks great and we should have it in the next release. If you don't have time to make the changes I requested, I can wrap this up and get it merged, but I want to make sure the fix still works for you with my changes.

logs/parse.go
```diff
@@ -69,7 +69,7 @@ var EscapeMatchers = map[rune]PrefixEscape{
 	},
 	// User name
 	'u': {
-		Regexp: `.+?`,
+		Regexp: `(?:[^ ]|[^ ]{2,})?`,
```
msakrejda, line 72: This comment applies both here and to the "Database name" change below:

```suggestion
		Regexp: `.*?`,
```

So, the original regexp here is a non-greedy match of one or more characters, because identifiers like user cannot be empty but (when quoted) can contain any character (up to the max identifier length, 63 bytes without recompiling Postgres). (The current bug is that while usernames cannot be empty, they can be logged as empty when logging from a non-session process, because there's no associated user.)

Your regexp here is an optional match of either a non-space, or two or more non-spaces. That will capture most valid usernames, and since it's an optional match, it will also match an empty string (fixing the bug). But it will also reject user names including spaces, or consisting entirely of spaces (neither of these is probably a good idea, but we should support it if possible).

I'm proposing that instead, we change the regexp to capture zero or more characters, instead of one or more, and still keep it as a non-greedy match. This passes the tests, including the test you've added, and I think it should also allow valid user names that have spaces.

Can you confirm if this works for you?

logs/parse_test.go
```diff
@@ -442,6 +442,22 @@ var parseTests = []parseTestpair{
 		},
 		true,
 	},
+	{
+		logs.LogPrefixCustom10,
+		// Database and user are empty
+		"2020-09-04 16:03:11.375 UTC [417880]: [1-1] db=,user= LOG:  pganalyze-collector-identify: myserver",
```
msakrejda, line 448: This seems to be copied from the example above, but I think log lines like that normally happen in a session context, where a user and database are present, so they would not be empty. Could we use a more realistic log line here? I think copying the line from the test for example 14 (`"2021-11-17 19:06:53.897 UTC [34][autovacuum worker][3/5][22996] LOG:  automatic analyze of table \"mydb.pg_catalog.pg_class\" system usage: CPU: user: 0.00 s, system: 0.00 s, elapsed: 0.01 s",`) should work here (you'll of course need to change the prefix to match).

msakrejda:
Hmm, actually, if I revert your changes to `parse.go` and only add your test, the test passes. That's because the whole group is treated as optional in those cases (with the `Optional` filed in the `PrefixEscape` struct). Can you add a test case that fails on `main`?

JamesMcNee:
Hi @msakrejda Thanks for looking through this, and apologies for the naff test that did not actually test the scenario 😂. I have pushed an updated test that does show the issue and also applied your fix, which is not only simpler but does fix the issue!

logs/parse_test.go
JamesMcNee: Have updated the the test to both actually capture the issue, but also include a realistic log line

logs/parse_test.go
```diff
@@ -442,29 +442,19 @@ var parseTests = []parseTestpair{
 		},
 		true,
 	},
-	// Custom 11 format
 	{
-		logs.LogPrefixCustom11,
```
seanlinsley, line 447: It looks like the existing test here was removed. I assume that was accidental?

JamesMcNee: Yes, sorry, added back 😅

msakrejda:
Thanks for updating the test. I'm still trying to understand the actual failure (since, as I noted earlier, the `Optional` handling should catch the issue). The problem is not that we don't get a match: the problem is that we get a match with a bogus LogLevel because of insufficient checks. Here's the failure on `main` from your added test:

```
    parse_test.go:679: For "2024-10-04 06:38:28.808 UTC [2569017]: [1-1] db=,user= LOG: automatic vacuum of table "some_database.some_schema.some_table": index scans: 0
        pages: 0 removed, 14 remain, 0 skipped due to pins, 0 skipped frozen
        tuples: 0 removed, 107 remain, 94 are dead but not yet removable, oldest xmin: 17243830
        index scan not needed: 0 pages from table (0.00% of total) had 0 dead item identifiers removed
        I/O timings: read: 0.000 ms, write: 0.000 ms
        avg read rate: 0.000 MB/s, avg write rate: 0.000 MB/s
        buffer usage: 52 hits, 0 misses, 0 dirtied
        WAL usage: 0 records, 0 full page images, 0 bytes
        system usage: CPU: user: 0.00 s, system: 0.00 s, elapsed: 0.00 s": log line diff: (-got +want)
         {
          OccurredAt: 2024-10-04 06:38:28.808 +0000 UTC,
        - Username: " LOG: automatic vacuum of table \"some_database.some_schema.some_table\": index",
        + LogLevel: 6,
          BackendPid: 2569017,
          LogLineNumber: 1,
        - Content: "0\npages: 0 removed, 14 remain, 0 skipped due to pins, 0 skipped frozen\ntuples: 0 removed, 107 remain, 94 are dead but not yet removable, oldest xmin: 17243830\nindex scan not needed: 0 pages from table (0.00% of total) had 0 dead item identifiers removed\nI/O timings: read: 0.000 ms, write: 0.000 ms\navg read rate: 0.000 MB/s, avg write rate: 0.000 MB/s\nbuffer usage: 52 hits, 0 misses, 0 dirtied\nWAL usage: 0 records, 0 full page images, 0 bytes\nsystem usage: CPU: user: 0.00 s, system: 0.00 s, elapsed: 0.00 s",
        + Content: "automatic vacuum of table \"some_database.some_schema.some_table\": index scans: 0\npages: 0 removed, 14 remain, 0 skipped due to pins, 0 skipped frozen\ntuples: 0 removed, 107 remain, 94 are dead but not yet removable, oldest xmin: 17243830\nindex scan not needed: 0 pages from table (0.00% of total) had 0 dead item identifiers removed\nI/O timings: read: 0.000 ms, write: 0.000 ms\navg read rate: 0.000 MB/s, avg write rate: 0.000 MB/s\nbuffer usage: 52 hits, 0 misses, 0 dirtied\nWAL usage: 0 records, 0 full page images, 0 bytes\nsystem usage: CPU: user: 0.00 s, system: 0.00 s, elapsed: 0.00 s",
         }
```
(The `+` lines are expected output, the `-` ones are the output fields we are actually getting.)

So yes, we're missing LogLevel, but how? Our LogLevel field is [a required part of the regexp](https://github.com/pganalyze/collector/blob/main/logs/parse.go#L198) (it's the `(\w+):`—the log level immediately follows the prefix). If you look at the Username, field in the diff it's actually capturing the wrong thing: it's capturing the log level plus some of the actual log output. But if the log level is required, how does that still match the overall regexp? I instrumented the code to just print the captured log level, before mapping it to the protobuf value, and it turns out the log level is captured as "scans". That matches the `(\w+):` expression we're using, because the autovacuum outputs includes a colon after the word "scans". When we [map that to a protobuf log level](https://github.com/pganalyze/collector/blob/main/logs/parse.go#L432), we don't do any error checking, so we end up with 0.

I think the fix I proposed (`.*?`) would work, but it could break some other cases with weird user names. Since the set of log levels is actually [quite small](https://github.com/pganalyze/collector/blob/main/protobuf/compact_log_snapshot.proto#L59-L78), I think we should instead just have a tighter regexp fragment for LogLevel:

```diff
diff --git a/logs/parse.go b/logs/parse.go
index 4908352d..95efe2ec 100644
--- a/logs/parse.go
+++ b/logs/parse.go
@@ -195,7 +195,7 @@ type LogParser struct {
 
 func NewLogParser(prefix string, tz *time.Location, isSyslog bool) *LogParser {
        prefixRegexp, prefixElements := parsePrefix(prefix)
-       lineRegexp := regexp.MustCompile("(?ms)^" + prefixRegexp + `(\w+):\s+(.*\n?)$`)
+       lineRegexp := regexp.MustCompile("(?ms)^" + prefixRegexp + `(DEBUG|INFO|NOTICE|WARNING|ERROR|LOG|FATAL|PANIC|DETAIL|HINT|CONTEXT|STATEMENT|QUERY):\s+(.*\n?)$`)
        lineRegexpWithoutLogLevel := regexp.MustCompile("(?ms)^" + prefixRegexp + `(.*\n?)$`)
        return &LogParser{
                prefix:   prefix,
```

This passes the test you added, and I think it's less likely to break anything (it should also perform a hair better, since the set of valid log levels to match is smaller.

Can you revert the change to the `EscapeMatchers` struct and check if having this change instead also fixes the issue for you?

JamesMcNee:
Hi @msakrejda, yep, just tested this alternate fix (reverting the changes to the `EscapeMatchers` and it passes both the unit test as you mention and also a manual test I have set up against an actual instance 👍.

How would you like to proceed here? I will push up commits undoing the changes and adding in your fix, but feel free to close the either push commits to the branch, or close the PR in favour of your own if preferred! It's late here now (I am in the UK), so will likely be tomorrow before I can reply again.

msakrejda:
Great, thanks for verifying. The updated patch looks good. I'll run CI and merge if that's green.

msakrejda (approved):

PR #607, opened 2024-10-07
Remove temp file usage for logs

seanlinsley:
Followup to #597, this removes temp file usage in log code paths, instead keeping the log content in memory.


input/system/rds/logs.go
```diff
@@ -182,71 +152,9 @@ func downloadRdsLogFilePortion(rdsSvc *rds.RDS, tmpFile *os.File, logger *util.L
...
-
-// Analyze and submit at most the trailing 10 megabytes of the retrieved RDS log file portions
-//
-// This avoids an OOM in two edge cases:
-// 1) When starting the collector, as we always load the last 10,000 lines (which may be very long)
-// 2) When extremely large values are output in a single log event (e.g. query parameters in a DETAIL line)
-//
-// We intentionally throw away data here (and warn the user about it), since the alternative
-// is often a collector crash (due to OOM), which would be less desirable.
-const maxLogParsingSize = 10 * 1024 * 1024
```
seanlinsley, line 207: I suspect this is no longer needed as of #384 and #597, though this will require production testing to confirm.

logs/analyze_test.go
```diff
@@ -4328,6 +4328,9 @@ var tests = []testpair{
 func TestAnalyzeLogLines(t *testing.T) {
 	for _, pair := range tests {
 		l, s := logs.AnalyzeLogLines(pair.logLinesIn)
+		for idx := range l {
+			l[idx].Content = ""
```
seanlinsley, line 4332: Multiple tests now do this to avoid needing to copy-paste the log content into the expected output struct. I could do that if desired, it would just be tedious.

logs/debug.go
```diff
@@ -35,15 +35,13 @@ func PrintDebugInfo(logFileContents string, logLines []state.LogLine, samples []
 	if len(unclassifiedLogLines) > 0 {
 		fmt.Printf("\nUnclassified log lines:\n")
 		for _, logLine := range unclassifiedLogLines {
-			fmt.Printf("%s\n", logFileContents[logLine.ByteStart:logLine.ByteEnd])
-			fmt.Printf("  Level: %s\n", logLine.LogLevel)
-			fmt.Printf("  Content: %#v\n", logFileContents[logLine.ByteContentStart:logLine.ByteEnd])
+			fmt.Printf("  %s: %s", logLine.LogLevel, logLine.Content)
```
seanlinsley, line 38: I haven't tried running the debug command yet. Someone might have a preference on the output format.

One downside to this PR is we can no longer debug the entire log content including the unparsed log prefix.

runner/logs.go
```diff
@@ -332,24 +329,16 @@ func postprocessAndSendLogs(ctx context.Context, server *state.Server, globalCol
 
 	for idx := range transientLogState.LogFiles {
 		logFile := &transientLogState.LogFiles[idx]
+		logFile.ByteSize = int64(logFile.LogLines[len(logFile.LogLines)-1].ByteEnd)
```
seanlinsley, line 332: Since we no longer have the original string at this point in the code, looking at the last log line's `ByteEnd` is the only way to know the file size. Maybe this should be moved to an earlier processing stage?

seanlinsley: Oh and in terms of this code's soundness, I vaguely remember that on production we do receive log files that don't contain any log lines, so this would panic. Should we really be submitting empty snapshots?

input/system/rds/logs.go
seanlinsley: Hmm, it seems the 10k requests on boot is unavoidable because of [API limitations](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DownloadDBLogFilePortion.html). I wonder if instead of using an arbitrary cutoff, we can ask the operating system how much free memory there is, and only use half of that?

runner/logs.go
seanlinsley: Looks like 11% of all log files are empty:
```sql
select count(*), byte_size = 0 as empty from log_files_7d group by 2 order by 1;
  count   | empty 
----------+-------
  6079570 | t
 55500294 | f

seanlinsley: In https://github.com/pganalyze/collector/pull/607/commits/7c6433c56ed260589017bc28ec02c351e1cb737e I've fixed the potential panic, and also made it so we don't submit empty log snapshots.

logs/debug.go
lfittl: I think that should be fine - the main use case for this is when the logs can't be analyzed, and that should still work with this updated output.

lfittl (approved):
:+1: Looks good!

PR #609, opened 2024-10-08
Split up main.go

seanlinsley:
At 737 lines long, I've found it difficult to navigate `main.go`. This PR breaks out the `run` function and its dependent functions into a separate `runner/run.go` file to try and improve that.


msakrejda (approved):
Nice, good call.

PR #610, opened 2024-10-09
Update get_column_stats for Postgres 17

seanlinsley:
As documented on https://github.com/pganalyze/pganalyze-docs/pull/285, Postgres 17 added new columns to `pg_stats` which broke the `pganalyze.get_column_stats` helper function. This PR updates the function to `RETURNS TABLE` so any future additions won't break the function again.

But since that's a backwards-incompatable change requiring `DROP FUNCTION` instead of `CREATE OR REPLACE FUNCTION` this PR also adds an additional runtime check to explain the necessary steps to users.

### DIFF
--- a/input/postgres/relation_column_stats.go
+++ b/input/postgres/relation_column_stats.go
@@ -46,7 +46,7 @@ func GetColumnStats(ctx context.Context, logger *util.Logger, db *sql.DB, global
 				server.SelfTest.HintDbCollectionAspect(dbName, state.CollectionAspectColumnStats, "Please set up"+
 					" the monitoring helper function pganalyze.get_column_stats (%s)"+
 					" or connect as superuser to get column statistics for all tables.", selftest.URLPrinter.Sprint("https://pganalyze.com/docs/install/troubleshooting/column_stats_helper"))
-				logger.PrintWarning("Limited access to table column statistics detected in database %s. Please set up"+
+				logger.PrintInfo("Warning: Limited access to table column statistics detected in database %s. Please set up"+
 					" the monitoring helper function pganalyze.get_column_stats (https://pganalyze.com/docs/install/troubleshooting/column_stats_helper)"+
 					" or connect as superuser, to get column statistics for all tables.", dbName)
 			}
--- a/input/postgres/relation_column_stats.go
+++ b/input/postgres/relation_column_stats.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"strings"
 
 	"github.com/pganalyze/collector/selftest"
 	"github.com/pganalyze/collector/state"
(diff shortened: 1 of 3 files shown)


input/postgres/superuser.go
```diff
@@ -100,3 +100,21 @@ func StatsHelperExists(ctx context.Context, db *sql.DB, statsHelper string) bool
 
 	return enabled
 }
+
+const statsHelperReturnTypeSQL string = `
+SELECT pg_catalog.format_type(prorettype, null)
```
seanlinsley, line 105: This returns `pg_stats` for the old function, and `record` for the new function.

input/postgres/relation_column_stats.go
```diff
@@ -33,7 +46,7 @@ func GetColumnStats(ctx context.Context, logger *util.Logger, db *sql.DB, global
 				server.SelfTest.HintDbCollectionAspect(dbName, state.CollectionAspectColumnStats, "Please set up"+
 					" the monitoring helper function pganalyze.get_column_stats (%s)"+
 					" or connect as superuser to get column statistics for all tables.", selftest.URLPrinter.Sprint("https://pganalyze.com/docs/install/troubleshooting/column_stats_helper"))
-				logger.PrintInfo("Warning: Limited access to table column statistics detected in database %s. Please set up"+
+				logger.PrintWarning("Limited access to table column statistics detected in database %s. Please set up"+
```
seanlinsley, line 49: It looks like this change is causing the [test failure](https://github.com/pganalyze/collector/actions/runs/11263876700/job/31322741758). Thoughts on how to handle this? I could back out the change, but AFAIK this should really be a warning.

msakrejda: The collector integration test output is a mess, but the message [is there](https://github.com/pganalyze/collector/actions/runs/11263876700/job/31322741758#step:6:7828):

> expected no warnings in collector log; test failed:

This is due to the intentional check for warnings [here](https://github.com/pganalyze/collector/blob/main/integration_test/Makefile#L288). So I think our options are:

 1. Leave this log message at Info level
 2. Update guided setup so that column stats helpers are installed, avoiding the warning
 3. Drop that check in the integration test

I think (2) is the best course of action, but that's clearly beyond the scope of this PR. I think we should do (1) for now and do (2) as a follow-up and then update the log level. For what it's worth, it looks like we have a bunch of other `PrintInfo` calls that start their output with "Warning" or "WARNING"...

input/postgres/relation_column_stats.go
```diff
@@ -20,9 +21,21 @@ func GetColumnStats(ctx context.Context, logger *util.Logger, db *sql.DB, global
 	var sourceTable string
 
 	if StatsHelperExists(ctx, db, "get_column_stats") {
-		logger.PrintVerbose("Found pganalyze.get_column_stats() stats helper")
-		sourceTable = "pganalyze.get_column_stats()"
-		server.SelfTest.MarkDbCollectionAspectOk(dbName, state.CollectionAspectColumnStats)
+		if strings.Contains(StatsHelperReturnType(ctx, db, "get_column_stats"), "pg_stats") {
```
msakrejda, line 24: This introduces an additional DB round trip for every `GetColumnStats` call. Since we only do that every full snapshot, though, I doubt it can possibly matter.

msakrejda (approved):

PR #611, opened 2024-10-15
Release 0.60.0

seanlinsley:


CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
...
 
+## 0.60.0      2024-10-15
+
+* Update `get_column_stats` for Postgres 17
+  - The function signature has changed, so must be dropped and recreated
+    https://pganalyze.com/docs/install/troubleshooting/column_stats_helper
+* Handle ambiguous log lines more reliably
+  - This fixes parsing of some log lines for Google CloudSQL
+* Include log text in snapshots instead of uploading to S3
+  - This simplifies the Enterprise Server setup by making object storage opional
```
lfittl, line 11:
```suggestion
  - This simplifies the Enterprise Server setup by making object storage optional
```

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Changelog
 
+## 0.60.0      2024-10-15
+
+* Update `get_column_stats` for Postgres 17
+  - The function signature has changed, so must be dropped and recreated
+    https://pganalyze.com/docs/install/troubleshooting/column_stats_helper
+* Handle ambiguous log lines more reliably
+  - This fixes parsing of some log lines for Google CloudSQL
+* Include log text in snapshots instead of uploading to S3
```
lfittl, line 10: We should reword this to make it clear that its changing the way these uploads are done (i.e. the old way is removed), and put an explicit notice of the impact of this change for Enterprise installs running the collector separately.

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Changelog
 
+## 0.60.0      2024-10-15
+
```
msakrejda, line 4: Per the discussion in our meeting, we could add a notice like
```suggestion
WARNING: For Enterprise Server releases older than 2024.10 using [a separate collector installation](https://pganalyze.com/docs/enterprise/setup/separate-collector-install), this release is partially incompatible. Log Insights and Automated EXPLAIN will not receive any data if using this collector. Enterprise Server installations using an integrated collector are not affected, nor are Scale and Production plans.
```

CHANGELOG.md
msakrejda: Yeah, maybe something like
```suggestion
* Change log upload to include log text directly in snapshots, instead of uploading it separately to S3
```
I think this with the warning I suggested above makes the situation clear.

msakrejda (approved):

CHANGELOG.md
seanlinsley: 👍, but Automated EXPLAIN will still work so I removed that part

PR #612, opened 2024-10-15
Update helper setup to use DROP FUNCTION IF EXISTS

seanlinsley:
Pair PR to https://github.com/pganalyze/pganalyze-docs/pull/287

### DIFF
--- a/runner/generate_helper_sql.go
+++ b/runner/generate_helper_sql.go
@@ -13,7 +13,8 @@ import (
 
 var statsHelpers = []string{
 	// Column stats
-	`CREATE OR REPLACE FUNCTION pganalyze.get_column_stats() RETURNS TABLE(
+	`DROP FUNCTION IF EXISTS pganalyze.get_column_stats;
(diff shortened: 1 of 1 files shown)


runner/generate_helper_sql.go
```diff
@@ -13,7 +13,8 @@ import (
 
 var statsHelpers = []string{
 	// Column stats
-	`CREATE OR REPLACE FUNCTION pganalyze.get_column_stats() RETURNS TABLE(
+	`DROP FUNCTION IF EXISTS pganalyze.get_column_stats;
+CREATE OR REPLACE FUNCTION pganalyze.get_column_stats() RETURNS TABLE(
```
msakrejda, line 17:
```suggestion
CREATE FUNCTION pganalyze.get_column_stats() RETURNS TABLE(
```
NBD, but we don't need the "OR REPLACE" if we DROP IF EXISTS right beforehand.

msakrejda (approved):

PR #619, opened 2024-10-18
Collect column stats using helper on Postgres < 17 despite warning

seanlinsley:
Followup to #610, this makes it so we only fall back to querying `pg_stats` directly on Postgres 17 and older, since the old function definition will still work until the database server is upgraded.


keiko713 (approved):
Change makes sense, but I don't see the context much. Was this discussed in https://github.com/pganalyze/collector/pull/610? Or got some feedback somewhere else?

seanlinsley:
No one's reported it, it was just an improvement I thought up. In the case there would be a permissions error when querying `pg_stats` directly on old Postgres versions, this will allow the collector to still collect column stats until they upgrade to Postgres 17.

keiko713:
Ah nice, thanks for the additional context 👍

PR #620, opened 2024-10-22
Store query texts in temp file

seanlinsley:
#605 succeeded in reducing memory usage, but it also caused collector query time in pg_stat_statements to increase. This PR tries to solve both issues by writing query texts to a tempfile so the query can finish quickly, and then reading each query text in one-by-one to keep memory usage down.

### DIFF
--- a/input/postgres/statements.go
+++ b/input/postgres/statements.go
@@ -5,6 +5,9 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
 	"strings"
 
 	"github.com/guregu/null"
@@ -106,6 +109,7 @@ func GetStatements(ctx context.Context, server *state.Server, logger *util.Logge
 	var extSchema string
 	var extMinorVersion int16
 	var foundExtMinorVersion int16
+	var tmpFile *os.File
 
 	if postgresVersion.Numeric >= state.PostgresVersion17 {
 		extMinorVersion = 11
@@ -230,6 +234,18 @@ func GetStatements(ctx context.Context, server *state.Server, logger *util.Logge
 	statementTextsByFp := make(state.PostgresStatementTextMap)
 	statementStats := make(state.PostgresStatementStatsMap)
 
+	queryKeys := make([]state.PostgresStatementKey, 0)
+	queryStats := make([]state.PostgresStatementStats, 0)
+	queryLength := make([]int, 0)
+	if showtext {
+		tmpFile, err = ioutil.TempFile("", "")
+		if err != nil {
+			return nil, nil, nil, err
+		}
+		defer tmpFile.Close()
+		defer os.Remove(tmpFile.Name())
+	}
+
 	for rows.Next() {
 		var key state.PostgresStatementKey
 		var queryID null.Int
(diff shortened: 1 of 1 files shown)


input/postgres/statements.go
```diff
@@ -230,6 +234,18 @@ func GetStatements(ctx context.Context, server *state.Server, logger *util.Logge
 	statementTextsByFp := make(state.PostgresStatementTextMap)
 	statementStats := make(state.PostgresStatementStatsMap)
 
+	queryKeys := make([]state.PostgresStatementKey, 0)
+	queryStats := make([]state.PostgresStatementStats, 0)
+	queryLength := make([]int, 0)
+	if showtext {
+		tmpFile, err = ioutil.TempFile("", "")
```
msakrejda, line 241: This function is deprecated in Go now; it just [calls another function](https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/io/ioutil/tempfile.go;l=23-26).
```suggestion
		tmpFile, err = os.CreateTemp("", "")
```
I'll clean up other usage, but we should avoid introducing more.

msakrejda (approved):
Minor comment, but looks good.

input/postgres/statements.go
```diff
@@ -253,38 +269,60 @@ func GetStatements(ctx context.Context, server *state.Server, logger *util.Logge
...
 
 	return statements, statementTextsByFp, statementStats, nil
 }
 
-func ignoreIOTiming(postgresVersion state.PostgresVersion, receivedQuery null.String) bool {
+func ignoreIOTiming(postgresVersion state.PostgresVersion, receivedQuery string) bool {
 	// Currently, Aurora gives wildly incorrect blk_read_time and blk_write_time values
 	// for utility statements; ignore I/O timing in this situation.
-	if !postgresVersion.IsAwsAurora || !receivedQuery.Valid {
+	if !postgresVersion.IsAwsAurora || receivedQuery != "" {
```
lfittl, line 321:
```suggestion
	if !postgresVersion.IsAwsAurora || receivedQuery == "" {
```

input/postgres/statements.go
```diff
@@ -253,38 +269,60 @@ func GetStatements(ctx context.Context, server *state.Server, logger *util.Logge
...
-			stats.BlkWriteTime = 0
-		}
-		statementStats[key] = stats
 	}
 
 	if err = rows.Err(); err != nil {
 		return nil, nil, nil, err
 	}
 
+	tmpFile.Seek(0, io.SeekStart)
```
lfittl, line 289: Whilst this doesn't crash when `tmpFile` is not initialized (I don't understand why, but [Go Playground confirms it](tmpFile.Seek(0, io.SeekStart))), it might be best if we wrap this and the for loop below in `if showtext`, just so a future reader doesn't think it might be a problem.

input/postgres/statements.go
```diff
@@ -253,38 +269,60 @@ func GetStatements(ctx context.Context, server *state.Server, logger *util.Logge
...
+	for idx, length := range queryLength {
+		bytes := make([]byte, length)
+		_, err = io.ReadFull(tmpFile, bytes)
+		if err != nil {
+			return nil, nil, nil, err
+		}
+		query := string(bytes)
```
lfittl, line 296: We could technically avoid the memory copy here (see e.g. https://www.reddit.com/r/golang/comments/14xvgoj/converting_string_byte/ for a way of doing it), but not sure if its worth saving on that (maybe worth a benchmark, not necessarily right now)

lfittl (approved):

input/postgres/statements.go
msakrejda: Ah, I think that's because the first thing Seek does is call `File.checkValid`, which [returns an error if File is nil](https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/os/file_posix.go;l=235-240). I agree wrapping this in a `showtext` check is worthwhile, and checking the return value from `Seek` may be worth doing, too, to avoid surprises.

PR #624, opened 2024-10-25
Start collecting query plan information

keiko713:
With this PR, we are going to start collecting query plan information. As a starter, we'll use `aurora_stat_plans` function that is available for newer Aurora instances. In the future, we can also use the extension like [pg_store_plans](https://ossc-db.github.io/pg_store_plans/) which returns a similar shape of plan info and stats.

* base plan information (e.g. explain plan text) is collected every 10 mins at the full snapshot
  * In protobuf, this is `QueryPlanInformation`
* plan stats information (very similar shape as query stats info) is collected every 1 min, and sent with the full snapshot
  * In protobuf, this is `QueryPlanStatistic` (for the stat collected at the full snapshot timing) or `HistoricQueryPlanStatistics` (for the stats collected every 1 min)


keiko713:
@msepga I'm not marking this as ready for review yet as I'd like to check the app side to be sure, but I'm not planning to change much from here (or may not change at all), so you could start reviewing if you feel like so (or totally okay to wait until it's ready for review)!

output/transform/postgres_statements.go
```diff
@@ -67,11 +67,27 @@ func transformQueryStatistic(stats state.DiffedPostgresStatementStats, idx int32
...
+	databaseOid state.Oid
+	userOid     state.Oid
+	queryID     int64
+}
+type StatementKeyToIdx map[postgresStatementKey]int32
+
+func transformPostgresStatements(s snapshot.FullSnapshot, newState state.PersistedState, diffState state.DiffState, transientState state.TransientState, roleOidToIdx OidToIdx, databaseOidToIdx OidToIdx) (snapshot.FullSnapshot, StatementKeyToIdx) {
 	// Statement stats from this snapshot
 	groupedStatements := groupStatements(transientState.Statements, diffState.StatementStats)
+	queryIdxMap := make(StatementKeyToIdx)
```
seanlinsley, line 80: Could you rename this variable to `statementKeyToIdx`? That way it's consistent across all uses in this PR. Or the other direction is fine to, renaming all usage to `queryIdxMap`. I'd just like it to be consistent to make it easier to navigate the code.

runner/queries.go
```diff
@@ -47,21 +47,31 @@ func gatherQueryStatsForServer(ctx context.Context, server *state.Server, global
...
 	newState.UnidentifiedStatementStats = server.PrevState.UnidentifiedStatementStats
+	newState.UnidentifiedPlanStats = server.PrevState.UnidentifiedPlanStats
 	if newState.UnidentifiedStatementStats == nil {
 		newState.UnidentifiedStatementStats = make(state.HistoricStatementStatsMap)
 	}
+	if newState.UnidentifiedPlanStats == nil {
+		newState.UnidentifiedPlanStats = make(state.HistoricPlanStatsMap)
+	}
 	newState.UnidentifiedStatementStats[timeKey] = diffedStatementStats
+	newState.UnidentifiedPlanStats[timeKey] = diffedPlanStats
```
seanlinsley, line 74: I feel like interleaving these two code paths is going to make it harder to maintain going forward. It'd be better to put all of the new code after the existing code.

output/transform/postgres_statements.go
```diff
@@ -67,11 +67,27 @@ func transformQueryStatistic(stats state.DiffedPostgresStatementStats, idx int32
...
 	// Statement stats from this snapshot
 	groupedStatements := groupStatements(transientState.Statements, diffState.StatementStats)
+	statementKeyToIdx := make(StatementKeyToIdx)
 	for key, value := range groupedStatements {
 		idx := upsertQueryReferenceAndInformation(&s, transientState.StatementTexts, roleOidToIdx, databaseOidToIdx, key, value)
+		for _, queryId := range value.queryIDs {
+			sKey := postgresStatementKey{
+				databaseOid: key.databaseOid,
+				userOid:     key.userOid,
+				queryID:     queryId,
```
lfittl, line 87: As discussed with Keiko in our 1:1, this currently does not keep the top-level identifier here - this is intentional because we currently do not send top_level to the server side. When we do so in the future we should expand this logic here as well to include it.

protobuf/full_snapshot.proto
```diff
@@ -625,3 +629,37 @@ message CustomTypeInformation {
     string type = 2;
   }
 }
+
+message QueryPlanInformation {
+  int32 query_idx = 1;
+  int64 plan_id = 2;
+  string explain_plan = 3;
+  google.protobuf.Timestamp plan_captured_time = 4;
```
lfittl, line 637: It might be helpful to track a boolean flag here as well, to remember whether the `plan_type` was `actual` - that way we could consider creating query samples for that in the future on the server side, if someone has enabled frequent recapturing of plans.

protobuf/full_snapshot.proto
```diff
@@ -625,3 +629,37 @@ message CustomTypeInformation {
...
+}
+
+message QueryPlanStatistic {
+  int32 query_idx = 1;
+  int64 plan_id = 2;
```
lfittl, line 642: It may be useful to instead point to a `plan_idx`, and follow our standard reference/information/stats logic in the snapshot - that should also make the server-side processing easier, I believe.

output/transform/postgres_plans.go
```diff
@@ -0,0 +1,107 @@
...
+func transformPostgresPlans(s snapshot.FullSnapshot, newState state.PersistedState, diffState state.DiffState, transientState state.TransientState, roleOidToIdx OidToIdx, databaseOidToIdx OidToIdx, statementKeyToIdx StatementKeyToIdx) snapshot.FullSnapshot {
+	planInformations := []*snapshot.QueryPlanInformation{}
+	planStats := []*snapshot.QueryPlanStatistic{}
+	for pKey, plan := range transientState.Plans {
+		sKey := postgresStatementKey{
+			databaseOid: pKey.DatabaseOid,
+			userOid:     pKey.UserOid,
+			queryID:     pKey.QueryID,
+		}
+		queryIdx, exist := statementKeyToIdx[sKey]
```
lfittl, line 20: As discussed, what can happen here is that we have multiple statement keys pointing to the same queryIdx (either because of a case where top_level is the only difference, or because we have multiple query IDs pointing to the same fingerprint, and as such queryIdx).

The easiest solution might be to make both `planInformations` and `planStats` a map that's indexed by `queryIdx` and `planId` (or a `planIdx` if you introduce that), and then do the addition of stats similar to how `groupStatements` does it.

lfittl: For the plan information we could keep the more recently captured plan (if we have two plans for the same query index and plan ID).

lfittl (commented):
Excited about this!

protobuf/full_snapshot.proto
keiko713: FYI I went with enum even though we don't collect "no plan" plan type

output/transform/postgres_plans.go
```diff
@@ -0,0 +1,164 @@
...
+
+type planValue struct {
+	plan      state.PostgresPlan
+	planStats state.DiffedPostgresStatementStats
+}
+
+func groupPlans(plans state.PostgresPlanMap, statsMap state.DiffedPostgresPlanStatsMap, queryIDKeyToIdx QueryIDKeyToIdx) map[planKey]planValue {
+	// We group multiple different queryIDs with the same fingerprints with one "statementKey"
+	// (e.g. SELECT * FROM users WHERE name in ($1) and SELECT * FROM users WHERE name in ($1, $2) have the same fingerprints even though queryID will be different)
+	// With query plans, we also group plans and plan stats with a statementKey + planID.
```
keiko713, line 24: Example (different query ID, same fingerprint, same plan ID - we group these two plans together):

```
-[ RECORD 1 ]-----------------------------------------------------------------------------
queryid | 1649607691878460349
planid  | 683693185
query   | select queryid, planid from aurora_stat_plans() where query in ($1) limit $2
-[ RECORD 2 ]-----------------------------------------------------------------------------
queryid | 8308867196947722868
planid  | 683693185
query   | select queryid, planid from aurora_stat_plans() where query in ($1, $2) limit $3
```

protobuf/shared.proto
```diff
@@ -64,6 +64,11 @@ message QueryReference {
   bytes fingerprint = 3;
 }
 
+message QueryPlanReference {
```
lfittl, line 67: Technically this could be in the `full_snapshot.proto` file, since we don't use these in compact snapshots (yet), but I think its also fine to have here.

protobuf/shared.proto
```diff
@@ -64,6 +64,11 @@ message QueryReference {
...
+  int32 query_idx = 1;
+  int64 plan_id = 2;
```
lfittl, line 69: Should we call this `original_plan_id` here as well, for clarity? (in case we ever assign a collector-side plan fingerprint)

input/postgres/plans.go
```diff
@@ -0,0 +1,106 @@
...
+FROM
+	aurora_stat_plans(%t)
+WHERE
+	plan_type IN ('estimate', 'actual')`
+
+// GetPlans collects query execution plans and stats
+func GetPlans(ctx context.Context, server *state.Server, logger *util.Logger, db *sql.DB, globalCollectionOpts state.CollectionOpts, postgresVersion state.PostgresVersion, showtext bool) (state.PostgresPlanMap, state.PostgresPlanStatsMap, error) {
+	var err error
+
+	// Currently only collects this with Aurora using aurora_stat_plans function
```
lfittl, line 33:
```suggestion
	// Only collect this with Aurora using aurora_stat_plans function for now
```

output/transform/postgres_statements.go
```diff
@@ -20,6 +20,7 @@ func groupStatements(statements state.PostgresStatementMap, statsMap state.Diffe
 
 		// Note we intentionally don't include sKey.TopLevel here, since we don't (yet)
 		// separate statistics based on that attribute in the pganalyze app
+		// Different TopLevel stats will be combined currently
```
lfittl, line 23:
```suggestion
		// As such, pg_stat_statements entries that only differ by toplevel will be aggregated here
```

protobuf/shared.proto
keiko713: Yeah good point, I did debate about it. I think I added here because most of the references are in `shared.proto` even when it's only used with full snapshot (like FunctionReference, IndexReference). I'll keep this here for that reason.

protobuf/shared.proto
keiko713: Good point, let's update the name, I think it's a good change 👍

lfittl (approved):
:+1:

PR #627, opened 2024-10-30
Fix over-eager PII filtering

msakrejda:
In a25800b1a1e32c62dcc74587a2dbf7503e977211, we changed the way
redaction works for PII filtering. The change did not account for
detail log lines, leading to all log lines always being redacted.

Update the redaction mechanism to check for a parent log line.

### DIFF
--- a/logs/replace.go
+++ b/logs/replace.go
@@ -4,6 +4,7 @@ import (
 	"slices"
 	"sort"
 
+	"github.com/google/uuid"
 	"github.com/pganalyze/collector/state"
 )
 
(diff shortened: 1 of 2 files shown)


logs/replace.go
```diff
@@ -17,7 +18,7 @@ func ReplaceSecrets(logLines []state.LogLine, filterLogSecret []state.LogSecretK
 		}
 	}
 	for idx, logLine := range logLines {
-		if filterUnidentified && logLines[idx].Classification == 0 {
+		if filterUnidentified && logLines[idx].Classification == 0 && logLines[idx].ParentUUID == uuid.Nil {
```
msakrejda, line 21: I'm not sure if this is actually sufficient, since this would mean that child log lines of unidentified parent log lines would never be redacted, right? Is there a better way to handle this?

seanlinsley: Is it even possible for an unidentified log line to have children? I'd have assumed the parent-child association relies on being able to parse the log line.

msakrejda: I think so: my reading of [the logic to assign parents to log lines](https://github.com/pganalyze/collector/blob/main/logs/analyze.go#L2212-L2246) is that it doesn't care what the parent log line actually is. That code is pretty intricate, though, so I may be missing something.

keiko713 (approved):

logs/replace.go
msakrejda: Per separate discussion, this may still be an issue, but it's an edge case and we can fix that in a follow-up patch.

PR #628, opened 2024-11-01
Add collector query runner

seanlinsley:
This query runner will initially allow users to request that the collector run `EXPLAIN ANALYZE` for a query on their behalf, though in the future it could be expanded to support maintenance tasks like `REINDEX`.

For now this is disabled by default, and can be enabled with the `enable_query_runner` setting or `PGA_ENABLE_QUERY_RUNNER` environment variable.


protobuf/compact_activity_snapshot.proto
```diff
@@ -363,3 +364,12 @@ message VacuumProgressStatistic {
   int64 indexes_total = 11; // Postgres 17+
   int64 indexes_processed = 12; // Postgres 17+
 }
+
+message QueryRun {
```
keiko713, line 368: I didn't see this well previously when Maciek mentioned, but now I'm reviewing the code, it is indeed confusing to see 3 things are named the same and representing slightly different things:
* QueryRun in server_message.proto: websocket incoming message
* QueryRun in compact_activity_snapshot.prot: websocket outgoing message
* QueryRun in state.go: kinda mix of both

That said, I don't think having three (or two) different names will help here that much honestly, or I at least don't have better suggestions here, so I think it's fine to keep it as is.

runner/query_run.go
```diff
@@ -0,0 +1,117 @@
...
+			server.QueryRunsMutex.Unlock()
+			continue
+		}
+
+		// The comment doesn't include /* pganalyze-collector */ so that query runs are reported separately in pganalyze
+		comment := fmt.Sprintf("/* pganalyze:no-alert,pganalyze-query-run:%d */ ", query.Id)
+		prefix := ""
+		result := ""
+		if query.Type == pganalyze_collector.QueryRunType_EXPLAIN {
+			prefix = "EXPLAIN (ANALYZE, BUFFERS, FORMAT JSON) "
```
keiko713, line 86: Do we want to add VERBOSE here? With auto_explain, we mention to enable verbose, and I think we do show things from verbose output if available.

runner/query_run.go
```diff
@@ -0,0 +1,117 @@
...
+		server.QueryRuns[idx].FinishedAt = time.Now()
+		server.QueryRuns[idx].Result = result
+		if err != nil {
+			server.QueryRuns[idx].Error = err.Error()
+		}
+		server.QueryRunsMutex.Unlock()
+	}
+}
+
+// Removes old query runs that have finished
```
keiko713, line 106: Wanted to double check: As this is the only place removing server.QueryRuns, we will be sending out duplicated query runs info even if it's already finished at minimum for 10 minutes?

keiko713 (commented):
Left some comments but overall looking good to me. It would be nice to see the websocket processing part as well before approving (aware of the draft PR, but looks like it has some remaining TODOs).

I have one concern for how to run query, which I'll drop note in Slack.

runner/query_run.go
```diff
@@ -0,0 +1,125 @@
...
+func cleanup(server *state.Server) {
+	server.QueryRunsMutex.Lock()
+	queryRuns := make([]state.QueryRun, 0)
+	for _, query := range server.QueryRuns {
+		if time.Since(query.FinishedAt) >= 10*time.Minute {
```
msakrejda, line 119: Isn't this check inverted?

state/state.go
```diff
@@ -270,6 +283,10 @@ type Server struct {
 	WebSocket      atomic.Pointer[websocket.Conn]
 	Pause          atomic.Bool
 
+	// State to track requests for the collector to run a query on behalf of a user
```
msakrejda, line 286: This is the requests _and_ the results, right?

runner/query_run.go
seanlinsley: It seems to be correct to me: if the query finished running 10+ minutes ago, this code removes it.

runner/query_run.go
```diff
@@ -0,0 +1,125 @@
...
+			server.QueryRunsMutex.Unlock()
+			continue
+		}
+
+		// We don't include QueryMarkerSQL so query runs are reported separately in pganalyze
+		comment := fmt.Sprintf("/* pganalyze:no-alert,pganalyze-query-run:%d */ ", query.Id)
+		prefix := ""
+		result := ""
+		if query.Type == pganalyze_collector.QueryRunType_EXPLAIN {
+			prefix = "EXPLAIN (ANALYZE, VERBOSE, BUFFERS, FORMAT JSON) "
```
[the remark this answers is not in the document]
seanlinsley: Wouldn't the collector already be capturing those settings separately? I'm also hesitant to add version checking here unless we have a clear use case for the data.

runner/query_run.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+
+		err = db.QueryRowContext(ctx, comment+prefix+query.QueryText).Scan(&result)
+
+		if query.Type == pganalyze_collector.QueryRunType_EXPLAIN {
+			// Run EXPLAIN ANALYZE a second time to get a warm cache result
```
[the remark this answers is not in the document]
seanlinsley: Yes, both for deadlocks and timeouts the second run could succeed. I'll update the code to run the query a third time if the first run failed out but the second run succeeded.

runner/query_run.go
seanlinsley: Ah, yes it is. Good catch.

runner/query_run.go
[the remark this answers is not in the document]
msakrejda: We've discussed that some query tuning query variants might essentially be just the baseline query with some planner settings turned off, so having `SETTINGS` would be a nice sanity check that we captured the correct thing. I guess the approach here doesn't support passing custom settings anyway, so it's moot for now, but I think it's something we'll want pretty soon. If it's easier to do that in a follow up patch, that's fine.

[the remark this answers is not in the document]
keiko713: (keiko memo) having settings will give us the following:

```
pgaweb=# set enable_nestloop = off;
SET
pgaweb=# explain (analyze, settings) select 1;
                                     QUERY PLAN                                     
------------------------------------------------------------------------------------
 Result  (cost=0.00..0.01 rows=1 width=4) (actual time=0.004..0.005 rows=1 loops=1)
 Settings: enable_nestloop = 'off'
 Planning Time: 0.049 ms
 Execution Time: 0.330 ms
(4 rows)
```

but yeah, this PR doesn't support passing custom settings so I think it's okay to add SETTINGS when we make that changes

runner/query_run.go
```diff
@@ -0,0 +1,131 @@
...
+			logger = logger.WithPrefixAndRememberErrors(server.Config.SectionName)
+			cleanupInterval := time.NewTicker(5 * time.Minute)
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				case <-cleanupInterval.C:
+					cleanup(server)
+				default:
+					if server.Config.EnableQueryRunner {
```
keiko713, line 30: Do you place this if statement here instead of using this to early return because you want to run cleanup? Curious why you're putting here.

keiko713 (approved):
Confirmed that it worked with the app side code, looks good to me 👍

runner/query_run.go
```diff
@@ -0,0 +1,131 @@
...
+
+		// We don't include QueryMarkerSQL so query runs are reported separately in pganalyze
+		comment := fmt.Sprintf("/* pganalyze:no-alert,pganalyze-query-run:%d */ ", query.Id)
+		prefix := ""
+		result := ""
+		if query.Type == pganalyze_collector.QueryRunType_EXPLAIN {
+			prefix = "EXPLAIN (ANALYZE, VERBOSE, BUFFERS, FORMAT JSON) "
+		}
+
+		err = db.QueryRowContext(ctx, comment+prefix+query.QueryText).Scan(&result)
```
keiko713, line 91: In today's meeting, we talked about wrapping this in the transaction (maybe the below is not needed as it's a simple EXPLAIN - or maybe we could wrap this in the transaction even wider).

Also I completely forgot about this, but I think we actually wanted to run this 3 times, then record the 3rd one (or fastest one), and we don't have that logic in place in this PR.

seanlinsley: Why would we want to run it three times? It should already be a warm cache after two runs (which this PR already does).

runner/websocket.go
```diff
@@ -143,8 +143,17 @@ func connect(ctx context.Context, server *state.Server, globalCollectionOpts sta
 				server.Grant.Store(&grant)
 			} else if message.GetPause() != nil {
 				server.Pause.Store(message.GetPause().Pause)
-			} else if message.GetExplainRun() != nil {
-				logger.PrintVerbose("ExplainRun: %v", message.GetExplainRun()) // TODO
+			} else if message.GetQueryRun() != nil {
+				q := message.GetQueryRun()
+				logger.PrintVerbose("Query run %d received: %s", q.Id, q.QueryText)
+				server.QueryRunsMutex.Lock()
+				server.QueryRuns = append(server.QueryRuns, state.QueryRun{
```
seanlinsley, line 150: As discussed during today's meeting, this should ignore duplicate requests (based on the ID) in case the server has a retry mechanism in the future. Maybe this should be a map instead of an array?

runner/query_run.go
keiko713: Ah 🤦 I wasn't looking at the line below this one, sorry, my bad.
So this is the first run, and line 96 is second, then there is the 3rd one in line 100 (and the explain without analyze is in line 106). Please disregard what I said re: running 3 times.

runner/query_run.go
```diff
@@ -0,0 +1,134 @@
...
+
+		if query.Type == pganalyze_collector.QueryRunType_EXPLAIN {
+			// Rollback any changes the query may perform
+			db.ExecContext(ctx, postgres.QueryMarkerSQL+"BEGIN READ ONLY")
```
keiko713, line 95: Nice update! Though, this means that we only runs queries in a read only transaction from the second run (IOW, the first run will still executed as a normal transaction) 🤔 Did you mean to add this to line 88 maybe?

PR #629, opened 2024-11-06
Add DB_URL_FILE and DB_PASSWORD_FILE

msakrejda:
Adds DB_URL_FILE and DB_PASSWORD_FILE

This lets users specify these credentials via the filesystem to better
support NixOS flakes and systemd credentials.

Based on #595

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -47,9 +47,11 @@ type ServerConfig struct {
 	EnableLogExplain bool `ini:"enable_log_explain"`
 
 	DbURL                 string `ini:"db_url"`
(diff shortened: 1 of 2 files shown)


msakrejda:
@Munksgaard could you check if this version works for you when you have a chance?

keiko713 (approved):
Looks good to me 👍 We'll need a docs PR too

Munksgaard:
Looks like it works here as well. Thanks!

msakrejda:
@Munksgaard great, thanks for confirming. We'll include this in the next release.

PR #630, opened 2024-11-07
Add transform test for query plan

keiko713:
Follow-up of https://github.com/pganalyze/collector/pull/624
While debugging some issue, I noticed that this test exist. To isolate the issue, I started adding the test. While it turned out not to be the collector side issue, I think it won't hurt to expand this test.

### DIFF
--- a/output/transform/transform_test.go
+++ b/output/transform/transform_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"testing"
+	"time"
 
 	"github.com/pganalyze/collector/output/pganalyze_collector"
 	"github.com/pganalyze/collector/output/transform"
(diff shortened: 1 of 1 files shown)


msakrejda (approved):

msakrejda:
I think it may have hurt to expand this test. It maybe caused [this Release failure](https://github.com/pganalyze/collector/actions/runs/11823836245/job/32943994958) (NBD, but inconvenient). I find the `expectedAlt` mechanism here completely impossible to maintain, so on second thought, maybe we shouldn't touch this: should we consider reverting?

Maybe if we touch anything here, we should refactor this first, to find another way to avoid this diff issue. Maybe we should just sort these arrays before printing them?

seanlinsley:
Maybe the results could be sorted before they're serialized to JSON with `sort.SliceStable`? I fixed a similar issue in #618 recently (the code already called `sort.SliceStable` but it was building the `logFileContent` expected result before the sorting occurred).

PR #632, opened 2024-11-13
Use groupadd instead of addgroup

msakrejda:
The latter seems to no longer be available on Amazon Linux 2023. Since
addgroup is just a convenience wrapper around groupadd and we are only
using its basic functionality, we can switch to using groupadd
directly.

### DIFF
--- a/packages/src/rpm-systemd/post.sh
+++ b/packages/src/rpm-systemd/post.sh
@@ -8,7 +8,7 @@ if ! getent passwd pganalyze > /dev/null; then
 fi
 
 if ! getent group pganalyze > /dev/null; then
-  addgroup --system --quiet pganalyze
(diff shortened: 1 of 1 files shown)


packages/src/rpm-systemd/post.sh
```diff
@@ -8,7 +8,7 @@ if ! getent passwd pganalyze > /dev/null; then
 fi
 
 if ! getent group pganalyze > /dev/null; then
-  addgroup --system --quiet pganalyze
+  groupadd --system pganalyze
```
msakrejda, line 11: `--quiet` is not an option for `groupadd`, but the command produces no output anyway, so the flag is moot.

keiko713 (approved):

PR #635, opened 2024-11-18
Aggregate stats for partition tables

seanlinsley:
Split out from #633, this PR aggregates table and index stats for partition parent tables. Most of the stats were previously zero except for the `size_bytes` query in `input/postgres/relation_stats.go` which this removes.

### DIFF
--- a/output/transform/merge_partition_sizes.go
+++ b/output/transform/merge_partition_sizes.go
@@ -6,6 +6,24 @@ import (
 )
 
 func mergePartitionSizes(s snapshot.FullSnapshot, newState state.PersistedState, ts state.TransientState, databaseOidToIdx OidToIdx) snapshot.FullSnapshot {
+	for idx, rel := range s.RelationInformations {
+		if !rel.HasParentRelation || rel.PartitionBoundary == "" {
+			continue
+		}
+		stat := s.RelationStatistics[idx]
+		parent := s.RelationStatistics[rel.ParentRelationIdx]
+		parent.NTupIns += stat.NTupIns
+		parent.NTupUpd += stat.NTupUpd
+		parent.NTupDel += stat.NTupDel
+		parent.NTupHotUpd += stat.NTupHotUpd
+		parent.NLiveTup += stat.NLiveTup
+		parent.NDeadTup += stat.NDeadTup
+		parent.SizeBytes += stat.SizeBytes
+		parent.ToastSizeBytes += stat.ToastSizeBytes
+		parent.CachedDataBytes += stat.CachedDataBytes
+		parent.CachedToastBytes += stat.CachedToastBytes
+	}
+
 	for idx, info := range s.IndexInformations {
 		rel := s.RelationInformations[info.RelationIdx]
 		if !rel.HasParentRelation || rel.PartitionBoundary == "" {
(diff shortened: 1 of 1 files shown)


input/postgres/relation_stats.go
```diff
@@ -20,10 +20,7 @@ locked_relids_with_parents AS (
 	UNION SELECT relid FROM locked_relids
 )
 SELECT c.oid,
-			 COALESCE(pg_catalog.pg_table_size(c.oid), 0) +
-			 COALESCE((SELECT pg_catalog.sum(pg_catalog.pg_table_size(inhrelid))
-			    FROM pg_catalog.pg_inherits
-				 WHERE inhparent = c.oid), 0) AS size_bytes,
+			 COALESCE(pg_catalog.pg_table_size(c.oid), 0) AS size_bytes,
```
msakrejda, line 23: You mention this in the PR description, but I don't follow why this change is being made. Can you clarify?

seanlinsley: This part of the query aggregated the table sizes from child partitions so the parent table wasn't zero-sized. But it was also inaccurate because it counted TOAST data as table data, which is why if you look at a partition parent table in pganalyze today, none of them will show TOAST usage even though the child tables use TOAST.

Instead, this PR handles this in `merge_partition_sizes.go` with correct handling of TOAST data.

output/transform/merge_partition_sizes.go
```diff
@@ -0,0 +1,56 @@
...
+	for idx, info := range s.IndexInformations {
+		rel := s.RelationInformations[info.RelationIdx]
+		if !rel.HasParentRelation || rel.PartitionBoundary == "" {
+			continue
+		}
+		for parentIdx, pi := range s.IndexInformations {
+			if pi.RelationIdx != rel.ParentRelationIdx {
+				continue
+			}
+			if info.IndexType == pi.IndexType && info.IsUnique == pi.IsUnique && intArrayEqual(info.Columns, pi.Columns) {
```
msakrejda, line 18: This will spuriously match different expression indexes on the same columns, but I think that's okay: I don't see a great way to avoid that without a considerably more expensive check, and it's unlikely we'll run into that in practice.

msakrejda (approved):

output/transform/merge_partition_sizes.go
```diff
@@ -0,0 +1,56 @@
...
+				s.IndexStatistics[parentIdx].SizeBytes += s.IndexStatistics[idx].SizeBytes
```
keiko713, line 19: I forgot what it was previously (when it was in the previous PR), but are we only tracking the size of Index stats?

output/transform/postgres_relations.go
```diff
@@ -253,20 +253,17 @@ func transformPostgresRelations(s snapshot.FullSnapshot, newState state.Persiste
...
-						IdxScan:     indexStats.IdxScan,
-						IdxTupRead:  indexStats.IdxTupRead,
-						IdxTupFetch: indexStats.IdxTupFetch,
-						IdxBlksRead: indexStats.IdxBlksRead,
-						IdxBlksHit:  indexStats.IdxBlksHit,
-						CachedBytes: index.CachedBytes,
-					}
-					s.IndexStatistics = append(s.IndexStatistics, &statistic)
-				}
+				indexStats := diffedSchemaStats.IndexStats[index.IndexOid]
```
keiko713, line 256: We discussed this in https://github.com/pganalyze/collector/pull/633#discussion_r1843234090 too, but I think your plan here is to have `s.IndexInformations` and `s.IndexStatistics` as 1:1.
This sadly still doesn't guarantee it as when `diffedSchemaStatsExist` is somehow false, it'll only create `s.IndexInformations` but not corresponding `s.IndexStatistics`.

As an `else` statement, you may want to add some simple thing like the following for the completeness sake.

```golang
			} else {
				s.IndexStatistics = append(s.IndexStatistics, &snapshot.IndexStatistic{
					IndexIdx:    indexIdx,
				})
			}
```

output/transform/merge_partition_sizes.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	for idx, rel := range s.RelationInformations {
+		if !rel.HasParentRelation || rel.PartitionBoundary == "" {
+			continue
+		}
+		stat := s.RelationStatistics[idx]
+		parent := s.RelationStatistics[rel.ParentRelationIdx]
+		parent.NTupIns += stat.NTupIns
+		parent.NTupUpd += stat.NTupUpd
```
keiko713, line 32: Note: currently, Index Advisor is relying on this relation stats info.
Since the current parent table doesn't have the "writes" information (as these value being 0), it was hard (well, impossible) for the partitioned table to have a write optimized as a table default config as writes per min was always being 0.
With this change, this will change and maybe some partitioned tables will now have write optimized config to run the index advisor, so people may see some different index suggestions because of this.
I don't think it's a bad change (rather, the current behavior is inaccurate so it is a good change), but I think we want to keep this in mind and anticipate some potential confusion.

PR #637, opened 2024-11-21
Shut down WebSocket after write error

seanlinsley:
If the WebSocket encounters an error when writing data to the connection, the previous implementation called `cancelConn` but that wouldn't actually shut down the connection because it then `return`ed, shutting down the goroutine that would be listening for that cancel event.

In that case `server.WebSocket.Swap(nil)` would never be called, so the collector would continue trying to submit snapshots to the dead websocket connection.

This PR resolves that by moving the cleanup code to a separate function so it can be called directly when a write error occurs.


msakrejda (approved):
Looks good. Thanks for the quick fix.

PR #639, opened 2024-11-26
Fix flaky transform_test.go

msakrejda:
Since snapshots include query references and query plan references,
and since those can be orderered nondeterministically, the test can
fail if the snapshot is built with fields in an unexpected order.

The existing test tries to account for that by enumerating all
possible combinations, but this is very difficult to maintain, and
very difficult to scale to additional fields with this pattern.

Instead, rewrite snapshots to a canonical format to ensure comparisons
are stable. We can't just sort the array fields before serializing the
snapshot for comparison, since QueryReferences and similar items are
referenced by an index into their array. To work around that, we sort
the QueryReferences (tracking their original index), and rewrite all
entries that reference them to reference their new index instead.

See discussion in #630 [1].

[1]: https://github.com/pganalyze/collector/pull/630#issuecomment-2495113567

### DIFF
--- a/output/transform/transform_test.go
+++ b/output/transform/transform_test.go
@@ -3,6 +3,8 @@ package transform_test
 import (
 	"encoding/binary"
 	"encoding/json"
+	"slices"
+	"sort"
 	"testing"
 	"time"
 
@@ -46,9 +48,9 @@ func TestStatements(t *testing.T) {
 	diffState.PlanStats[pKey2] = state.DiffedPostgresStatementStats{Calls: 24}
 
 	actual := transform.StateToSnapshot(newState, diffState, transientState)
+	makeCanonical(actual)
 	actualJSON, _ := json.Marshal(actual)
 
-	// Query: 0, 1, Plan: 0, 1
 	expected := pganalyze_collector.FullSnapshot{
 		Config:             &pganalyze_collector.CollectorConfig{},
 		CollectorStatistic: &pganalyze_collector.CollectorStatistic{},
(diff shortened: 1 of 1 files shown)


output/transform/transform_test.go
```diff
@@ -66,189 +68,198 @@ func TestStatements(t *testing.T) {
 		ServerStatistic: &pganalyze_collector.ServerStatistic{},
 		Replication:     &pganalyze_collector.Replication{},
 		QueryReferences: []*pganalyze_collector.QueryReference{
-			&pganalyze_collector.QueryReference{
```
msakrejda, line 69: This is unrelated, but element types can be inferred when the containing array type is specified, so including the type here is unnecessary.

keiko713:
Oh noes, I didn't realize that you were working on this and created this PR 🙃 https://github.com/pganalyze/collector/pull/640
My fix is simpler as it's a bug fix but maybe this is better as it's addressing the part that the test gets complex as we add more things in the future. I'll close my PR

msakrejda:
Oh sorry! Technically I did mention in Slack I would try to work on this, but it was a couple of weeks ago now :grimacing: .

keiko713:
I haven't looked into the code of `makeCanonical` closely, but I just ran test locally 10 times in a row with this branch and it sadly fails consistently currently. Maybe it's missing to consider some situation?

msakrejda:
Interesting, thanks for confirming. I ran this a bunch of times locally and had no errors before submitting, but I just ran it in a loop and it did fail eventually. Looking into it.

keiko713 (approved):

PR #640, opened 2024-11-26
Fix expected result set of transform test

keiko713:
The test was having the wrong expected result set, so I updated it. Without this fix, test was consistently failing (e.g. if I run test 10 times, it guaranteed to fail a few times), but with this fix, I ran test 50 times and it was working fine (I also triple checked the logic).

https://github.com/pganalyze/collector/pull/630#issuecomment-2495113567

I agree with others that the current testing logic is not ideal and also very hard to maintain. I thought about removing this completely, but it's a nice test to have and it was still fixable for now (as it had a clear bug) so I went forward and fixed it. We shouldn't expand this test any further and if we want to add something, we should reconsider how we do the test.
I'm not so sure if the sorting works, this test is passing input to the blackbox (`transform.StateToSnapshot`) and we don't know in which combination the blackbox is going to return the result with.

### DIFF
--- a/output/transform/transform_test.go
+++ b/output/transform/transform_test.go
@@ -48,7 +48,7 @@ func TestStatements(t *testing.T) {
 	actual := transform.StateToSnapshot(newState, diffState, transientState)
 	actualJSON, _ := json.Marshal(actual)
 
-	// Query: 0, 1, Plan: 0, 1
+	// Query: 0, 1, Plan: 0, 1 (w/ QueryIdx 1)
 	expected := pganalyze_collector.FullSnapshot{
 		Config:             &pganalyze_collector.CollectorConfig{},
 		CollectorStatistic: &pganalyze_collector.CollectorStatistic{},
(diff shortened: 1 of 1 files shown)


output/transform/transform_test.go
```diff
@@ -134,7 +134,7 @@ func TestStatements(t *testing.T) {
 	}
 	expectedJSON, _ := json.Marshal(expected)
 
-	// Query: 1, 0, Plan: 0, 1
+	// Query: 1, 0, Plan: 0, 1 (w/ QueryIdx 0)
```
keiko713, line 137: note: the "bug" part is; I forgot that I was using the second query as a plan's query, so the QueryIdx of QueryPlanReference does change (previously, it was using the fixed number `1` here).
Now QueryPlanReference has the right set of combinations.

keiko713:
Closing in favor of https://github.com/pganalyze/collector/pull/639

PR #641, opened 2024-11-26
Simplify the query run to make sure to report the correct error

keiko713:
Currently, if the query is failed to run with some errors other than the timeout (e.g. syntax error), the error won't be correctly reported back and the error becomes "pq: current transaction is aborted, commands ignored until end of transaction block", as the second run's error will be reported.

```
pgaweb=# BEGIN READ ONLY;
BEGIN
pgaweb=*# select pg_sleep('abc');
ERROR:  invalid input syntax for type double precision: "abc"
LINE 1: select pg_sleep('abc');
                        ^
pgaweb=!# select pg_sleep('abc');
ERROR:  current transaction is aborted, commands ignored until end of transaction block
```

I updated and simplified a bit to avoid this, but I'm not actually sure if this is what we want. Looks like currently we are not running the query 3 times unless the 1. error 2. success case. Maybe the intention here is that the first run failed due to the timeout, but second run went well because some half-cached data from the first run there? If so, I'm happy to update the logic, but I think we need to narrow down the error checking and try the second run only when the first run failed with the timeout error.

### DIFF
--- a/runner/query_run.go
+++ b/runner/query_run.go
@@ -94,18 +94,12 @@ func run(ctx context.Context, server *state.Server, collectionOpts state.Collect
 			db.ExecContext(ctx, postgres.QueryMarkerSQL+"BEGIN READ ONLY")
 
 			err = db.QueryRowContext(ctx, comment+prefix+query.QueryText).Scan(&result)
-			firstErr := err
 
-			// Run EXPLAIN ANALYZE a second time to get a warm cache result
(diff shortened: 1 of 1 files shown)


seanlinsley:
>Maybe the intention here is that the first run failed due to the timeout, but second run went well because some half-cached data from the first run there? 

Yes, that's the intention. I put together a different way of fixing this that doesn't change the flow of the code: #642

PR #642, opened 2024-11-26
Improve query runner error handling

seanlinsley:
This PR updates the query runner to submit the first error it sees, to avoid reporting an unhelpful `current transaction is aborted` error back to the server. `BEGIN READ ONLY` was a late addition to #628, so this wasn't caught in manual testing.

This also fixes an error I saw while testing this, where `server.QueryRuns` is nil:
```
panic: assignment to entry in nil map

goroutine 99 [running]:
runner.connect.func3()
    runner/websocket.go:150 +0x4ac
created by runner.connect in goroutine 64
    runner/websocket.go:113 +0xa44

### DIFF
--- a/runner/query_run.go
+++ b/runner/query_run.go
@@ -84,31 +84,25 @@ func run(ctx context.Context, server *state.Server, collectionOpts state.Collect
 
 		// We don't include QueryMarkerSQL so query runs are reported separately in pganalyze
 		comment := fmt.Sprintf("/* pganalyze:no-alert,pganalyze-query-run:%d */ ", query.Id)
-		prefix := ""
 		result := ""
-		if query.Type == pganalyze_collector.QueryRunType_EXPLAIN {
-			prefix = "EXPLAIN (ANALYZE, VERBOSE, BUFFERS, FORMAT JSON) "
-		}
 
 		if query.Type == pganalyze_collector.QueryRunType_EXPLAIN {
-			// Rollback any changes the query may perform
-			db.ExecContext(ctx, postgres.QueryMarkerSQL+"BEGIN READ ONLY")
-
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: runner/query_run.go [hunk 1]
```diff
@@ -94,7 +95,7 @@ func run(ctx context.Context, server *state.Server, collectionOpts state.Collect
 			db.ExecContext(ctx, postgres.QueryMarkerSQL+"BEGIN READ ONLY")
 
 			err = db.QueryRowContext(ctx, comment+prefix+query.QueryText).Scan(&result)
-			firstErr := err
+			firstErr = err
 
 			// Run EXPLAIN ANALYZE a second time to get a warm cache result
```
keiko713, line 100: In https://github.com/pganalyze/collector/pull/641#issuecomment-2501026138, you mentioned that this is executed even if the firstErr is not nil for the case of first one failed with a timeout. However, even though the cause was a timeout, the second one will fail regardless:

```
pgaweb=# set statement_timeout = 1000;
SET
pgaweb=# begin read only;
BEGIN
pgaweb=*# select pg_sleep(2);
ERROR:  canceling statement due to statement timeout
pgaweb=!# select pg_sleep(0.5);
ERROR:  current transaction is aborted, commands ignored until end of transaction block
```

Meaning this code doesn't quite do what you want it to do. When the firstErr is not nil (for whatever the reason), this second query execution will always be ignored.

seanlinsley: Good catch. Maybe the query text should contain `BEGIN READ ONLY` and `COMMIT` so each query is in its own transaction.

keiko713 (approved):
Nice, confirmed that it's reporting the correct error message 👍

PR #646, opened 2024-12-02
Update pg_query to v6 / Postgres 17 parser

seanlinsley:


lfittl (approved):
Looks good - I'm not sure why we're seeing the changes in the AWS SDK here, but since they are only removing unused files, I assume this occurred due to a toolchain improvement (similar to how `go.sum` now lists a few extra checksums with this change).

PR #648, opened 2024-12-03
Azure Metrics: fix getFirstMetricValue to return the first value

keiko713:
Spotted by Maciek in https://github.com/pganalyze/collector/pull/591/files#r1866387023

This function was previously returning the last value instead of the first value. This fix changes it and returns the value as soon as reading it.

Tested with the staging Azure database and it didn't break things.

### DIFF
--- a/input/system/azure/system.go
+++ b/input/system/azure/system.go
@@ -286,6 +286,7 @@ func getFirstMetricValue(metric *azquery.Metric) (metricValue *azquery.MetricVal
 	for _, timeSeriesElement := range metric.TimeSeries {
 		for _, mValue := range timeSeriesElement.Data {
 			metricValue = mValue
+			return
 		}
 	}
 	return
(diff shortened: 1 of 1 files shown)


input/system/azure/system.go
```diff
@@ -286,6 +286,7 @@ func getFirstMetricValue(metric *azquery.Metric) (metricValue *azquery.MetricVal
 	for _, timeSeriesElement := range metric.TimeSeries {
 		for _, mValue := range timeSeriesElement.Data {
 			metricValue = mValue
+			return
 		}
 	}
 	return
```
msakrejda, line 292: Why not
```suggestion
        if len(metric.TimeSeries) == 0 || len(metric.TimeSeries[0].timeSeriesElement.Data) == 0 {
                return
        }
        return metric.TimeSeries[0].timeSeriesElement.Data[0]
```
?

msakrejda (approved):
I find the loops a little puzzling: I'm wondering if I'm missing something and we can't just do an array index lookup. But either works.

input/system/azure/system.go
keiko713: I see, yeah I agree that it can be more straight forward. Just for the context, this loop was originally coming from "what we should be getting values" like this example:
https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/monitor/azquery/example_test.go#L261

Though, above example is to print out all data, hence the loop. Since we narrow down the usage here and only returning the first one, we can simplify this.

PR #651, opened 2024-12-05
Added ability to use raw envrionment variables maps in deployment

asaf400:
Hello, Internally we use external secret sources,
This PR adds the Ability for the environment variables to reference k8s objects that are per-defined under the pganalyze target namespace by cluster operators.

This solves an issue where a secret that does not conform to either existing methods:

- `extraEnv`: simple `key: value` strings
- `configMap` or `secret`: injection by using `envFrom` restricts the data fields conform to target environment variable names

By allowing helm operator to specify raw elements of the `env` spec of a container, it allows for additional options that are not currently covered.

I do not know how you do versioning for same appVersion patching of the chart-side only changes,
I see that most chart releases correspond to app releases, I ran `helm-docs` which is the only section in CONTRIBUTING.md that seems relevant for this type of PR

Edits: typo, versioning note

### DIFF
--- a/contrib/helm/pganalyze-collector/values.yaml
+++ b/contrib/helm/pganalyze-collector/values.yaml
@@ -28,7 +28,7 @@ extraEnvRaw: []
   #   valueFrom:
   #     secretKeyRef:
   #       name: pganalyze
-  #       key: api-key        
+  #       key: api-key
 
 configMap:
   # -- Specifies whether a config map should be created.
--- a/contrib/helm/pganalyze-collector/values.yaml
+++ b/contrib/helm/pganalyze-collector/values.yaml
@@ -21,6 +21,15 @@ extraEnv: {}
 #   DB_ALL_NAMES: true (monitor all databases on the server)
 #   DB_USERNAME: your_monitoring_user
 
+# -- Environment variables to be passed to the container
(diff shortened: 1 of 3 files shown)


contrib/helm/pganalyze-collector/values.yaml
```diff
@@ -21,6 +21,15 @@ extraEnv: {}
 #   DB_ALL_NAMES: true (monitor all databases on the server)
 #   DB_USERNAME: your_monitoring_user
 
+# -- Environment variables to be passed to the container
+# Config settings can be defined in raw form, for use with externally maintained env value sources (configMapKeyRef, fieldRef, resourceFieldRef, secretKeyRef)
+extraEnvRaw: []
```
keiko713, line 26: Just a memo, I can see that this pattern and the name `extraEnvRaw` is used in the other repos too, so this looks good 👍 

(e.g. https://github.com/apache/superset/commit/b35645c3f4db8526c12ef88d9056a348fcc1f249)

contrib/helm/pganalyze-collector/values.yaml
```diff
@@ -21,6 +21,15 @@ extraEnv: {}
...
+  # - name: PGA_API_KEY
+  #   valueFrom:
+  #     secretKeyRef:
+  #       name: pganalyze
+  #       key: api-key
```
keiko713, line 31:
```suggestion
  #       key: api-key
```

looks like there is trailing spaces here

PR #655, opened 2024-12-19
Implement new pganalyze.explain_analyze() helper

lfittl:
This executes on-demand query runs for the Query Tuning feature via
a helper function that prevents reading the table data directly, as well
as granting the collector user unnecessary permissions.

This is now the only way that query runs can execute EXPLAIN ANALYZE,
direct execution without the helper is intentionally not supported.


input/postgres/explain_analyze_test.go
```diff
@@ -0,0 +1,387 @@
...
+
+	for _, pair := range queryRunTests {
+		var errStr string
+		output, err := postgres.RunExplainAnalyzeForQueryRun(context.Background(), db, pair.query, pair.params, pair.paramTypes, "", true)
+		if err != nil {
+			errStr = fmt.Sprintf("%s", err)
+		}
+
+		// Avoid differences in test runs by masking total planning/execution time stats
+		re := regexp.MustCompile(`("(Planning|Execution) Time":) [\d.]+`)
```
msakrejda, line 259: I like that we're testing this, but the `testMode` argument is pretty gross. If I change the regexp to
```suggestion
               re := regexp.MustCompile(`("(?:Planning Time|Execution Time|Startup Cost|Total Cost|Actual Startup Time|Actual Total Time)":) [\d.]+`)
```
I get what looks like fairly stable output (there's buffer hit info, but it's all zeros, and I think given the trivial queries, we can assume it will stay that way). Thoughts?

lfittl: Yeah, you're right - will revise.

util/explain_analyze_helper.sql
```diff
@@ -0,0 +1,46 @@
...
+  param_types_str text;
+  explain_prefix text;
+  explain_flag text;
+  result text;
+BEGIN
+  SET TRANSACTION READ ONLY;
+
+  PERFORM 1 FROM pg_roles WHERE (rolname = current_user AND rolsuper) OR (pg_has_role(oid, 'MEMBER') AND rolname IN ('rds_superuser', 'azure_pg_admin', 'cloudsqlsuperuser'));
+  IF FOUND THEN
+    RAISE EXCEPTION 'cannot run: EXPLAIN ANALYZE helper is owned by superuser - recreate function with lesser privileged user';
```
lfittl, line 14: Reviewing this error message in-app, I wonder if for consistency we instead say here and in the other messages:

```suggestion
    RAISE EXCEPTION 'cannot run: pganalyze.explain_analyze(...) function is owned by superuser - recreate function with lesser privileged user';
```

Before:
Screenshot 2024-12-18 at 11 24 41 PM


After:
Screenshot 2024-12-18 at 11 25 08 PM

keiko713:
I think overall this is looking good (I admit that I haven't checked the SQL part).
Are we going to retire the `enable_query_runner` flag in the different PR? Also, we will (well, the app will) want to know if the helper function exists or not, so it would be great if it's reported as a collector config.

keiko713:
> The main reason to stay with that existing flow that I could see (vs also including it in collector info), is that this is a per-database helper, like the Index Advisor helpers. Storing this information in collector info for many databases might not be ideal?

Ah, interesting, so you'll need to create a helper function to all databases (if you have many databases). Yeah in that case, sending out in the collector info doesn't make sense.
In that case, we may want to consider adding the function to https://github.com/pganalyze/collector/blob/main/runner/generate_helper_sql.go or providing some similar things for this.

I think I can workaround by checking the function definition on the app side 👍

util/explain_analyze_helper.sql
seanlinsley: Is the `(...)` really necessary? I'd tend to think that saying "the pganalyze.explain_analyze function" is plenty clear. Not a big deal, just trying to reduce character count.

lfittl: Yeah, fair point, especially since we say function right after.

.github/workflows/ci.yml
```diff
@@ -13,6 +13,15 @@ jobs:
   build:
     # Use new enough OS that has CGroupsv2 enabled (required by the integration tests)
     runs-on: ubuntu-22.04
+    env:
+      TEST_DATABASE_URL: postgresql://postgres:postgres@localhost:5432/postgres?sslmode=disable
+    services:
+      postgres:
+        image: postgres:14
```
lfittl, line 20: Using 14 as a minimum here since the EXPLAIN ANALYZE tests rely on being able to turn off `compute_query_id` in the output (and that was added in 14).

lfittl:
@keiko713 @seanlinsley FYI, this is ready for re-review. I ended up adding both parameter support and Postgres setting support here as well, since it was quick enough, and even if the server doesn't use it that works fine.

input/postgres/explain_analyze.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	// Warm up caches without collecting timing info (slightly faster)
+	_, err = runExplainAnalyzeWithHelper(ctx, db, query, parameters, parameterTypes, []string{"ANALYZE", "TIMING OFF"}, marker)
+
+	if err != nil && !strings.Contains(err.Error(), "statement timeout") {
+		// Run again if it was a timeout error, to make sure we got the caches warmed up all the way
+		_, err = runExplainAnalyzeWithHelper(ctx, db, query, parameters, parameterTypes, []string{"ANALYZE", "TIMING OFF"}, marker)
+		if ctx.Err() != nil {
+			return "", err
```
seanlinsley, line 54: Why is this checking `ctx.Err()` instead of `err`, like all the other cases?

input/postgres/explain_analyze.go
```diff
@@ -0,0 +1,78 @@
...
+		err = fmt.Errorf("query is not permitted to run (multi-statement or utility command?)")
+		return err
+	}
+
+	// TODO: Consider adding additional checks here (e.g. blocking known bad function calls)
+
+	return nil
+}
+
+func RunExplainAnalyzeForQueryRun(ctx context.Context, db *sql.DB, query string, parameters []null.String, parameterTypes []string, marker string) (_result string, _err error) {
```
seanlinsley, line 41: If the underscores on the return types were removed, and `explainOutput` was renamed to `result`, the code could be simplified to just do `return` instead of `return "", err`. Go will automatically return the defined variables for you.

main.go
```diff
@@ -73,7 +74,9 @@ func main() {
 	flag.BoolVar(&testRunLogs, "test-logs", false, "Tests whether log collection works (does not test privilege dropping for local log collection, use --test for that)")
 	flag.BoolVar(&testExplain, "test-explain", false, "Tests whether EXPLAIN collection works by issuing a dummy query (ensure log collection works first)")
 	flag.StringVar(&testSection, "test-section", "", "Tests a particular section of the config file, i.e. a specific server, and ignores all other config sections")
-	flag.StringVar(&generateStatsHelperSql, "generate-stats-helper-sql", "", "Generates a SQL script for the given server (name of section in the config file, or \"default\" for env variables), that can be run with \"psql -f\" for installing the collector stats helpers on all configured databases")
+	flag.StringVar(&generateHelperSql, "generate-stats-helper-sql", "", "Deprecated alias for --generate-helper-sql")
```
seanlinsley, line 77: Do we really need to deprecate this? Why not remove it?

util/explain_analyze_helper.sql
```diff
@@ -0,0 +1,46 @@
...
+  SET TRANSACTION READ ONLY;
+
+  PERFORM 1 FROM pg_roles WHERE (rolname = current_user AND rolsuper) OR (pg_has_role(oid, 'MEMBER') AND rolname IN ('rds_superuser', 'azure_pg_admin', 'cloudsqlsuperuser'));
+  IF FOUND THEN
+    RAISE EXCEPTION 'cannot run: pganalyze.explain_analyze helper function is owned by superuser - recreate function with lesser privileged user';
+  END IF;
+
+  SELECT pg_catalog.regexp_replace(query, ';+\s*\Z', '') INTO prepared_query;
+  IF prepared_query LIKE '%;%' THEN
+    RAISE EXCEPTION 'cannot run pganalyze.explain_analyze helper function when query contains semicolon';
```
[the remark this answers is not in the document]
seanlinsley: Also, how does this handle a single query that ends with a semicolon? I'd expect that to be allowed, since there isn't a statement following the semicolon.

runner/query_run.go
```diff
@@ -39,28 +39,63 @@ func SetupQueryRunnerForAllServers(ctx context.Context, servers []*state.Server,
...
 }
 
+func runQueryOnDatabase(ctx context.Context, server *state.Server, collectionOpts state.CollectionOpts, logger *util.Logger, id int64, query *state.QueryRun) (string, error) {
+	db, err := postgres.EstablishConnection(ctx, server, logger, collectionOpts, query.DatabaseName)
+	if err != nil {
+		return "", err
+	}
+	defer db.Close()
+
+	if postgres.StatsHelperExists(ctx, db, "explain_analyze") {
```
keiko713, line 49: This `runQueryOnDatabase` function is expected to be used for other types of QueryRun too, but this helper method check, as well as Postgres settings applying queries are used for any types of QueryRun.
I wonder if we should actually check the type in the beginning of this func and early return unless `query.Type == pganalyze_collector.QueryRunType_EXPLAIN` 🤔

main.go
keiko713: Somewhat related to that, can you explain a bit more around here?

Reading the code, looks like:
* by passing `--generate-helper-explain-analyze-role new-role-name` with `--generate-helper-sql`, create an explain analyze role with new-role-name + generate a explain analyze helper function

I think instead of renaming `generate-helper-sql`, generating the explain analyze role can be completely separate thing. IOW, keep `generate-stats-helper-sql` to generate stats related helpers, then introduce a new separate flag something like `generate-explain-analyze-helper-sql` which will generate the helper SQL + create roles.
In that way, the docs will be more straight forward and saying "run collector with `--generate-explain-analyze-helper-sql myrole` then apply the SQL". Just an idea though.

protobuf/server_message.proto
```diff
@@ -39,10 +39,18 @@ message ServerMessage {
...
+  }
+
   message QueryRun {
     int64 id = 1;
     QueryRunType type = 2;
     string database_name = 3;
     string query_text = 4;
+    repeated NullString query_parameters = 5;
+    repeated string query_parameter_types = 6;
+    repeated QueryRunPostgresSetting postgres_settings = 7;
```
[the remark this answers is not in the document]
seanlinsley: On the app side we're calling these planner settings, not Postgres settings. Are we planning to extend this beyond planner settings to the less-specific name would make sense in the future?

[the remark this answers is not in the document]
keiko713: Uhmm good point. I'm happy to also update the app side to be more generic, and use postgres settings. If we do so, we can potentially specify any other things like overwriting `statement_timeout`.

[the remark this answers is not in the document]
lfittl: Hmm hmm. I think the main purpose here is to use it for anything that could affect the planner, which notably could include things like `work_mem` (though that is 50/50, since it partially affects execution, partially planning, depending on the node type). I'd argue though that `statement_timeout` is best managed separately (i.e. explicitly), since it has broader implications.

I'm fine either way, but we should be consistent. I think its probably most logical if I rename this here to `planner_settings` for now (I'll do that tomorrow unless I hear reasons not to).

util/sql_helpers.go
```diff
@@ -0,0 +1,8 @@
+package util
+
+import (
+	_ "embed"
+)
+
+//go:embed explain_analyze_helper.sql
+var ExplainAnalyzeHelper string
```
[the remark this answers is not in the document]
lfittl: Yeah, I hadn't used it before but it felt like the right way to go here. We could even use this for the SQL we use in other places (so its split out from the files running the queries, and syntax highlighted).

input/postgres/explain_analyze.go
lfittl: Since its only trying to exit early if the context was canceled (e.g. collector shutting down, etc), otherwise we want to retry and ignore the error.

util/explain_analyze_helper.sql
```diff
@@ -0,0 +1,46 @@
...
+    EXECUTE explain_prefix || 'EXECUTE pganalyze_explain_analyze' || params_str INTO STRICT result;
+  EXCEPTION WHEN OTHERS THEN
+    DEALLOCATE pganalyze_explain_analyze;
+    RAISE;
+  END;
+  DEALLOCATE pganalyze_explain_analyze;
+
+  RETURN result;
+END
+$$ LANGUAGE plpgsql VOLATILE SECURITY DEFINER;
```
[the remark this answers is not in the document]
lfittl: Not intentional, just a copy and paste issue I think.

protobuf/server_message.proto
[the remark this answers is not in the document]
seanlinsley: A reason to stick with `postgres_settings` is to have less code when in the future query runs can support more maintenance tasks (vacuum, reindex, create index, drop index), where it might make sense to set things like `work_mem`.

util/explain_analyze_helper.sql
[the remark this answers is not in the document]
lfittl: The single query ending in a semicolon is handled by the `regexp_replace` in the prior line.

I'd argue this check is essential - if we don't do it, the whole purpose of having a function to turn a read query into the EXPLAIN ANALYZE output only would be defeated, since you could just run arbitrary queries. Unless you have another idea of how to protect against this at the SQL level? (as mentioned previously, we really do need multiple layers of defense, and a function is much easier to audit than all the collector's Go code).

protobuf/server_message.proto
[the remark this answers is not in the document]
lfittl: Yeah, that's fair - the data structure here can be a bit more agnostic, and you're right that we could have utility commands that could benefit from having certain settings set. In that light I'm also okay with keeping this as-is. 

@keiko713 thoughts?

main.go
lfittl: Yeah, that's a good point, let me revise it to just make it a completely separate helper - the intent here is different, and so maybe best to keep it split in two (and avoid the deprecation too).

protobuf/server_message.proto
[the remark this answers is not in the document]
keiko713: Yeah, I agree that `postgres_settings` is a better naming here.
I think we can still debate a bit about if we should also update the wording we use in the app side or not (e.g. query run will use postgres_settings, but maybe explain_query can still use planner_settings as a column name, then generate postgres_settings based on planner_settings).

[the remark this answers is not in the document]
seanlinsley: Unresolving this conversation: it looks like the [first comment](https://github.com/pganalyze/collector/pull/655#discussion_r1896868627) got missed

input/postgres/explain_analyze.go
lfittl: Actually, I realized this is unnecessary since we return early on non-timeout errors anyway - will drop this.

util/explain_analyze_helper.sql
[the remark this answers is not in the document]
lfittl: Regarding the earlier feedback, I removed the function name from the messages (since you're right, it would be contained in the CONTEXT as part of error messages), and also reworded the semicolon error to instead refer to multi-statement queries.

lfittl:
@seanlinsley @keiko713 This is ready for re-review, all feedback should be addressed.

runner/query_run.go
```diff
@@ -39,85 +39,74 @@ func SetupQueryRunnerForAllServers(ctx context.Context, servers []*state.Server,
 	}
 }
 
+func runQueryOnDatabase(ctx context.Context, server *state.Server, collectionOpts state.CollectionOpts, logger *util.Logger, id int64, query *state.QueryRun) (string, error) {
```
seanlinsley, line 42: Nitpick, but I find it confusing to use C-style function ordering where the entrypoints to a file are at the bottom of a file. Go doesn't impose function ordering requirements like C, so I would expect this function to be below the top-level `run` function

runner/run.go
```diff
@@ -130,6 +130,34 @@ func Run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
 		return
 	}
 
+	if globalCollectionOpts.GenerateExplainAnalyzeHelperSql != "" {
+		wg.Add(1)
+		testRunSuccess = make(chan bool)
```
seanlinsley, line 135: It doesn't have to be part of this PR, but we may want to add test run output for the presence of the helper function and success on running a test explain.

util/explain_analyze_helper.sql
```diff
@@ -0,0 +1,46 @@
...
+  explain_prefix text;
+  explain_flag text;
+  result text;
+BEGIN
+  SET TRANSACTION READ ONLY;
+
+  PERFORM 1 FROM pg_roles WHERE (rolname = current_user AND rolsuper) OR (pg_has_role(oid, 'MEMBER') AND rolname IN ('rds_superuser', 'azure_pg_admin', 'cloudsqlsuperuser'));
+  IF FOUND THEN
+    RAISE EXCEPTION 'cannot run: helper is owned by superuser - recreate function with lesser privileged user';
+  END IF;
```
seanlinsley, line 15: This error occurs in local testing because my Postgres user is a superuser. Maybe the OR should be an AND instead?

For reference, these are the values for the different where clauses:
```sql
select rolsuper, pg_has_role(oid, 'MEMBER'), rolname IN ('rds_superuser', 'azure_pg_admin', 'cloudsqlsuperuser') from pg_roles where rolname = current_user;
 rolsuper | pg_has_role | ?column? 
----------+-------------+----------
 t        | t           | f

File: runner/query_run.go [hunk 15]
```diff
@@ -39,85 +39,74 @@ func SetupQueryRunnerForAllServers(ctx context.Context, servers []*state.Server,
...
+	err = db.QueryRow(postgres.QueryMarkerSQL + "SELECT pg_backend_pid()").Scan(&pid)
+	if err != nil {
+		return "", err
+	}
+	server.QueryRunsMutex.Lock()
+	server.QueryRuns[id].BackendPid = pid
+	server.QueryRunsMutex.Unlock()
+
+	for name, value := range query.PostgresSettings {
+		_, err = db.ExecContext(ctx, postgres.QueryMarkerSQL+"SET %s = %s", pq.QuoteIdentifier(name), pq.QuoteLiteral(value))
```
seanlinsley, line 75: When testing this locally with `statement_timeout` set to `1s`, this fails with `pq: syntax error at or near "%"`. Looks like a `fmt.Sprintf` call is needed.

util/explain_analyze_helper.sql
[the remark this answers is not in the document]
seanlinsley: After testing this locally, it looks like the `CONTEXT` doesn't make it to the server as part of the query run's error string. So maybe the function name should be added back?

runner/query_run.go
lfittl: Fair point, I can reorder.

runner/run.go
lfittl: Yeah, that's a good point. I'll add this in a follow-up PR.

util/explain_analyze_helper.sql
lfittl: If you follow the instructions on setting up the separate user ([preview](https://pganalyze-we-query-tuni-nl5vxq.herokuapp.com/docs/query-tuning/collector-workflow#creating-the-pganalyze_explain-role)) it should work as expected, even in a local setup.

runner/websocket.go
```diff
@@ -142,11 +143,22 @@ func connect(ctx context.Context, server *state.Server, globalCollectionOpts sta
...
-						Type:         q.Type,
-						DatabaseName: q.DatabaseName,
-						QueryText:    q.QueryText,
+						Id:                  q.Id,
+						Type:                q.Type,
+						DatabaseName:        q.DatabaseName,
+						QueryText:           q.QueryText,
+						QueryParameters:     parameters,
+						QueryParameterTypes: q.QueryParameterTypes,
+						PostgresSettings:    settings,
```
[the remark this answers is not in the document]
lfittl: Excellent point!

util/explain_analyze_helper.sql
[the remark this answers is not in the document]
lfittl: Ah, yeah, you're right. I'll add it back for now to make it clearer for the user.

runner/query_run.go
lfittl: Oh yeah, strange, I must have dropped that in a refactoring by accident.

lfittl: This now works as expected for your test case (that you had in the other PR).

That said, I do wonder if we should explicitly block changing `statement_timeout` through the parameters in the query run, or at least have some higher limit that's a hard limit? (i.e. can't be changed by the server side)

Maybe something like 1 minute as the default limit, can be raised to up to 10 minutes by the server side, but nothing higher?

Assuming we'd want to do that in the future, its likely easier to enforce by passing it as a separate value (not a free form parameter, i.e block it in `parameters`), so we don't have to deal with `20min` vs `20s` parameter value parsing.

seanlinsley: Agreed it makes sense to block it here and make it a standalone setting so we can validate it. But applying a max value is going to depend on the type of query run that's being performed; if in the future we're doing a concurrent reindex it should be fine for that to run for an arbitrary amount of time.

lfittl: Yeah, agreed. I'll filter it out here for now and we can add customizing it in a separate PR.

seanlinsley: Actually on second thought a separate field isn't really needed: we can validate the value in `postgres_settings` since the key value is always going to be named `statement_timeout`.

For now though, we can bypass this issue by making the statement timeout non-configurable. That can be done by moving the existing `postgres.SetStatementTimeout` call after this code, so the hardcoded 1 minute timeout is always used.

lfittl: > Actually on second thought a separate field isn't really needed: we can validate the value in `postgres_settings` since the map key is always going to be named `statement_timeout`.

The problem is not the key, but the value - since you can use different duration units (seconds vs minutes), we'd have to parse them in order to validate the value.

> For now though, we can bypass this issue by making the statement timeout non-configurable. That can be done by moving the existing postgres.SetStatementTimeout call after this code, so the hardcoded 1 minute timeout is always used.

Sure, I'll move it down for now so it takes effect after the other settings.

seanlinsley:
This looks good overall, but I did run into one bug: the statement timeout doesn't seem to take effect. For example if I run a `SELECT pg_sleep(90)` query through the system, the collector responds back with JSON including `"Execution Time": 90001.987`.

lfittl:
> This looks good overall, but I did run into one bug: the statement timeout doesn't seem to take effect. For example if I run a `SELECT pg_sleep(90)` query through the system, the collector responds back with JSON including `"Execution Time": 90001.987`.

Strange - I had tested that and it worked for me, but I can recheck tomorrow. It sounds like you got an actual result vs the EXPLAIN without ANALYZE that should happen when the query times out.

runner/generate_helper_sql.go
```diff
@@ -68,3 +68,36 @@ func GenerateStatsHelperSql(ctx context.Context, server *state.Server, globalCol
...
+	databases, _, err := postgres.GetDatabases(ctx, logger, db, version)
+	if err != nil {
+		return "", fmt.Errorf("error collecting pg_databases: %s", err)
+	}
+
+	output := strings.Builder{}
+	for _, dbName := range postgres.GetDatabasesToCollect(server, databases) {
+		output.WriteString(fmt.Sprintf("\\c %s\n", pq.QuoteIdentifier(dbName)))
+		output.WriteString("CREATE SCHEMA IF NOT EXISTS pganalyze;\n")
+		output.WriteString(fmt.Sprintf("GRANT USAGE ON SCHEMA pganalyze TO %s;\n", server.Config.GetDbUsername()))
```
msakrejda, line 93: Aren't we connecting as `server.Config.GetDbUsername()`? Should this be `globalCollectionOpts.GenerateExplainAnalyzeHelperRole`?

msakrejda: Also, technically users may select a helper role name that needs quoting as well, right? But it's probably fine not to support that.

msakrejda (commented):
I don't follow the username usage in the helper generation script. Otherwise, looks good.

runner/generate_helper_sql.go
lfittl: > Aren't we connecting as `server.Config.GetDbUsername()`? Should this be `globalCollectionOpts.GenerateExplainAnalyzeHelperRole`?

This `GRANT USAGE` is not for the creation step, but rather so that the regular pganalyze monitoring user can use the function later - so using `server.Config.GetDbUsername()` is correct here.

Note that the generate logic doesn't connect itself, but emits a psql script that you execute as a superuser - so this does not refer to the user that the statement runs as.

lfittl: > Also, technically users may select a helper role name that needs quoting as well, right? But it's probably fine not to support that.

Good point - I can wrap this, and the other uses of the role name in `pq.QuoteIdentifier`.

input/postgres/explain_analyze.go
```diff
@@ -0,0 +1,66 @@
...
+
+		// If it timed out again, capture a non-ANALYZE EXPLAIN instead
+		if err != nil && strings.Contains(err.Error(), "statement timeout") {
+			return runExplainAnalyze(ctx, db, query, parameters, parameterTypes, []string{}, marker)
+		} else if err != nil {
+			return
+		}
+	} else if err != nil {
+		return
+	}
```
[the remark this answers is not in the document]
lfittl: Yeah, I like it - thanks for proposing that. I took your code structure but moved the comments around a bit (so they sit here right before the actions they're describing).

runner/generate_helper_sql.go
msakrejda: Ah, yes, okay. That makes sense.

msakrejda (approved):

PR #656, opened 2024-12-19
Track parent partition size when child partitions are untracked

### LINKED ISSUE #654: 0.63.0 broke size reporting for TimescaleDB hypertables
It seems the changes in:

https://github.com/pganalyze/collector/commit/550439b020db7c8e634d9f3486ec68696b0c072a#diff-89c88e46957c7be3a80002749ee13e4611914aff1acf0eb5e27d05b119e3b98aR34

broke size reporting for TimescaleDB hypertables.

We're using TimescaleDB extensively and after an update of one of our clusters we noticed that size was no longer reported correctly for hypertables.

I guess for some reason pganalyze-collector does not find the partitions of these tables?

seanlinsley:
#635 improved the fidelity of statistics tracked for parent partitions, but in moving the code from SQL to Go it stopped reporting the size of parent partitions when `ignore_schema_regexp` is set to ignore the child partitions. This PR fixes #654 by bringing back the original SQL, but only applying it to child partitions that match `ignore_schema_regexp`.

### DIFF
--- a/input/postgres/relation_stats.go
+++ b/input/postgres/relation_stats.go
@@ -20,7 +20,16 @@ locked_relids_with_parents AS (
 	UNION SELECT relid FROM locked_relids
 )
 SELECT c.oid,
-			 COALESCE(pg_catalog.pg_table_size(c.oid), 0) AS size_bytes,
+			 COALESCE(pg_catalog.pg_table_size(c.oid), 0) +
+			 COALESCE((
+				SELECT pg_catalog.sum(pg_catalog.pg_table_size(inhrelid))
+				FROM pg_catalog.pg_inherits
+				JOIN pg_class t ON inhrelid = t.oid
+				WHERE inhparent = c.oid
+					-- Only include sizes from child partitions skipped by ignore_schema_regexp.
+					-- Child partitions tracked by the collector have their sizes added by mergePartitionSizes.
+					AND ($1 != '' AND (n.nspname || '.' || t.relname) ~* $1)
+			 ), 0) AS size_bytes,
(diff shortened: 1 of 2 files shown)


input/postgres/relation_stats.go
```diff
@@ -20,7 +20,16 @@ locked_relids_with_parents AS (
...
+			 COALESCE(pg_catalog.pg_table_size(c.oid), 0) +
+			 COALESCE((
+				SELECT pg_catalog.sum(pg_catalog.pg_table_size(inhrelid))
+				FROM pg_catalog.pg_inherits
+				JOIN pg_class t ON inhrelid = t.oid
+				WHERE inhparent = c.oid
+					-- Only include sizes from child partitions skipped by ignore_schema_regexp.
+					-- Child partitions tracked by the collector have their sizes added by mergePartitionSizes.
+					AND ($1 != '' AND (n.nspname || '.' || t.relname) ~* $1)
+			 ), 0) AS size_bytes,
```
seanlinsley, line 32: The screenshot below is from local testing. 11:40 shows the detailed stats from #635. I then updated the collector config with `ignore_schema_regexp = 35d_`, and then 12:10 shows the updated behavior with this PR. Ignore the in-between timestamps where I was fixing bugs.

Screenshot 2024-12-19 at 12 12 08 PM

output/transform/merge_partition_sizes.go
```diff
@@ -5,6 +5,12 @@ import (
 	"github.com/pganalyze/collector/state"
 )
 
+// Since Postgres reports parent partition tables as being zero-sized,
+// this backfills stats for the parent table as a summation of all child tables.
+//
+// When ignore_schema_regexp is set, GetRelationStats bypasses this for tables not tracked by the collector.
+//
+// TODO: recursively build up stats when nested partitioning is used
```
seanlinsley, line 13: This TODO is an existing issue with the code.

msakrejda (approved):

PR #660, opened 2025-01-07
Query runner: Block known problematic functions if found in parse tree

lfittl:
This acts as an additional defense in case the user does not revoke permissions from non-superuser for dblink as we recommend.


input/postgres/explain_analyze.go
```diff
@@ -53,16 +58,54 @@ func runExplainAnalyze(ctx context.Context, db *sql.DB, query string, parameters
...
-	if err != nil || len(isUtil) != 1 || isUtil[0] {
-		err = fmt.Errorf("query is not permitted to run (multi-statement or utility command?)")
+	parseResult, err := pg_query.Parse(query)
+	if err != nil {
+		return fmt.Errorf("query is not permitted to run - failed to parse")
+	}
+	if len(parseResult.Stmts) != 1 {
+		return fmt.Errorf("query is not permitted to run - multi-statement query string")
+	}
+	for _, rawStmt := range parseResult.Stmts {
```
keiko713, line 68: I got a bit confused by this range (it almost looks like it's doing a tree walking while it's not doing).
Maybe better to do `rawStmt := parseResult.Stmts[0]`?

lfittl: Good idea, I made that change, and clarified the treewalk code so its easier to understand.

input/postgres/explain_analyze.go
```diff
@@ -53,16 +58,72 @@ func runExplainAnalyze(ctx context.Context, db *sql.DB, query string, parameters
...
 	return nil
 }
+
+var blockedFunctions = []string{
+	// Blocked because these functions allow exfiltrating data to external servers
+	"dblink",
+	"dblink_connect",
+	"dblink_exec",
+	// Blocked because these functions allow executing arbitrary SQL as input (which can workaround other checks)
+	"xpath_table",
```
msakrejda, line 94: Is [tablefunc](https://www.postgresql.org/docs/current/tablefunc.html) something we want to block?

msakrejda (approved):

input/postgres/explain_analyze.go
```diff
@@ -53,16 +58,72 @@ func runExplainAnalyze(ctx context.Context, db *sql.DB, query string, parameters
...
+	if len(parseResult.Stmts) != 1 {
+		return fmt.Errorf("query is not permitted to run - multi-statement query string")
 	}
 
-	// TODO: Consider adding additional checks here (e.g. blocking known bad function calls)
+	stmt := parseResult.Stmts[0].Stmt.Node
+	switch stmt.(type) {
+	case *pg_query.Node_SelectStmt:
+		// Allowed, continue
+		// Note that we permit wCTEs here (for now), and instead rely on the read-only transaction to block them
```
keiko713, line 73: Talked a bit offline too, but we could potentially do these checks in the below treewalk so that we can also check wCTEs. I think it's fine as is for now, but we could come back later potentially.

keiko713 (approved):
Thanks for improving readability with treewalker 👍

input/postgres/explain_analyze.go
lfittl: Yeah, lets revise that at a later point, but agreed we could adjust this to match the check we do on the server side for wCTEs.

PR #663, opened 2025-01-08
Fix DB connection leak in buffercache logic, other connection related cleanup

lfittl:
See individual commits.

### DIFF
--- a/input/postgres/helpers.go
+++ b/input/postgres/helpers.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/guregu/null"
 	"github.com/pganalyze/collector/state"
-	"github.com/pganalyze/collector/util"
 )
 
 func unpackPostgresInt32Array(input null.String) (result []int32) {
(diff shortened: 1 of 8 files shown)


input/postgres/helpers.go
```diff
@@ -72,16 +71,10 @@ SELECT setting
 	FROM pg_settings
  WHERE name = '%s'`
 
-func GetPostgresSetting(ctx context.Context, settingName string, server *state.Server, globalCollectionOpts state.CollectionOpts, prefixedLogger *util.Logger) (string, error) {
+func GetPostgresSetting(ctx context.Context, db *sql.DB, settingName string) (string, error) {
```
keiko713, line 74: This looks cleaner, but I wonder why we weren't doing this previously 🤔

keiko713 (approved):
Looks good 👍 Thanks for fixing!

input/postgres/helpers.go
lfittl: Looks like it started out being used in a context where we didn't have a pre-existing connection (`runner/logs.go` in https://github.com/pganalyze/collector/commit/cbf2f087d6ee12fc3b3eefdd88ebd6402d97264c), and we later used the same helper in other places, without refactoring it.

PR #666, opened 2025-01-23
Start collecting and sending pg_stat_statements_info stats

keiko713:
Often it is helpful to learn about the information of pg_stat_statements, to help debugging some collector OOM issue, or understanding how accurately we are collecting stats of statements.
With this PR, we start reporting the deallocated count info and last reset time info using `pg_stat_statements_info` view, available from Postgres 14 (pg_stat_statements 1.9).
https://www.postgresql.org/docs/current/pgstatstatements.html#PGSTATSTATEMENTS-PG-STAT-STATEMENTS-INFO

This will be reported as a part of the server statistic, and planning to be persisted every 10 mins info.
It might be too much to have this data for every 10 mins, but it's also not a large data + only per server, so I'd say that it's acceptable.

Here is how the protobuf looks like (from `--dry-run`):

```
	"serverStatistic": {
		"currentXactId": "3172097",
		"nextMultiXactId": "1",
		"xminHorizonBackend": "3172097",
		"pgStatStatementsDealloc": "660",
		"pgStatStatementsReset": {
			"valid": true,
			"value": "2024-11-29T10:49:02.044562Z"
		}
	},
```


output/transform/postgres.go
```diff
@@ -164,6 +164,8 @@ func transformPostgresServerStats(s snapshot.FullSnapshot, transientState state.
 		XminHorizonReplicationSlotCatalog: transientState.ServerStats.FullXminHorizonReplicationSlotCatalog(),
 		XminHorizonPreparedXact:           transientState.ServerStats.FullXminHorizonPreparedXact(),
 		XminHorizonStandby:                transientState.ServerStats.FullXminHorizonStandby(),
+		PgStatStatementsDealloc:           transientState.ServerStats.PgStatStatementsDealloc,
```
lfittl, line 167: I think we should run our usual diffing logic for the `dealloc` field, so we don't send the absolute value, but rather a diff to the previous one - that may seem unnecessarily complex to implement, but I think its necessary to be able to answer "did any deallocations occur in this timeframe" (e.g. when investigating missing query texts for workloads at a given time)

keiko713: Ah, totally forgot about doing the diff-ing, will do 👍

keiko713: This is addressed, now it's sending a diff-ed version

msakrejda (approved):

PR #667, opened 2025-01-29
Sign built Go binaries for macOS

keiko713:
Due to security enhancements in newer macOS versions, unsigned Go binaries may hang when built and executed locally. To prevent this, add code signing to `make build`.


msakrejda (approved):
Can't test this here, but looks reasonable to me.

PR #668, opened 2025-01-30
Add --force-record mode for testing snapshots

keiko713:
Unsure this is a good solution, but it is a solution, so let's start discussion from here. Any feedback from how we should implement this to the name of the flag is welcome.

During the development, we have a problem that we can only send and process full snapshots every 10 minutes. This is problematic when we want to have a better feedback loop, especially implementing new stats to record and trying out the integration test locally.

With this change, we can run the test like `./pganalyze-collector --test --force-record`, then snapshots created for the test will send out like a normal one, will be processed on the server side like a normal one.
Currently, with all test mode, we add a special header to the websocket request, and the server side will simply disregard these snapshots, therefore we can't use `--test` for the integration test.

### DIFF
--- a/main.go
+++ b/main.go
@@ -66,6 +66,7 @@ func main() {
 	var reloadRun bool
 	var noReload bool
 	var benchmark bool
+	var forceRecord bool
 
 	logFlags := log.LstdFlags
 	logger := &util.Logger{}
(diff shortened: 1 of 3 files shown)


keiko713:
We discussed in the meeting and decided to address this on the server side, closing this.

Issue #348, opened 2022-12-18
Documentation/support for Google Cloud SQL Auth Proxy or Connector


jawnsy:
**Summary**

Support the Cloud SQL Auth Proxy or Connector.

**Background**

The [current setup instructions for Cloud SQL](https://pganalyze.com/docs/install/google_cloud_sql/01_create_monitoring_user) uses password authentication:

```sql
CREATE USER pganalyze WITH PASSWORD 'waJzOTWiQQjCZLr4' CONNECTION LIMIT 5;
```

Cloud SQL instances that use the (somewhat misleadingly-named) [Require SSL setting](https://cloud.google.com/sql/docs/postgres/configure-ssl-instance) do not support password authentication; instead, mutual TLS with client certificates must be used instead (from the referenced page):

> Enforcing SSL/TLS encryption as described below will make mutual TLS required for all clients
 
The CIS Benchmark for Google Cloud Platform, version 1.3.0, also recommends enabling this flag, so anyone running compliance scans will see this as a finding.

For anyone using Require SSL, it is often convenient to use the [Cloud SQL Auth Proxy](https://cloud.google.com/sql/docs/postgres/sql-proxy), which handles generating and rotating ephemeral certificates using an IAM identity. This can run as a sidecar and is compatible with `pganalyze-collector` as it exists today. Additionally, enabling the [Connector Enforcement feature](https://cloud.google.com/sql/docs/postgres/admin-api/rest/v1beta4/instances#connectorenforcement) _requires_ connections to the database to use the proxy or connector library.

While it is still possible to use password authentication with the auth proxy, it is less convenient than using IAM authentication, and IAM permissions are necessary anyways to issue the ephemeral certificate and establish a connection in the first place.

For customers running [Cloud SQL with IAM Authentication](https://cloud.google.com/sql/docs/postgres/authentication), and running the pganalyze-collector using [Cloud Run](https://cloud.google.com/sql/docs/postgres/connect-run) or [GKE with Workload Identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity), it is convenient to use IAM authentication, since it means that passwords will not need periodic rotation, and it also provides a stronger form of workload identity.

**Approach 1: Cloud SQL Auth Proxy**

[...]

PR #671, opened 2025-02-06
Support IAM Authentication with Google Cloud SQL

keiko713:
For https://github.com/pganalyze/collector/issues/348, using approach 2: Cloud SQL Connector.
This is on top of bumping Go version PR, as the latest cloud-sql-go-connector library drops the support of 1.21.

This introduce a new config variable `GCP_REGION`, as it's needed to make a host name.

Here is a docs for how users can use this:

* creating a monitoring service account: https://pganalyze-we-update-doc-4smwxx.herokuapp.com/docs/install/google_cloud_sql/01_create_monitoring_user#create-monitoring-user
* setting up with VM: https://pganalyze-we-update-doc-4smwxx.herokuapp.com/docs/install/google_cloud_sql/02_install_the_collector
* configuring the collector: https://pganalyze-we-update-doc-4smwxx.herokuapp.com/docs/install/google_cloud_sql/03_configure_the_collector_package#using-iam-authentication

docs PR: https://github.com/pganalyze/pganalyze-docs/pull/305


input/postgres/establish_connection.go
```diff
@@ -42,36 +43,49 @@ func EstablishConnection(ctx context.Context, server *state.Server, logger *util
...
+				return nil, err
+			} else {
+				dbPasswordOverride = dbToken
+			}
+		} else if config.SystemType == "google_cloudsql" {
+			if config.GcpProjectID == "" || config.GcpRegion == "" || config.GcpCloudSQLInstanceID == "" {
+				return nil, fmt.Errorf("To use IAM auth with Google Cloud SQL, you must specify project ID, region, and instance ID")
+			}
+			hostOverride = strings.Join([]string{config.GcpProjectID, config.GcpRegion, config.GcpCloudSQLInstanceID}, ":")
+			// When using cloud-sql-go-connector, this needs to be set as disable
```
keiko713, line 72: This is a good write-up for why this needs to be disabled.
I can also add the link as a comment too if that helps:
https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/issues/889

runner/run.go
```diff
@@ -42,6 +45,12 @@ func Run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
 		return
 	}
 
+	driverCleanup, err = pgxv5.RegisterDriver("cloudsql-postgres", cloudsqlconn.WithIAMAuthN())
```
keiko713, line 48: I need to return this driverCleanup to the main.go, since the cleanup should happen over there. Also, registering driver needs to happen only one time per the process, therefore it lives here.

https://github.com/GoogleCloudPlatform/cloud-sql-go-connector?tab=readme-ov-file#using-the-dialer-with-databasesql

runner/run.go
```diff
@@ -55,6 +59,14 @@ func Run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
 				logger.PrintError("Failed to initialize OpenTelemetry tracing provider, disabling exports: %s", err)
 			}
 		}
+
+		if cfg.DbUseIamAuth && cfg.SystemType == "google_cloudsql" && driverCleanup == nil {
+			driverCleanup, err = pgxv5.RegisterDriver("cloudsql-postgres", cloudsqlconn.WithIAMAuthN())
```
keiko713, line 64: I can't really do `defer driverCleanup()` here as the cleanup needs to happen at the caller, main.go.
Since there was already a nice `shutdown` func, I piggybacked it and called `driverCleanup()` there.
Registering driver needs to happen only one time per the process, therefore it lives here and doing the nil check of driverCleanup.

https://github.com/GoogleCloudPlatform/cloud-sql-go-connector?tab=readme-ov-file#using-the-dialer-with-databasesql

runner/run.go
```diff
@@ -55,6 +59,14 @@ func Run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
 				logger.PrintError("Failed to initialize OpenTelemetry tracing provider, disabling exports: %s", err)
 			}
 		}
+
+		if cfg.DbUseIamAuth && cfg.SystemType == "google_cloudsql" && driverCleanup == nil {
```
keiko713, line 63: I was initially registering the driver for everyone (as I thought registering won't do any change unless we actually use it), but turned out that it doesn't work and integration tests (aka testing outside of Google Cloud SQL) failed with the following error.

> Failed to register cloudsql-postgres driver: failed to create default credentials: credentials: could not find default credentials. See https://cloud.google.com/docs/authentication/external/set-up-adc for more information

Since there was no problem running this with the GCP VM, I added this condition to only register when it's needed.

PR #672, opened 2025-02-08
Fix stats handling in partition merging when some relations are locked

msakrejda:
When some relations are locked, we add them to the
RelationInformations array, but do not add their stats to the
RelationStatistics array. This can cause problems because
mergePartitionSizes assumes that the arrays are parallel.

Update mergePartitionSizes to reference relations by the RelationIdx
field in RelationStatistic, which avoids this issue.

### DIFF
--- a/output/transform/merge_partition_sizes.go
+++ b/output/transform/merge_partition_sizes.go
@@ -12,11 +12,17 @@ import (
 //
 // TODO: recursively build up stats when nested partitioning is used
 func mergePartitionSizes(s snapshot.FullSnapshot, newState state.PersistedState, ts state.TransientState, databaseOidToIdx OidToIdx) snapshot.FullSnapshot {
+	relIdxToStatsIdx := make(map[int32]int, len(s.RelationStatistics))
+	for idx, stat := range s.RelationStatistics {
+		relIdxToStatsIdx[stat.RelationIdx] = idx
+	}
+
 	for idx, rel := range s.RelationInformations {
 		if !rel.HasParentRelation || rel.PartitionBoundary == "" {
 			continue
 		}
-		stat := s.RelationStatistics[idx]
+		statIdx := relIdxToStatsIdx[int32(idx)]
+		stat := s.RelationStatistics[statIdx]
 		parent := s.RelationStatistics[rel.ParentRelationIdx]
 		parent.NTupIns += stat.NTupIns
 		parent.NTupUpd += stat.NTupUpd
--- a/output/transform/merge_partition_sizes.go
+++ b/output/transform/merge_partition_sizes.go
@@ -21,7 +21,11 @@ func mergePartitionSizes(s snapshot.FullSnapshot, newState state.PersistedState,
 		if !rel.HasParentRelation || rel.PartitionBoundary == "" {
 			continue
 		}
-		statIdx := relIdxToStatsIdx[int32(idx)]
+		statIdx, ok := relIdxToStatsIdx[int32(idx)]
+		if !ok {
+			continue
+		}
+
 		stat := s.RelationStatistics[statIdx]
 		parent := s.RelationStatistics[rel.ParentRelationIdx]
 		parent.NTupIns += stat.NTupIns


output/transform/merge_partition_sizes.go
```diff
@@ -12,11 +12,17 @@ import (
 //
 // TODO: recursively build up stats when nested partitioning is used
 func mergePartitionSizes(s snapshot.FullSnapshot, newState state.PersistedState, ts state.TransientState, databaseOidToIdx OidToIdx) snapshot.FullSnapshot {
+	relIdxToStatsIdx := make(map[int32]int, len(s.RelationStatistics))
+	for idx, stat := range s.RelationStatistics {
```
msakrejda, line 16: Should we add the same check as lines 21-23 below here?

lfittl: Whilst that'd keep the map smaller, I don't think its worth the effort (either solution should be correct).

output/transform/merge_partition_sizes.go
```diff
@@ -12,11 +12,17 @@ import (
...
+		relIdxToStatsIdx[stat.RelationIdx] = idx
+	}
+
 	for idx, rel := range s.RelationInformations {
 		if !rel.HasParentRelation || rel.PartitionBoundary == "" {
 			continue
 		}
-		stat := s.RelationStatistics[idx]
+		statIdx := relIdxToStatsIdx[int32(idx)]
```
lfittl, line 24: Is there a reason the map couldn't use an `int` type to avoid this cast?

lfittl: Ah, but I suppose `stat.RelationIdx` is `int32` - that seems fine then.

msakrejda: Protobuf makes `stat.RelationIdx` an `int32`, so I think we need a cast somewhere...

output/transform/merge_partition_sizes.go
msakrejda: Right, that's what I was thinking about. Okay, thanks.

output/transform/merge_partition_sizes.go
keiko713: Uhm, so reading PR description, let's say there is a case of:
* s.RelationInformations has the table tbl1 (idx: 3) information
* s.RelationStatistics doesn't have the table tb1 stats

Correct? In that case, `relIdxToStatsIdx` doesn't have something in `relIdxToStatsIdx[3]` and we should skip doing below in that case?

msakrejda: Yeah, you're absolutely right: I'm building the map, but not using it correctly. Will fix.

output/transform/merge_partition_sizes.go
```diff
@@ -12,11 +12,21 @@ import (
...
 			continue
 		}
-		stat := s.RelationStatistics[idx]
+		statIdx, ok := relIdxToStatsIdx[int32(idx)]
+		if !ok {
+			continue
+		}
+
+		stat := s.RelationStatistics[statIdx]
 		parent := s.RelationStatistics[rel.ParentRelationIdx]
```
lfittl, line 30: I think this line is also incorrect - we need to do the same lookup here as well.

output/transform/merge_partition_sizes.go
```diff
@@ -12,11 +12,21 @@ import (
...
+	for idx, stat := range s.RelationStatistics {
+		relIdxToStatsIdx[stat.RelationIdx] = idx
+	}
+
 	for idx, rel := range s.RelationInformations {
 		if !rel.HasParentRelation || rel.PartitionBoundary == "" {
 			continue
 		}
-		stat := s.RelationStatistics[idx]
+		statIdx, ok := relIdxToStatsIdx[int32(idx)]
```
lfittl, line 24: I realized that technically we should be using `int32(rel.RelationIdx)` here, instead of `idx` - in practice they are the same I think, but the way its supposed to work is that the only array you can reference by index is `RelationReferences`, and both stats and info have an explicit `RelationIdx` field that should be used to look things up.

lfittl (commented):
I noticed two more minor things, see comments.

That said, looking at this code more does make me think if there isn't an alternate way to do this earlier in the process, before we're in the output format. I'll see if I can make a proposal for that in a separate PR shortly (since I'm already looking at the code anyway).

output/transform/merge_partition_sizes.go
lfittl: And we have the same problem for index statistics as well further down..

output/transform/merge_partition_sizes.go
lfittl: And we have the same problem for index statistics as well further down (i.e. we'd also need to build a map for index relation index to index stats entry index)

msakrejda:
Closing in favor of #674.

PR #674, opened 2025-02-11
Refactor partition stats aggregation

lfittl:

Alternate to #672 that refactors the code substantially in hopes to avoid bugs like that going forward instead.


input/postgres/relations.go
```diff
@@ -18,7 +18,8 @@ const relationsSQLOidField = "c.relhasoids AS relation_has_oids"
 const relationsSQLpg12OidField = "false AS relation_has_oids"
 
 const relationsSQL string = `
-	 WITH locked_relids AS (SELECT DISTINCT relation relid FROM pg_catalog.pg_locks WHERE mode = 'AccessExclusiveLock' AND relation IS NOT NULL AND locktype = 'relation')
+	 WITH locked_relids AS (SELECT DISTINCT relation relid FROM pg_catalog.pg_locks WHERE mode = 'AccessExclusiveLock' AND relation IS NOT NULL AND locktype = 'relation'),
+ partition_children AS (SELECT inhparent, array_agg(inhrelid) AS relids FROM pg_catalog.pg_inherits JOIN pg_catalog.pg_partitioned_table ON (partrelid = inhrelid) GROUP BY 1)
```
lfittl, line 22: In the (now removed) mergePartitionSizes code there was a TODO that we should include nested child partitions as well.

This is would now fairly easy to implement by making this a recursive CTE (or at least checking a second nesting level), but I think it may be best to change this behavior in a future PR (or even release, to let this change here settle for one cycle).

output/transform/postgres_relations.go
```diff
@@ -256,6 +273,19 @@ func transformPostgresRelations(s snapshot.FullSnapshot, newState state.Persiste
 			if diffedSchemaStatsExist {
 				indexStats = diffedSchemaStats.IndexStats[index.IndexOid]
 			}
+			// Add child index statistics to parent index statistics
```
lfittl, line 276: Note this part is substantially different from how this worked before (see commit message).

[In the original PR](https://github.com/pganalyze/collector/pull/635#discussion_r1847463790), @msakrejda also noted "I don't see a great way to avoid that without a considerably more expensive check" - I'm not clear if that referred to relying on `pg_inherits` (like we do here), or something else (possibly checking the index definitions somehow?).

msakrejda: I wasn't thinking about `pg_inherits`—that should avoid the problem.

msakrejda (approved):
Nice, I think this approach makes more sense.

seanlinsley:
I don't see anything wrong with this code, but when testing locally `log_lines_packed_7d` is missing all stats in the app.

lfittl:
> I don't see anything wrong with this code, but when testing locally `log_lines_packed_7d` is missing all stats in the app.

Thanks, I introduced a bug there in a refactoring in this PR - fixed that now and verified both table and index stats are handled correctly (I squashed this into the last commit, the bug was in what we counted as child OIDs).

PR #675, opened 2025-02-11
Rename cron scheduler

seanlinsley:
This PR updates the scheduler code so that the individual schedules are named by their frequency instead of using a purpose-specific name like, so for example `activity` becomes `TenSecond`. This change is needed for new features; it wouldn't make sense to capture 1 minute system stats using a `query_stats` schedule.

Since there's no need for this to be a dynamic map of schedules, I've replaced it with a statically-typed `Scheduler` struct. Note: `GetScheduler` was moved to the top of the file since it's the entrypoint into the file.

### DIFF
--- a/runner/run.go
+++ b/runner/run.go
@@ -29,7 +29,7 @@ func Run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
 	writeStateFile = func() {}
 	shutdown = func() {}
 
-	schedulerGroups, err := scheduler.GetSchedulerGroups()
+	scheduler, err := scheduler.GetScheduler()
 	if err != nil {
 		logger.PrintError("Error: Could not get scheduler groups")
 		return
@@ -245,7 +245,7 @@ func Run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
 		return
 	}
 
-	schedulerGroups["stats"].Schedule(ctx, func(ctx context.Context) {
+	scheduler.TenMinute.Schedule(ctx, func(ctx context.Context) {
 		wg.Add(1)
 		CollectAllServers(ctx, servers, globalCollectionOpts, logger)
 		wg.Done()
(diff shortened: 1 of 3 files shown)


msakrejda (approved):

runner/run.go
```diff
@@ -259,18 +259,18 @@ func Run(ctx context.Context, wg *sync.WaitGroup, globalCollectionOpts state.Col
...
 		}, logger, "activity snapshot of all servers")
 	}
 
-	schedulerGroups["query_stats"].ScheduleSecondary(ctx, func(ctx context.Context) {
+	scheduler.OneMinute.ScheduleSecondary(ctx, func(ctx context.Context) {
 		wg.Add(1)
 		GatherQueryStatsFromAllServers(ctx, servers, globalCollectionOpts, logger)
 		wg.Done()
-	}, logger, "high frequency query statistics of all servers", schedulerGroups["stats"])
+	}, logger, "high frequency query statistics of all servers", scheduler.TenMinute)
```
lfittl, line 273: I wonder if this could benefit from a comment, since we're not losing the context that this the primary group this is scheduled in relation to, is the `CollectAllServers` stats collection. Maybe:

```suggestion
	}, logger, "high frequency query statistics of all servers",
	scheduler.TenMinute // Skip at every 10 minute mark, when we collect full snapshots with query text
      )
```

scheduler/scheduler.go
```diff
@@ -8,6 +8,36 @@ import (
 	"github.com/pganalyze/collector/util"
 )
 
+type Scheduler struct {
+	TenSecond Group
+	OneMinute Group
+	TenMinute Group
+}
+
+func GetScheduler() (scheduler Scheduler, err error) {
```
lfittl, line 17: Personally I found `GetSchedulerGroups` to be more descriptive, since the structs being returned are also called `Group`, so I'd be inclined to keep the old name with the new structure.

I feel a "scheduler" might be more something that comes up with a schedule given user input (and returns that schedule) which is not what we're doing here.

lfittl (approved):
(and to be clear I'm also good to merge this as-is, just stylistic notes :) )

runner/run.go
seanlinsley: I wonder if the the schedule should be moved earlier in the argument list since it's fairly important. Then the comment can be above the function call:
```go
// Query stats are captured every minute, except for minute 10 when full snapshot collection takes over
scheduler.OneMinute.ScheduleSecondary(ctx, scheduler.TenMinute, func(ctx context.Context) {
	wg.Add(1)
	GatherQueryStatsFromAllServers(ctx, servers, globalCollectionOpts, logger)
	wg.Done()
}, logger, "high frequency query statistics of all servers")

scheduler/scheduler.go
seanlinsley: If we must have a two-word name for this struct, the best option might be `TaskScheduler`. But since there are no other schedule-like structs this might conflict with, I think it's fine to use `Scheduler`.

runner/run.go
lfittl: Yeah, good idea, I think moving it earlier works.

scheduler/scheduler.go
lfittl: Eh, I think `Scheduler` is fine if you prefer - we can leave as-is.

PR #676, opened 2025-02-13
Resolve data directory symlinks

msakrejda:
Resolve symlinks when finding the Postgres data directory

We currently do not resolve symlinks when looking up $PGDATA from the
environment or the `data_directory` GUC setting. This can misattribute
storage costs to the wrong partition.

Always follow symlinks in the helper that resolves the data directory.

### DIFF
--- a/helper/main.go
+++ b/helper/main.go
@@ -49,7 +49,7 @@ func getStatus(dataDirectory string) {
 		fmt.Fprintf(os.Stderr, "%s\n", err)
 	} else {
 		status.DataDirectory = dataDirectory
-		if dataDirectory == "" {
+		if status.DataDirectory == "" {
 			status.DataDirectory = os.Getenv("PGDATA")
 		}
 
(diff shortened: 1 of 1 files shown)


helper/main.go
```diff
@@ -48,17 +48,18 @@ func getStatus(dataDirectory string) {
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "%s\n", err)
 	} else {
-		if dataDirectory != "" {
-			status.DataDirectory = dataDirectory
-		} else {
+		status.DataDirectory = dataDirectory
+		if dataDirectory == "" {
```
lfittl, line 52: I was wondering if this should read `if status.DataDirectory == ""` instead, which feels ever so slightly more consistent to me.

lfittl (approved):

helper/main.go
msakrejda: Ah yes, good point.

PR #679, opened 2025-02-17
Refactor input logic with collection helper

lfittl:
See individual commits.

The motivation for doing this, besides making the code easier to work with, is to reduce repeated small queries we can answer from information already collected instead. This PR does that by moving pg_has_role lookups to the collection helper, and a separate follow-up PR will do this for the checks for helper functions using a single pg_proc query (see #680).


input/postgres/collection_helper.go
```diff
@@ -31,11 +35,37 @@ func NewCollectionHelper(ctx context.Context, logger *util.Logger, server *state
...
+	collectorRole := roleByName[server.Config.GetDbUsername()]
+	memberOf := collectorRole.MemberOf
+	for _, m := range memberOf { // Allow one level of indirect role memberships
+		memberOf = append(memberOf, roleByOid[m].MemberOf...)
+	}
+	connectedAsSuperUser := collectorRole.SuperUser ||
+		slices.Contains(memberOf, roleByName["rds_superuser"].Oid) ||
+		slices.Contains(memberOf, roleByName["azure_pg_admin"].Oid) ||
+		slices.Contains(memberOf, roleByName["cloudsqlsuperuser"].Oid)
+	connectedAsMonitoringRole := slices.Contains(memberOf, roleByName["pg_monitor"].Oid)
```
lfittl, line 59: Per PR review, preference to keep a database connection check here (vs implementing our own pg_has_role).

input/postgres/collection_helper.go
```diff
@@ -0,0 +1,71 @@
...
+	"database/sql"
+	"fmt"
+	"slices"
+
+	"github.com/pganalyze/collector/config"
+	"github.com/pganalyze/collector/state"
+	"github.com/pganalyze/collector/util"
+)
+
+type CollectionHelper struct {
```
lfittl, line 14: Add comment explaining what this does

Also rename to "Collection" or "CollectionState"?

input/full.go
```diff
@@ -20,23 +19,14 @@ import (
 
 // CollectFull - Collects a "full" snapshot of all data we need on a regular interval
 func CollectFull(ctx context.Context, server *state.Server, connection *sql.DB, globalCollectionOpts state.CollectionOpts, logger *util.Logger) (ps state.PersistedState, ts state.TransientState, err error) {
-	systemType := server.Config.SystemType
 	ps.CollectedAt = time.Now()
 
-	ts.Version, err = postgres.GetPostgresVersion(ctx, logger, connection)
+	h, err := postgres.NewCollectionHelper(ctx, logger, server, globalCollectionOpts, connection)
```
msakrejda, line 24: I still like just `postgres.NewCollection`, as I mentioned in PR Review. `Helper` doesn't really say anything, and conceptually I think it makes sense that a Collection is what eventually builds a Snapshot.

runner/generate_helper_sql.go
```diff
@@ -45,12 +45,7 @@ func GenerateStatsHelperSql(ctx context.Context, server *state.Server, globalCol
 	}
 	defer db.Close()
 
-	version, err := postgres.GetPostgresVersion(ctx, logger, db)
-	if err != nil {
-		return "", fmt.Errorf("error collecting Postgres version: %s", err)
-	}
```
msakrejda, line 51: Are these removed because we're moving the version check earlier in later patches?

input/postgres/collection_helper.go
msakrejda: Yeah, I think doing this once is great, but doing it in Go does not seem like a good trade-off to me.

input/postgres/collection_helper.go
msakrejda: As we discussed in PR review, it's not really just state, so I would avoid the State suffix. We could also use a synonym of Collection like Lot (as in "a number of units of an article, a single article, or a parcel of articles offered as one item (as in an auction sale)") or Batch (I think there's too much potential for confusion with that one, but throwing it out there).

runner/generate_helper_sql.go
lfittl: This was removed because we don't actually use the version information anywhere in this method - it was presumably done before because `GetDatabases` unnecessarily had a `version` argument (that it did not use).

input/full.go
```diff
@@ -20,39 +19,25 @@ import (
...
+	ts.Databases, ps.DatabaseStats, err = postgres.GetDatabases(ctx, connection)
 	if err != nil {
 		logger.PrintError("Error collecting pg_databases: %s", err)
 		return
 	}
 
 	bufferCacheReady := make(chan state.BufferCache)
 	go func() {
-		postgres.GetBufferCache(ctx, server, globalCollectionOpts, logger, ts.Version, bufferCacheReady)
+		postgres.GetBufferCache(ctx, c, server, globalCollectionOpts, bufferCacheReady)
```
msakrejda, line 40: Not related to the refactor, but why does this establish its own connection instead of using the one we establish above? :confused:

input/postgres/roles.go
```diff
@@ -27,15 +28,17 @@ SELECT oid,
...
-	if err != nil {
-		return nil, err
+func getRoles(ctx context.Context, db *sql.DB, systemType string) ([]state.PostgresRole, error) {
+	cloudSuperuserName := ""
+	if systemType == "amazon_rds" {
+		cloudSuperuserName = "rds_superuser"
+	} else if systemType == "azure_database" {
+		cloudSuperuserName = "azure_pg_admin"
+	} else if systemType == "google_cloudsql" {
+		cloudSuperuserName = "cloudsqlsuperuser"
```
msakrejda, line 38: So this works, but I wonder if we get in a situation where `systemType` is not detected properly, if that could cascade to detecting this improperly. I was thinking an alternative would be to just hard-code these all (since an unintentional name clash on a different system is vanishingly unlikely), but I think in practice it won't make much difference either way, so maybe it's best to just leave it.

input/postgres/version.go
```diff
@@ -6,11 +6,10 @@ import (
 	"strings"
 
 	"github.com/pganalyze/collector/state"
-	"github.com/pganalyze/collector/util"
 )
 
 // GetPostgresVersion - Reads the version of the connected PostgreSQL server
-func GetPostgresVersion(ctx context.Context, logger *util.Logger, db *sql.DB) (version state.PostgresVersion, err error) {
+func GetPostgresVersion(ctx context.Context, db *sql.DB) (version state.PostgresVersion, err error) {
```
msakrejda, line 12: `NewCollection` is now the only caller of this method—should we lowercase the name to avoid exporting it and make it clear that info should come from `Collection`? Unfortunately, `GetIsAwsAurora` (below) does not look as easy to eliminate, but we can leave that for another time.

msakrejda (approved):
Looks great!

input/full.go
lfittl: That's intentional since it runs in a separate Go routine (since fetching buffercache stats is slow yet independent of all other data) and we don't have multiplexing enabled on the DB connection.

input/postgres/roles.go
lfittl: Yeah, I was wondering about that too. I think I'll leave as is for now mainly because that's what the existing logic does - but not opposed to changing this in a future PR.

input/postgres/version.go
lfittl: Ah, yes, good catch, I did that for `GetRoles` => `getRoles` but missed it here.

input/full.go
msakrejda: Makes sense. We should probably add a comment explaining that (but not in this PR).

PR #680, opened 2025-02-17
Improve helper function checks for consistency and lower overhead

lfittl:
This moves the checking for the presence of helper functions to the collection helper, and relies on an upfront querying of pg_proc, instead of querying for each helper individually. Note that we have to support this both for the main database (for all server-wide data points), as well as for each database to get column/extended statistics information.

In passing this also improves the support for checking for different (potentially overloaded) versions of the same functions, by requiring passing the function argument types to the lookup. That is specifically necessary for the multiple historic variants of get_stat_statements that may be present on the same database server, and lets us remove special casing that code.

On top of https://github.com/pganalyze/collector/pull/679.


input/postgres/collection_helper.go
```diff
@@ -67,5 +92,69 @@ func NewCollectionHelper(ctx context.Context, logger *util.Logger, server *state
...
+			}
+		}
+		if !mismatch {
+			return f, true
+		}
+	}
+	return state.PostgresFunction{}, false
+}
+
+func (s CollectionHelper) HelperExists(name string, inputTypes []string) bool {
```
lfittl, line 149: For context, I plan to add additional behavior here soon that warns the user (and flags the SelfTest logic) in case the function ownership is not as expected. To keep reviews easy I've not yet added this here.

runner/query_run.go
```diff
@@ -80,7 +80,12 @@ func runQueryOnDatabase(ctx context.Context, server *state.Server, collectionOpt
 	}
 	defer db.Close()
 
-	if postgres.StatsHelperExists(ctx, db, "explain_analyze") {
+	h, err := postgres.NewCollectionHelper(ctx, logger, server, collectionOpts, db)
```
lfittl, line 83: Maybe consider adding a different constructor or argument here to initialize without the data that's not needed (like Postgres version in this case).

lfittl: After thinking about it some more I think it'd be best if we always fetch version and role information (i.e. not make it conditional).

We'll need roles soon enough (to check helper function ownership), and whilst we don't need version in this code path right now, I could anticipate that changing (e.g. to conditionally set certain EXPLAIN flags).

input/postgres/statements.go
```diff
@@ -180,7 +156,7 @@ func GetStatements(ctx context.Context, c *Collection, db *sql.DB, showtext bool
 	}
 
 	usingStatsHelper := false
-	if statementStatsHelperExists(ctx, db, showtext) {
+	if c.HelperExists("get_stat_statements", []string{"boolean"}) || (showtext && c.HelperExists("get_stat_statements", nil)) {
```
lfittl, line 159: Its worth noting that this logic is slightly different from before (its basically more precise).

This condition says that we either:
- Need the helper with arg that allows us to differentiate between with-query-text and without-query-text modes
- Or if we're looking for the version with text we can also use the old-style helper that doesn't take the argument (and always returns with text)

I think the practical intent here is that we don't fetch the version with query text every once-a-minute. Now, whether that actually works well in practice I'm not 100% sure (maybe we should just not support the old-style helper here at all anymore?), but its the same intent as before.

input/postgres/collection.go
```diff
@@ -56,5 +80,69 @@ func NewCollection(ctx context.Context, logger *util.Logger, server *state.Serve
...
+		}
+		mismatch := false
+		for idx, arg := range args {
+			// Split by this assumed pattern: "NAME TYPE DEFAULT DEFAULT_VALUE"
+			//
+			// Note this currently does not handle data types with spaces in
+			// them, such as "double precision", or other cases not expected
+			// with our known set of helper functions.
+			parts := strings.Split(arg, " ")
+			if idx >= len(inputTypes) && parts[2] != "DEFAULT" {
```
msakrejda, line 120: So extra arguments with a default are okay, right? Maybe we should add a comment to make that more obvious.

input/postgres/collection.go
```diff
@@ -56,5 +80,69 @@ func NewCollection(ctx context.Context, logger *util.Logger, server *state.Serve
...
+	}
+	for _, f := range funcs {
+		args := strings.Split(f.Arguments, ", ")
+		if len(inputTypes) > len(args) {
+			// We're expecting more arguments than the function has
+			continue
+		}
+		mismatch := false
+		for idx, arg := range args {
+			// Split by this assumed pattern: "NAME TYPE DEFAULT DEFAULT_VALUE"
```
msakrejda, line 114:
```suggestion
			// Split by the assumed output pattern of pg_get_function_arguments: "<name> <type> DEFAULT <default>"
```
(or some other rendering that distinguishes the literal in this list from the other values).

input/postgres/schema.go
```diff
@@ -182,14 +190,6 @@ func collectSchemaData(ctx context.Context, c *Collection, db *sql.DB, ps state.
 		}
 	}
 
-	if c.GlobalOpts.CollectPostgresFunctions {
```
msakrejda, line 185: We've discussed that this is probably not useful, but if we're going to stop using it, we should fully remove it.

msakrejda (approved):
Some minor comments, but looks good.

PR #681, opened 2025-02-19
GH Actions: Pin QEMU action binfmt to qemu-v7.0.0-28

lfittl:
Newer versions cause segmentation fault issues, see https://github.com/docker/setup-qemu-action/issues/198

### DIFF
--- a/.github/workflows/test-packages.yml
+++ b/.github/workflows/test-packages.yml
@@ -23,6 +23,7 @@ jobs:
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v3
       with:
+        image: tonistiigi/binfmt:qemu-v7.0.0-28
         platforms: arm64
 
     - name: Build packages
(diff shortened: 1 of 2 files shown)


lfittl:
Kicked off a test packages run here: https://github.com/pganalyze/collector/actions/runs/13418343017

.github/workflows/test-packages.yml
```diff
@@ -23,6 +23,7 @@ jobs:
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v3
       with:
+        image: tonistiigi/binfmt:qemu-v7.0.0-28
```
msakrejda, line 26: I _think_ you actually need the "full" image tag here, starting with `docker.io/`: see how the default is handled [here](https://github.com/docker/setup-qemu-action/blob/master/src/context.ts#L12).

lfittl: Hmm, interesting. I think that's technically more correct, however I've seen other repositories fix it like this without adding the `docker.io/` prefix.

lfittl:
Superseded by #682

PR #683, opened 2025-02-20
Tweak explain_analyze helper invocation to avoid pg_hint_plan limitation

msakrejda:
For the collector-based Query Tuning workflow, the collector receives
a query text from the pganalyze server and runs it on the target
database. This query text may include pg_hint_plan hints, as users use
workbooks to optimize queries.

However, it looks like pg_hint_plan ignores hints if they are not in
the _first_ comment in a query. The collector prepends a query marker
as a separate comment, and this causes pg_hint_plan to ignore any
hints specified in the query.

Avoid passing our query marker, since we're already prepending it to
the top-level explain_analyze invocation, and nested queries will not
show up in pg_stat_activity anyway.

### DIFF
--- a/input/postgres/explain_analyze.go
+++ b/input/postgres/explain_analyze.go
@@ -52,7 +52,7 @@ func runExplainAnalyze(ctx context.Context, db *sql.DB, query string, parameters
 	}
 	defer tx.Rollback()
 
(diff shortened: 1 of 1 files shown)


lfittl (approved):
:+1:

PR #684, opened 2025-02-20
GH release action: Fix "packages" name conflict

lfittl:
During the release process we group together the created packages as an artifact, to then upload them individually to the GitHub release entry.

Previously this was able to use a single "packages" zip file for both x86_64 and arm64 packages, but we now need to split this since we split the actions.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,9 +7,7 @@ on:
 
 jobs:
   build:
-    # Use older Ubuntu release to avoid building with a too new glibc, which will cause
-    # errors on older distributions, if anyone uses the generated binary directly.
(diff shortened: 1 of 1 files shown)


msakrejda (approved):

.github/workflows/release.yml
```diff
@@ -56,7 +54,7 @@ jobs:
     - name: Set up Go 1.x
       uses: actions/setup-go@v5
       with:
-        go-version: 1.21
+        go-version: 1.23
```
lfittl, line 57: Looks like this was missed in the Go version change by accident.

keiko713: ah good catch, thanks for fixing this!

PR #687, opened 2025-03-07
Automatically vendor protoc if not found

msakrejda:
Automatically vendor protoc if the correct version is not found in
PATH (or not already vendored).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,17 @@ PATH := $(PWD)/protoc/bin:$(PWD)/bin:$(PATH)
 SHELL := env PATH=$(PATH) /bin/sh
 
 PROTOC_VERSION_NEEDED := 28.2
-PROTOC_VERSION := $(shell command -v protoc > /dev/null 2>&1 && protoc --version)
+PROTOC_VERSION := $(shell command -v protoc > /dev/null 2>&1 && protoc --version || echo 'none')
+PROTOC_BASE_URL := https://github.com/protocolbuffers/protobuf/releases/download/
+
+ARCH := $(shell uname -m)
+ifeq ($(shell uname), Linux)
+	PROTOC_FILENAME := protoc-$(PROTOC_VERSION_NEEDED)-linux-$(ARCH).zip
+endif
+ifeq ($(shell uname), Darwin)
+	PROTOC_FILENAME := protoc-$(PROTOC_VERSION_NEEDED)-osx-$(ARCH).zip
+endif
+PROTOC_URL := $(PROTOC_BASE_URL)/v$(PROTOC_VERSION_NEEDED)/$(PROTOC_FILENAME)
 
 .PHONY: default build build_dist vendor test docker_release packages integration_test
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -5,7 +5,17 @@ PATH := $(PWD)/protoc/bin:$(PWD)/bin:$(PATH)
...
-PROTOC_VERSION := $(shell command -v protoc > /dev/null 2>&1 && protoc --version)
+PROTOC_VERSION := $(shell command -v protoc > /dev/null 2>&1 && protoc --version || echo 'none')
+PROTOC_BASE_URL := https://github.com/protocolbuffers/protobuf/releases/download/
+
+ARCH := $(shell uname -m)
+ifeq ($(shell uname), Linux)
+	PROTOC_FILENAME := protoc-$(PROTOC_VERSION_NEEDED)-linux-$(ARCH).zip
+endif
+ifeq ($(shell uname), Darwin)
+	PROTOC_FILENAME := protoc-$(PROTOC_VERSION_NEEDED)-osx-$(ARCH).zip
```
keiko713, line 16: sadly this generates things like the following:

```
curl --location --output tmp/protoc-28.2-osx-arm64.zip https://github.com/protocolbuffers/protobuf/releases/download//v28.2/protoc-28.2-osx-arm64.zip
```

Though arm one is likely `aarch_64` in the download location, so you'll need to update this. (using `aarch_64` instead worked well with my machine)

msakrejda: Thanks for checking--can you try the updated code?

keiko713: worked well 🎉

Makefile
```diff
@@ -73,13 +89,14 @@ else
...
-	@echo "ℹ️  Note the download's name will look like this: protoc-${PROTOC_VERSION_NEEDED}-osx-x86_64.zip"
-	@echo "ℹ️  Copy the unzipped folder into this project, and rename it to \"protoc\""
+	@echo "ℹ️  Vendoring protoc $(PROTOC_VERSION_NEEDED)"
+  
+	@mkdir -p tmp
+	curl --location --output tmp/$(PROTOC_FILENAME) $(PROTOC_URL)
+	unzip -d protoc tmp/$(PROTOC_FILENAME)
+	rm tmp/$(PROTOC_FILENAME)
+  
 	@echo "ℹ️  If this is macOS, you will need to try running the binary yourself, then go to Security & Privacy to explicitly allow it."
```
keiko713, line 101: Definitely not in the scope of this PR but I realized that we have this line, but we kinda addressed this in https://github.com/pganalyze/collector/pull/667 so maybe we can remove this line.

keiko713 (approved):
We could ask Sean to test also maybe, but looks good to me 👍

Makefile
msakrejda: This is talking about the `protoc` binary, though, right? Wasn't #667 was about the generated collector binary?

keiko713: ahhh good call! well ok let's leave it as is

PR #688, opened 2025-03-11
unpackPostgresOidArray: Correctly support 64-bit OIDs

lfittl:
Postgres allows OIDs to get very large (beyond the 32-bit integer boundary) which can cause problems in practice when unpacking arrays of OIDs.


msakrejda (commented):
I take that back—reviewing that original PR, I don't see any other problem cases (assuming the "integer" type for config settings is 32 bits, which I think is right).

msakrejda (commented):
Oh, though we should update the PR description that this is about unsigned 32-bit values, not 64-bit ones.

PR #690, opened 2025-03-14
Remove special error handling as func no longer returns that error

keiko713:
While I was checking some other place, I realized that this part of the code is no longer reachable.
That particular error message from `azure.SetupLogSubscriber` was retired in https://github.com/pganalyze/collector/pull/415

https://github.com/pganalyze/collector/pull/415/files#diff-6f5880431e90b30b7cc480560bbf2f8e4a80fcbe85c03ca3cacae1276cb3512eL67

Let's make sure to remove it, also given that it's been a while since that change happened, I'd say that it's safe to remove the docs too (will do that in the docs PR).


lfittl (approved):
:+1: Makes sense, this is clearly a no-op now.

PR #691, opened 2025-03-14
Collection helper detection: Improve handling of default arguments

lfittl:
This previously crashed in some cases due to an incomplete refactoring I had done in a prior commit. Add test coverage to prevent recurrence.

In passing move SQL helper definitions to a dedicated util/helpers directory and add the helper definition for "get_stat_statements", which whilst not used much these days, is a good test case here.


input/postgres/collection_test.go
```diff
@@ -0,0 +1,92 @@
...
+	{
+		"get_stat_statements",
+		[]string{},
+		true,
+		"SETOF pg_stat_statements",
+	},
+}
+
+func TestCollectionFindHelper(t *testing.T) {
+	db := setupTest(t)
```
lfittl, line 63: Note this is defined in the [explain_analyze_test.go](https://github.com/pganalyze/collector/blob/main/input/postgres/explain_analyze_test.go) file.

input/postgres/collection_test.go
```diff
@@ -0,0 +1,97 @@
...
+		"SETOF pg_stat_statements",
+	},
+	// Verify we don't match on a shorter set of arguments
+	{
+		"explain_analyze",
+		[]string{"text", "text[]", "text[]"},
+		false,
+		"",
+	},
+	// Verify we don't match on a shorter set of arguments
```
msakrejda, line 39:
```suggestion
	// Verify we don't match on a longer set of arguments
```
I assume?

msakrejda (approved):

PR #692, opened 2025-03-19
Avoid invoking non-existent helper for system stats with Crunchy Bridge

keiko713:
Since we're running `selfhosted.GetSystemState` with Crunchy Bridge, the test result shows the following a bit confusing message:

```
	✗ System:              	error running system stats helper process: fork/exec /usr/bin/pganalyze-collector-helper: no such file or directory
```

This is an expected error as the helper file doesn't exist, but it is alarming, so let's not flag this error with Crunchy.

### DIFF
--- a/input/system/selfhosted/system.go
+++ b/input/system/selfhosted/system.go
@@ -27,7 +27,7 @@ type helperStatus struct {
 }
 
 // GetSystemState - Gets system information about a self-hosted (physical/virtual) system
-func GetSystemState(server *state.Server, logger *util.Logger, ignoreHelperRun bool) (system state.SystemState) {
+func GetSystemState(server *state.Server, logger *util.Logger, skipHelperRun bool) (system state.SystemState) {
 	config := server.Config
 	var status helperStatus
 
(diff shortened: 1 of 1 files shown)


input/system/selfhosted/system.go
```diff
@@ -38,8 +38,10 @@ func GetSystemState(server *state.Server, logger *util.Logger) (system state.Sys
 
 	statusBytes, err := exec.Command("/usr/bin/pganalyze-collector-helper", "status", config.DbDataDirectory).Output()
```
lfittl, line 39: Shouldn't we skip the whole helper invocation, since its not going to do anything useful on Crunchy Bridge?

keiko713: Yeah I was thinking about that, let me tweak a bit more

msakrejda (approved):
Nice, thanks for fixing this.

I think this approach works (and feel free to merge as-is), but it complicates the existing Crunchy dependency on self-managed (a somewhat dubious idea in the first place), which I think makes the code harder to follow. One thing we could do instead is factor out the shared code and split out the self-managed specific part. E.g., something like

```golang
		systemHelperState := selfhosted.GetSystemStateFromHelper(server, logger)
		system = selfhosted.GetSystemState(server, logger, systemHelperState)
	} else {
```
And the Crunchy code could just pass an empty systemHelperState struct or something. These are terrible names, but I think you get the idea.

msakrejda:
I also took the liberty of updating the PR title to avoid confusion. Feel free to edit further, of course.

keiko713:
> I think this approach works (and feel free to merge as-is), but it complicates the existing Crunchy dependency on self-managed (a somewhat dubious idea in the first place), which I think makes the code harder to follow. One thing we could do instead is factor out the shared code and split out the self-managed specific part. E.g., something like

Thanks for the idea! I agree that this is really a workaround (well, Crunchy using self managed one is already a workaround). I tried to extract the part that pulls data from a helper, though it didn't really click me (like you noticed, we do need to return `systemHelperState` (`helperStatus`), then we need to have a special handling of putting the values of `helperStatus` to `state.SystemState`, etc. etc.), so I'll merge this as is.
Also thanks for updating the title!

PR #693, opened 2025-03-20
Document manual testing of receiving logs over syslog

msakrejda:
I had to figure this out when testing https://github.com/pganalyze/collector/pull/673 . I was not able to automate the test, but I think we should at least document the manual steps.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,6 +61,8 @@ make integration_test
 
 Note the integration tests require Docker, and will take a while to run through.
 
+#### Test receiving Postgres logs over syslog
+
 The ability to receive Postgres logs over syslog is not covered by
 automated tests. To test this manually, you'll need to relay the
 Postgres logs through a syslog daemon.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -61,6 +61,39 @@ make integration_test
 
 Note the integration tests require Docker, and will take a while to run through.
 
+The ability to receive Postgres logs over syslog is not covered by
```
keiko713, line 64: Nice, thanks for adding the section. I haven't tried it but given that you tried, I think it's fine to simply add it.
I think it'd be nice to add some title with this section (maybe with `####`, like `#### Test receiving Postgres logs over syslog`), to make it clear that it's one chunk and kinda optional thing to do. What do you think?

keiko713 (approved):

CONTRIBUTING.md
msakrejda: Sure. I thought about that, but the rest of the section is not really subdivided. However, this really is a sort of footnote to the testing docs, so I think you're right. I'll adjust.

msakrejda, line 64:
```suggestion
#### Test receiving Postgres logs over syslog

The ability to receive Postgres logs over syslog is not covered by
```

PR #695, opened 2025-03-28
Helper function handling: Fix detection of functions without arguments

lfittl:
This accidentally broke in a recent refactoring, and caused helpers with no arguments (e.g. "get_column_stats") to not be detected correctly.

In passing refactor definitions of "get_column_stats" and "get_relation_stats_ext" helpers to be in SQL files.

### DIFF
--- a/util/helpers/get_column_stats.sql
+++ b/util/helpers/get_column_stats.sql
@@ -1,4 +1,5 @@
-CREATE OR REPLACE FUNCTION pganalyze.get_column_stats() RETURNS TABLE(
+DROP FUNCTION IF EXISTS pganalyze.get_column_stats;
+CREATE FUNCTION pganalyze.get_column_stats() RETURNS TABLE(
   schemaname name, tablename name, attname name, inherited bool, null_frac real, avg_width int, n_distinct real, correlation real
 ) AS $$
   /* pganalyze-collector */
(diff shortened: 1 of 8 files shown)


util/helpers/get_column_stats.sql
```diff
@@ -0,0 +1,8 @@
+CREATE OR REPLACE FUNCTION pganalyze.get_column_stats() RETURNS TABLE(
```
seanlinsley, line 1: Since the return type changed in #610, `CREATE OR REPLACE FUNCTION` can raise an error. We should probably use `DROP FUNCTION IF EXISTS` across the board to avoid errors like that in the future.

lfittl: Yeah, I was on the fence whether its okay to switch to `CREATE OR REPLACE FUNCTION` but I can just keep this as DROP IF EXISTS + CREATE.

seanlinsley: Since we're on the topic, maybe it's good to update the other helper functions?

seanlinsley (approved):

util/helpers/get_column_stats.sql
lfittl: Sure, I can update it for consistency.

msakrejda (approved):

PR #696, opened 2025-04-03
Update pg_query_go to v6.1.0

lfittl:
This fixes builds on macOS 15.4+


vendor/github.com/pganalyze/pg_query_go/v6/parser/build_cgo.go
```diff
@@ -0,0 +1,54 @@
+//go:build required
```
lfittl, line 1: See https://github.com/pganalyze/pg_query_go/pull/131 for context on these files (I'll look into dropping modvendor separately in another PR after this one is in)

seanlinsley (approved):

PR #700, opened 2025-04-10
Better support for running EXPLAIN with pganalyze.explain_analyze helper

keiko713:
With more tests, we noticed that passing parameter values and types separately to this function is not working well for several cases.

For values, as `pg_catelog.string_agg` will never emit NULLs and only aggregates non-null values, it fails when we actually want to pass NULL as a value. Use array_to_string instead as a workaround.

For types, when something like int4[] is passed, as it was quoting as "int4[]" previously, this also didn't work. We check the input type and if it matches to the regexp (type pattern, especially ones the pganalyze app would pass), it skips quote_ident.

### DIFF
--- a/util/helpers/explain_analyze.sql
+++ b/util/helpers/explain_analyze.sql
@@ -2,6 +2,7 @@ CREATE OR REPLACE FUNCTION pganalyze.explain_analyze(query text, params text[], 
 DECLARE
   prepared_query text;
   params_str text;
+  params_str2 text;
   param_types_str text;
   explain_prefix text;
   explain_flag text;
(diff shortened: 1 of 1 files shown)


util/helpers/explain_analyze.sql
```diff
@@ -2,6 +2,7 @@ CREATE OR REPLACE FUNCTION pganalyze.explain_analyze(query text, params text[],
 DECLARE
   prepared_query text;
   params_str text;
+  params_str2 text;
```
seanlinsley, line 5: Looks like this isn't used

seanlinsley (approved):

PR #702, opened 2025-05-02
Pick up the correct oldest xmin for Aurora replication

msakrejda:
Aurora does not include the correct replication details in
pg_stat_replication, so check aurora_replica_status and use the older
of the two values to track XminHorizonStandby on the platform.

### DIFF
--- a/input/postgres/server_stats.go
+++ b/input/postgres/server_stats.go
@@ -61,8 +61,17 @@ COALESCE((
 COALESCE((
 	SELECT
 		backend_xmin
-	FROM %s
-	WHERE backend_xmin IS NOT NULL
+	FROM (
+		SELECT
+			backend_xmin
+		FROM %s
+			UNION ALL
+		SELECT
+			feedback_xmin as backend_xmin
+		FROM
+			%s
+		WHERE feedback_xmin IS NOT NULL
+	) _(backend_xmin)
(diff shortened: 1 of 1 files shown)


input/postgres/server_stats.go
```diff
@@ -61,8 +61,17 @@ COALESCE((
...
 		backend_xmin
-	FROM %s
-	WHERE backend_xmin IS NOT NULL
+	FROM (
+		SELECT
+			backend_xmin
+		FROM %s
+			UNION ALL
+		SELECT
+			feedback_xmin as backend_xmin
```
keiko713, line 70: I have some sad news for you:

```
postgres=> select pg_typeof(feedback_xmin) from aurora_replica_status() limit 1;
 pg_typeof 
-----------
 bigint
(1 row)
```

input/postgres/server_stats.go
```diff
@@ -61,8 +61,17 @@ COALESCE((
...
+		FROM
+			%s
```
keiko713, line 72: super nitpick but it would be nice to use either `FROM %s` (used above) or this

msakrejda:
CI failed on fetching the Postgres container. Retrying.

input/postgres/server_stats.go
keiko713: Tested and it worked 🎉

keiko713 (approved):

PR #703, opened 2025-05-06
Fetch system metrics from API with Crunchy Bridge

keiko713:
Currently, we're using gopsutil to fetch Crunchy's system metrics, assuming that the collector is deployed on Container Apps. Recently, with the change of cgroups with Container Apps, we are no longer able to pull these system info using gopsutil.

With this PR, instead of trying to get the system metrics using gopsutil, fetch these metrics from the API. Below is how it looks in the Crunchy Bridge UI and the pganalyze UI.

### CPU & Load Average section
* pganalyze UI has "idle", which is kinda hack not to make other percentage inflate - you can see this more in the code comment
* In the pganalyze UI, we should likely be hiding load avg 5 mins and 15 mins as it doesn't exist and it's likely confusing to people for why it's zero
image
image

### Memory
* "memory used" in Crunchy is shown as "Process Memory" in pganalyze
* In the pganalyze UI, maaaybe we should hide metrics that are not available in Crunchy (like we want to do above in load average)
image
image

### I/O & Storage
* Storage space graph hasn't changed so it should be in a good shape
* IOPS is the one that is now using the API metrics, looking good
* We have other things in the pganalyze UI that we no longer report (I/O utilization in the screenshot, I/O Latency, etc.) - we should hide them
image
image

### DIFF
--- a/input/system/crunchy_bridge/system.go
+++ b/input/system/crunchy_bridge/system.go
@@ -49,6 +49,12 @@ func GetSystemState(ctx context.Context, server *state.Server, logger *util.Logg
 		logger.PrintError("CrunchyBridge/System: Encountered error when getting cluster CPU metrics %v\n", err)
 		return
 	}
+	system.CPUStats = make(state.CPUStatisticMap)
+	// In the pganalyze UI, we re-calculate these percentage based on all available percentages
+	// Meaning, even the user only has 10%, it can inflate to 40% if there is no other item that is using lots of %
+	// To avoid this, use idle as "others %" - it might not be correct as there would be other CPU usages, but it at least helps
+	// from showing the wrong percentage for more important ones like system and user
+	cpuIdlePercent := 100 - cpuMetrics.Iowait - cpuMetrics.System - cpuMetrics.User - cpuMetrics.Steal
 	system.CPUStats["all"] = state.CPUStatistic{
 		DiffedOnInput: true,
 		DiffedValues: &state.DiffedSystemCPUStats{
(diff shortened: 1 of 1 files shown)


input/system/crunchy_bridge/system.go
```diff
@@ -50,6 +43,73 @@ func GetSystemState(ctx context.Context, server *state.Server, logger *util.Logg
...
+	if err != nil {
+		server.SelfTest.MarkCollectionAspectError(state.CollectionAspectSystemStats, "error getting cluster memory metrics: %s", err)
+		logger.PrintError("CrunchyBridge/System: Encountered error when getting cluster memory metrics %v\n", err)
+		return
+	}
+	totalMemoryInBytes := clusterInfo.Memory * 1024 * 1024 * 1024
+	system.Memory.TotalBytes = uint64(totalMemoryInBytes)
+	// ApplicationBytes is not the best way for representing "used bytes",
+	// but in the UI side, we use this as "process" memory if this value exits
+	// which would be the closest to the used bytes
```
lfittl, line 90: This matches Heroku and makes sense to me.

input/system/crunchy_bridge/system.go
```diff
@@ -50,6 +43,73 @@ func GetSystemState(ctx context.Context, server *state.Server, logger *util.Logg
...
+		server.SelfTest.MarkCollectionAspectError(state.CollectionAspectSystemStats, "error getting cluster CPU metrics: %s", err)
+		logger.PrintError("CrunchyBridge/System: Encountered error when getting cluster CPU metrics %v\n", err)
+		return
+	}
+	system.CPUStats = make(state.CPUStatisticMap)
+	// In the pganalyze UI, we re-calculate these percentage based on all available percentages
+	// Meaning, even the user only has 10%, it can inflate to 40% if there is no other item that is using lots of %
+	// To avoid this, use idle as "others %" - it might not be correct as there would be other CPU usages, but it at least helps
+	// from showing the wrong percentage for more important ones like system and user
+	cpuIdlePercent := 100 - cpuMetrics.Iowait - cpuMetrics.System - cpuMetrics.User - cpuMetrics.Steal
```
lfittl, line 57: Makes sense, I think its reasonable to do this math here to make the graph work as expected.

lfittl (approved):

PR #704, opened 2025-05-07
Updated cloudsql module and defaulted to using private IP.

ryanbooz:
The CloudSQL driver discovers the IP address for an instance based on the project, instance, and region settings. If public IP access is not available, we must specify to use the private IP, which is presumably more often the default.

Added an option to use the public IP if needed.


runner/run.go
```diff
@@ -32,6 +32,7 @@ func Run(ctx context.Context, wg *sync.WaitGroup, opts state.CollectionOpts, log
 	writeStateFile = func() {}
 	shutdown = func() {}
 	var driverCleanup func() error
+	var driverCleanupPublic func() error
```
lfittl, line 35: We also need to call this in the `shutdown` function later

config/config.go
```diff
@@ -111,6 +111,7 @@ type ServerConfig struct {
 	GcpPubsubSubscription string `ini:"gcp_pubsub_subscription"`
 	GcpCredentialsFile    string `ini:"gcp_credentials_file"`
 	GcpRegion             string `ini:"gcp_region"`
+	GcpUsePublicIP        bool   `ini:"gcp_use_public_ip"`
```
lfittl, line 114: We also need to add the environment variable for this around here: https://github.com/pganalyze/collector/blob/main/config/read.go#L252 (use the `parseConfigBool` method to handle the boolean value there)

lfittl (commented):
Overall looks good :+1:, to minor comments, then we should be good to merge.

lfittl (approved):

PR #708, opened 2025-05-26
Adding AlloyDB IAM connection

ryanbooz:
To connect to AlloyDB with IAM authentication requires a new Go package and driver connection. I think this would be cleaner if we added a new `systemType` but that will require coordination with the `pganalyze` application.


input/postgres/establish_connection.go
```diff
@@ -66,20 +66,32 @@ func connectToDb(ctx context.Context, config config.ServerConfig, logger *util.L
 				dbPasswordOverride = dbToken
 			}
 		} else if config.SystemType == "google_cloudsql" {
-			if config.GcpProjectID == "" || config.GcpRegion == "" || config.GcpCloudSQLInstanceID == "" {
-				return nil, errors.New("To use IAM auth with Google Cloud SQL, you must specify project ID, region, and instance ID")
+			// if config.GcpProjectID == "" || config.GcpRegion == "" || config.GcpCloudSQLInstanceID == "" {
+			// 	return nil, errors.New("To use IAM auth with Google Cloud SQL, you must specify project ID, region, and instance ID")
```
ryanbooz, line 70: Because we can't totally distinguish CloudSQL from AlloyDB without a `systemType`, I'm not sure we can do much here. Without this check, we assume that if a user has added the `db_use_iam_auth` with either, they mean it and have followed the configuration instructions.

ryanbooz: Interestingly, i just saw this conversation when helping CME group on another issue. Maybe there is value in adding "alloydb" as a `systemType` for clarity in the UI as well?

[https://pganalyze.zendesk.com/agent/tickets/6569](https://pganalyze.zendesk.com/agent/tickets/6569)

runner/run.go
```diff
@@ -61,24 +65,46 @@ func Run(ctx context.Context, wg *sync.WaitGroup, opts state.CollectionOpts, log
...
+		if cfg.DbUseIamAuth && cfg.SystemType == "google_cloudsql" && cfg.GcpAlloyDBClusterID != "" && driverCleanupAlloyDb == nil {
+			driverCleanupAlloyDb, err = alloydb_pgxv5.RegisterDriver("alloydb-postgres", alloydbconn.WithIAMAuthN())
+
+			if err != nil {
+				logger.PrintError("Failed to register alloydb-postgres driver: %s", err)
+				return
+			}
+		}
+
+		if cfg.DbUseIamAuth && cfg.SystemType == "google_cloudsql" && driverCleanupPublicAlloyDb == nil {
```
ryanbooz, line 97:
```suggestion
		if cfg.DbUseIamAuth && cfg.SystemType == "google_cloudsql" && cfg.GcpAlloyDBClusterID != "" && driverCleanupPublicAlloyDb == nil {
```

lfittl (approved):
:+1:

input/postgres/establish_connection.go
```diff
@@ -66,20 +66,31 @@ func connectToDb(ctx context.Context, config config.ServerConfig, logger *util.L
 				dbPasswordOverride = dbToken
 			}
 		} else if config.SystemType == "google_cloudsql" {
-			if config.GcpProjectID == "" || config.GcpRegion == "" || config.GcpCloudSQLInstanceID == "" {
-				return nil, errors.New("To use IAM auth with Google Cloud SQL, you must specify project ID, region, and instance ID")
+			if config.GcpCloudSQLInstanceID != "" {
+				hostOverride = strings.Join([]string{config.GcpProjectID, config.GcpRegion, config.GcpCloudSQLInstanceID}, ":")
+			} else {
```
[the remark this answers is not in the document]
ryanbooz: Made a slight re-write so that we're not doing the same checks twice. Thanks @keiko713

PR #711, opened 2025-06-05
Allow sending otel payload in JSON format

### LINKED ISSUE #580: OTEL Collector: allow plain json payloads
I'm using vector.dev for collecting kubernetes logs and attempted to route cnpg cluster logs to pganalyze collector.

However, vector doesn't support proper OTLP sink, so initially I thought I could send plain otlp 'flavored' json as http payload, but pganalyze collector refused after failing to unmarshal protobuf.

rauanmayemir:
When user specifies `Content-Type: application/json`, handler will attempt to parse logsData from JSON instead of assuming it's a raw protobuf payload. Fixes #580.

This PR also includes a fix for panic error due to CNPG output not containing `"kubernetes"` section.
K8s pod/namespace information is not part of a record anyway as it only contains what CNPG container logs into stderr.

If we want to act on kubernetes metadata like k8s namespace, pod name, labels, it likely needs to be in `ResourceLogs[].Resource`.

@lfittl I would like to include native support for more kinds of logs from https://cloudnative-pg.io/documentation/1.26/logging/#other-logs, e.g `wal-archive` in particular. It would give great visibility into issues like failure to upload wal. Seeking guidance on how to approach this best.

image

### DIFF
--- a/input/system/selfhosted/otel_handler.go
+++ b/input/system/selfhosted/otel_handler.go
@@ -76,6 +76,10 @@ func skipDueToK8sFilter(kubernetes *common.KeyValueList, server *state.Server, p
 	var k8sPodName string
 	var k8sNamespaceName string
 
+	if kubernetes == nil {
+		return false
+	}
+
 	k8sLabels := make(map[string]string)
 	for _, rv := range kubernetes.Values {
 		if rv.Key == "pod_name" {


input/system/selfhosted/otel_handler.go
```diff
@@ -75,6 +76,10 @@ func skipDueToK8sFilter(kubernetes *common.KeyValueList, server *state.Server, p
 	var k8sPodName string
 	var k8sNamespaceName string
 
+	if kubernetes == nil {
+		return false
+	}
+
```
vfoucault, line 82: I'd personally keep that away form `skipDueToK8sFilter` which might be missleading, because this function is only about filtering out results and not assessing payload's contents.

input/system/selfhosted/otel_handler.go
```diff
@@ -115,9 +120,18 @@ func setupOtelHandler(ctx context.Context, server *state.Server, rawLogStream ch
 				prefixedLogger.PrintError("Could not read otel body")
 			}
 			logsData := &otlpLogs.LogsData{}
-			if err := proto.Unmarshal(b, logsData); err != nil {
-				prefixedLogger.PrintError("Could not unmarshal otel body")
+			contentType := r.Header.Get("Content-Type")
+			switch contentType {
+			case "application/json":
+				if err := protojson.Unmarshal(b, logsData); err != nil {
+					prefixedLogger.PrintError("Could not unmarshal otel body, json expected")
```
vfoucault, line 127: Couldn't we add here the `err.Error()` for clarity. In my experience it was hard to understand what was the issue without running locally the patched collector. 

Of course, protobuf's error `wire protocol error` wouldn't help much though.

vfoucault (approved):
🎉 I struggled to debug / make it work without touching the collector's code. Thanks for that. Once I've found your issue I was relieved that one's has already done it.

input/system/selfhosted/otel_handler.go
rauanmayemir: What do you mean? I'm only guarding against `kubernetes *common.KeyValueList` which could be nil, but the function doesn't seem to expect it to be nil so it panic crashes..

vfoucault: Something like

```go
if kubernetes != nil && skipDueToK8sFilter(kubernetes, server, prefixedLogger) {
	continue
}
```

This is just a style comment where to leave the `skipDueToK8sFilter` doing only what it is meant to be. No big deal here.

ryanbooz: I actually didn't end up including this in my open PR for other OTEL things, but I did originally have this and also prefer failing early here rather than inside of the function for what it's worth.

https://github.com/pganalyze/collector/pull/719/files#diff-9b41a22ed8cdaeb1a5377b0a3245b598b04071bfda5edd565de5707b5b5cef24R170

rauanmayemir: Should kubernetes then still be a reference instead of concrete type?

vfoucault:
Hi there, is there any news about this ?

input/system/selfhosted/otel_handler.go
lfittl: I agree with the idea to move this to where we call the function, I've applied that change in my rebased version in https://github.com/pganalyze/collector/pull/747

Regarding the use of a pointer/reference here - I think its better to keep it as a reference, to avoid a potential unnecessary copy of the struct.

input/system/selfhosted/otel_handler.go
lfittl: Agreed that its helpful to log more here - in the rebased version (https://github.com/pganalyze/collector/pull/747) I log both the error message and the original data that we received that we failed to unmarshal.

PR #714, opened 2025-06-13
Support newline inside of Query Text value

keiko713:
Auto EXPLAIN with Heroku + JSON format, the value of "Query Text" may contain \n (new line).
JSON's value can't contain \n, so replacing it with a white space.

### DIFF
--- a/logs/querysample/querysample.go
+++ b/logs/querysample/querysample.go
@@ -11,6 +11,7 @@ import (
 	"github.com/pganalyze/collector/logs/util"
 	"github.com/pganalyze/collector/output/pganalyze_collector"
 	"github.com/pganalyze/collector/state"
+	cUtil "github.com/pganalyze/collector/util"
 )
 
 func TransformAutoExplainToQuerySample(logLine state.LogLine, explainText string, queryRuntime string) (state.PostgresQuerySample, error) {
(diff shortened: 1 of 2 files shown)


logs/querysample/querysample.go
```diff
@@ -31,6 +32,21 @@ func TransformAutoExplainToQuerySample(logLine state.LogLine, explainText string
 func transformExplainJSONToQuerySample(logLine state.LogLine, explainText string, queryRuntimeMs float64) (state.PostgresQuerySample, error) {
 	var explainJSONOutput state.ExplainPlanContainer
 
+	// With Heroku + JSON format, the value of "Query Text" can contain \n (new line).
+	// JSON's value can't contain \n, so replacing it with a white space
+	if cUtil.IsHeroku() {
```
keiko713, line 37: We can technically expand this outside of Heroku too (I think we have seen this invalid JSON with the other cloud provider before too...), but for now, I'd like to minimize the change.

ryanbooz: Our of curiosity, what was the error you were seeing? This sounds close to an issue I had last week trying to get EXPLAIN plans to work with rsyslog and couldn't resolve ATM and wanted to revisit.

keiko713: It's: Invalid character '\n' in string literal

msakrejda (commented):
This looks reasonable, but I'm confused as to why it's necessary: the escapes here do not look specific to auto_explain logs. Are we missing this sort of transformation at a higher level?

logs/querysample/querysample.go
ryanbooz: Yeah - that was the same error I was getting with rsyslog messages and I couldn't get around it. 🤔

logs/querysample/querysample.go
```diff
@@ -31,6 +32,21 @@ func TransformAutoExplainToQuerySample(logLine state.LogLine, explainText string
...
+			submatches := re.FindStringSubmatch(m)
+			key := submatches[1]
+			val := submatches[2]
+			// Remove newlines and replace with space
+			cleaned := regexp.MustCompile(`[\r\n]+`).ReplaceAllString(val, " ")
+			return key + cleaned + `"`
+		})
+	}
+
 	if err := json.Unmarshal([]byte(explainText), &explainJSONOutput); err != nil {
```
lfittl, line 50: As discussed in PR review, it may be useful to see if we can find a JSON parser that is more lenient with newlines, as an alternate to the regexp-based approach.

keiko713: I looked around this (more lenient one is called "json5"), but none of json5 modules of golang is popular enough or widely used.
According to JSON RFC ([RFC 8259](https://datatracker.ietf.org/doc/html/rfc8259)), the control characters including newlines must be escaped, and also golang community is not really keen to support json5 in general according to [the forum post](https://groups.google.com/g/golang-nuts/c/mhj1D8HKZNU).

With that, I'd say that let's start with this.

msakrejda:
Right, okay. This does ring a bell now. So I think this is a reasonable workaround, but the problem does occur "lower" in the log pipeline, and we may need to revisit this at some point.

PR #716, opened 2025-06-25
Add more details to the release step

keiko713:
It's been a while since I last did the release and as I try to remember small bits of the release step, I figured that it'd be the best to document.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,14 +103,14 @@ time="2024-04-26T02:01:07Z" level=info msg="Generating README Documentation for 
 
 ### Release
 
-1. Create a PR to update the version numbers and CHANGELOG.md
+1. Create a PR to update the version numbers, CHANGELOG.md, and Helm Docs (see above)
 2. Once PR is merged, create a new tag `git tag v0.x.y`, then push it `git push origin v0.x.y`
 3. Once a new tag is pushed, GitHub Action Release will be kicked and create a new release (this will take about 2 hours, due to the package build and test)
 4. Modify the newly created release's description to match to CHANGELOG.md
 5. Release docker images using `make docker_release` (this requires access to the Quay.io push key, as well as "docker buildx" with QEMU emulation support, see below)
 6. Sign and release packages using `make -C packages repo` (this requires access to the Keybase GPG key)
 
-To run step 5 from an Ubuntu 22.04 VM, do the following:
+To run step 5 from an Ubuntu 24.04 VM, do the following (use the t3.large instance or higher):
 
 ```
 # Add Docker's official GPG key:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -103,14 +103,14 @@ time="2024-04-26T02:01:07Z" level=info msg="Generating README Documentation for
...
-1. Create a PR to update the version numbers and CHANGELOG.md
+1. Create a PR to update the version numbers, CHANGELOG.md, and Helm Docs (see above)
 2. Once PR is merged, create a new tag `git tag v0.x.y`, then push it `git push origin v0.x.y`
 3. Once a new tag is pushed, GitHub Action Release will be kicked and create a new release (this will take about 2 hours, due to the package build and test)
 4. Modify the newly created release's description to match to CHANGELOG.md
 5. Release docker images using `make docker_release` (this requires access to the Quay.io push key, as well as "docker buildx" with QEMU emulation support, see below)
 6. Sign and release packages using `make -C packages repo` (this requires access to the Keybase GPG key)
 
-To run step 5 from an Ubuntu 22.04 VM, do the following:
+To run step 5 from an Ubuntu 24.04 VM, do the following (use the c6i.xlarge instance or higher):
```
keiko713, line 113: Well, this is a great example of why we should document this - I just used t3.large and the CPU usage is pegged at 100% now, t3.large is definitely not enough and burstable performance is not preferred here. I do remember that I made this mistake before but I am still making the same mistake again.

keiko713: Well it still took ~17 mins with c6i.xlarge and CPU was busy, so we could say c6i.2xlarge here.

image

CONTRIBUTING.md
```diff
@@ -103,14 +103,14 @@ time="2024-04-26T02:01:07Z" level=info msg="Generating README Documentation for
 
 ### Release
 
-1. Create a PR to update the version numbers and CHANGELOG.md
+1. Create a PR to update the version numbers, CHANGELOG.md, and Helm Docs (see above)
```
keiko713, line 106: Note: we're very good at doing the Helm Docs regen part (which I also forgot with v0.66.2)

CONTRIBUTING.md
```diff
@@ -132,8 +132,9 @@ sudo apt-get install docker-ce docker-ce-cli containerd.io docker-buildx-plugin
 sudo apt update
 sudo apt install qemu-user-static binfmt-support make
 
-# Get these credentials from Quay.io
-sudo docker login -u="REPLACE_ME" -p="REPLACE_ME" quay.io
+# Get password (entered interactively) from from Quay.io
```
lfittl:
```suggestion
# Get password (entered interactively) from Quay.io
```

lfittl (approved):

CONTRIBUTING.md
keiko713: Note: this regenerates the README in https://github.com/pganalyze/collector/tree/main/contrib/helm/pganalyze-collector, especially the following part (which... you can see we've been forgetting to do so since 0.63.0 🙃 )
image

PR #717, opened 2025-06-27
Raise error when collector config has duplicate servers

seanlinsley:
Currently if the collector config has duplicate entries for the same server, we log out an error message but it's treated as a non-fatal issue: the collector skips duplicate servers and a test run succeeds. That can lead to users not realizing there's an issue with their config, so this change turns it into a fatal error that must be addressed.

This can be tested locally with:
```
make build && PGA_API_SYSTEM_SCOPE=test ./pganalyze-collector --test
```
Note that `PGA_API_SYSTEM_SCOPE=test` is needed because a self-hosted server has its first database name [appended to the system scope](https://github.com/pganalyze/collector/blob/b3ccc143d638008bd39301bf6651f1b9ab05dab6/config/identify_system.go#L111), so by default it's identified as a separate server. We could remove the database name from the system scope and move the current behavior to a fallback system scope so existing servers aren't churned, but there could be unexpected consequences to that kind of change so I've omitted it for now.

### DIFF
--- a/config/read.go
+++ b/config/read.go
@@ -3,6 +3,7 @@ package config
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
@@ -752,7 +753,7 @@ func preprocessConfig(config *ServerConfig) (*ServerConfig, error) {
 }
 
 // Read - Reads the configuration from the specified filename, or fall back to the default config
-func Read(logger *util.Logger, filename string) (Config, error) {
+func Read(testRun bool, logger *util.Logger, filename string) (Config, error) {
 	var conf Config
 	var err error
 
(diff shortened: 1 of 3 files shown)


runner/run.go
```diff
@@ -48,6 +48,10 @@ func Run(ctx context.Context, wg *sync.WaitGroup, opts state.CollectionOpts, log
 	if err != nil {
 		logger.PrintError("Config Error: %s", err)
 		keepRunning = !opts.TestRun && !opts.DiscoverLogLocation
+		if opts.TestRun {
+			testRunSuccess = make(chan bool, 1)
+			testRunSuccess <- false
+		}
```
seanlinsley, line 54: Without this change, config parse errors cause the process to hang indefinitely.

config/read.go
```diff
@@ -819,8 +819,7 @@ func Read(logger *util.Logger, filename string) (Config, error) {
 			// Ensure we have no duplicate identifiers within one collector
 			for _, server := range conf.Servers {
 				if config.Identifier == server.Identifier {
-					logger.PrintError("Skipping config section %s, detected as duplicate. Note: To monitor multiple databases on the same server, db_name accepts a comma-separated list.", config.SectionName)
-					continue
+					return conf, fmt.Errorf("%s and %s are duplicates. To monitor multiple databases on the same server, db_name accepts a comma-separated list.", server.SectionName, config.SectionName)
```
seanlinsley: Since we have both server names available, I've updated the error message to include them.

runner/run.go
keiko713: Is this only true for test run process?

lfittl: As discussed this is intentional to allow the background process to keep running after config errors.

(also, as we talked, there is a separate problem today that if you reload with a broken config, we don't use the old config, but instead stay in a failed state until the config is fixed and another reload is issued)

config/read.go
lfittl: On an unrelated note, as discussed, would be good to change this to only return early if its a test run (to not break existing installs that rely on this and have an automated collector update through the packages).

keiko713 (approved):

PR #719, opened 2025-07-08
Adds the ability to create multiple OTel endpoints on different ports for the same collector

ryanbooz:
Currently, the collector only supports one OTel endpoint for each running collector. The endpoint is also tied to a specific server configuration.

This allows creating multiple OTel V1 API endpoints to receive protobuf messages from Fluentbit in a Kubernetes environment. Multiple [OUTPUT] sections can be configured to send log data for specific server logs to the given endpoint which allows the collector to segregate and publish it with the appropriate server.

Additionally, I added the option to print out the Fluentbit payload if `--debug-logs` is enabled at runtime. The lingering issue with receiving logs correctly from some users comes down to how the `Tag` and `Match` properties of the Fluentbit configuration are set for [INPUT] and [OUTPUT]. Without being able to print out the actual JSON payload and examine for all of the expected properties, some of the code paths fail silently. The only current solution, however, is to ensure that the tags are configured correctly.

So, this breaks the Kubernetes plugin that adds metadata through an additional `kubernetes` node in the payload which users can filter on in the `skipDueToK8sFilter` function.

```
    [INPUT]
        Name tail
        Path /var/log/containers/cluster-example-*_default_postgres*.log
        Tag kube.cluster-example
        multiline.parser docker, cri
        DB /var/log/flb_kube1.db
```

The documentation implies that a user should be able to customize the tag and then add the `kubernetes` metadata using a [FILTER], but the auto expanding of the tag prefix proved unreliable in nearly every configuration.

```
  filters: |
    [FILTER]
        Name kubernetes
        Match kube.*
        Merge_Log On
        Keep_Log Off
        K8S-Logging.Parser On
        K8S-Logging.Exclude off
        Kube_Tag_Prefix kube.var.log.containers.
```

Without the `kubernetes` node doesn't exist in the payload, that fails. 

Therefore, the proper way, as far as I can find, is to set each [INPUT] to access the desired logs, and then filter different log [OUTPUT]s to the different OTel endpoints for that server.

```
    [OUTPUT]
        Name  opentelemetry
        Match kube.*cluster-example2-*
        Host  192.168.12.242
        Port  4319

    [OUTPUT]
        Name  opentelemetry
        Match kube.*cluster-example-*
        Host  192.168.12.242
        Port  4318
```

### DIFF
--- a/input/system/selfhosted/logs.go
+++ b/input/system/selfhosted/logs.go
@@ -138,7 +138,7 @@ func SetupOtelHandlerForServer(ctx context.Context, wg *sync.WaitGroup, opts sta
 	}
 
 	logStream := setupLogTransformer(ctx, wg, server, opts, logger, parsedLogStream)
-	return setupOtelHandler(ctx, server, logStream, parsedLogStream, logger)
+	return setupOtelHandler(ctx, server, logStream, parsedLogStream, logger, opts)
 }
 
 // SetupLogTails - Sets up continuously running log tails for all servers with a
--- a/input/system/selfhosted/otel_handler.go
+++ b/input/system/selfhosted/otel_handler.go
@@ -2,6 +2,7 @@ package selfhosted
 
 import (
 	"context"
+	"encoding/json"
 	"io"
 	"net/http"
 	"strconv"
@@ -14,6 +15,9 @@ import (
 	otlpLogs "go.opentelemetry.io/proto/otlp/logs/v1"
 	"google.golang.org/protobuf/proto"
 )
+
+// A list of all OpenTelemetry servers that we have started to prevent
+var otelServers []string
 
 // There currently are three kinds of log formats we aim to support here:
 //
@@ -67,7 +71,6 @@ func logLineFromJsonlog(record *common.KeyValueList, logParser state.LogParser) 
 		detailLine.LogLevel = pganalyze_collector.LogLineInformation_DETAIL
 		return logLine, &detailLine
 	}
-
 	return logLine, nil
 }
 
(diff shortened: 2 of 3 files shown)


keiko713:
Sorry it took forever for me to take a look at this, but finally getting to this! I have some question.

> Currently, the collector only supports one OTel endpoint for each running collector. The endpoint is also tied to a specific server configuration.

Uhmm, is "the collector only supports one OTel endpoint for each running collector" really true? I think you can specify otel endpoint per db server (though sounds like that's what you're saying in the second sentence so I'm confused), you just need to be careful not to specify the same port when you do so. For the current code, `setupOtelHandler` would run with each database server, and spin up the HTTP server for the otel endpoint.

I think your code is not changing this part that much really actually - rather, looks like it's adding the safeguard of potentially specifying the duplicated [db_log_otel_server](https://pganalyze.com/docs/collector/settings#self-managed-servers) with different db servers.

For example, you can do something like the following now:

```
[pganalyze]
api_key = your_pga_organization_api_key

[server1]
db_host = your_db_host
...
db_log_otel_server = 0.0.0.0:4318

[server2]
db_host = your_db_host2
...
db_log_otel_server = 0.0.0.0:4319
```

Though, when using our helm chart, it's gonna be a bit more complicated as we currently only support creating one service for this - meaning the collector can only expose one port that Fluent Bit can talk to. If you mean to address this part too, we need to tweak the Helm chart template and values too.
https://github.com/pganalyze/collector/blob/9a9afacdf60ff05b039d63071193c536da49e220/contrib/helm/pganalyze-collector/values.yaml#L70

input/system/selfhosted/otel_handler.go
```diff
@@ -105,73 +108,115 @@ func skipDueToK8sFilter(kubernetes *common.KeyValueList, server *state.Server, p
...
+			go func() {
+				err := http.ListenAndServe(otelLogServer, serverMux)
+				prefixedLogger.PrintInfo("Registered OpenTelemetry log handler on %s", otelLogServer)
+				if err != nil {
+					prefixedLogger.PrintError("Error starting server on %s: %v\n", otelLogServer, err)
+				}
+			}()
+			otelServers = append(otelServers, otelLogServer)
+		} else {
+			prefixedLogger.PrintInfo("OpenTelemetry log handler on %s already registered, skipping. Check your configuration for duplicate entries.", otelLogServer)
```
keiko713: It would be nice to mention the config name (`db_log_otel_server`) here. Uhmm I wasn't able to find an example, but maybe something like "for duplicate entries in db_log_otel_server"...?

input/system/selfhosted/otel_handler.go
```diff
@@ -105,73 +108,115 @@ func skipDueToK8sFilter(kubernetes *common.KeyValueList, server *state.Server, p
...
+	if otelLogServer != "" {
+		if !util.SliceContains(otelServers, otelLogServer) {
+			serverMux := http.NewServeMux()
+			serverMux.HandleFunc("/v1/logs", func(w http.ResponseWriter, r *http.Request) {
+				otelV1LogHandler(w, r, server, rawLogStream, parsedLogStream, prefixedLogger, opts)
+			})
+
+			go func() {
+				err := http.ListenAndServe(otelLogServer, serverMux)
+				prefixedLogger.PrintInfo("Registered OpenTelemetry log handler on %s", otelLogServer)
```
keiko713: Sooo this log line will only show up when the server exits or error happens, and I think you meant to show this when the collector starts. We actually already have the following log with the caller of `setupOtelHandler`:

https://github.com/pganalyze/collector/blob/9a9afacdf60ff05b039d63071193c536da49e220/input/system/selfhosted/logs.go#L136-L138

If you think it'd be nice to actually make this log always show up, maybe you can remove above part and bring this line above the `http.ListenAndServe`.

input/system/selfhosted/otel_handler.go
```diff
@@ -105,73 +108,115 @@ func skipDueToK8sFilter(kubernetes *common.KeyValueList, server *state.Server, p
...
+	/ available can quickly help us see if the payload matches our expectations
+	/ and reduce the time spent debugging.
+	/
+	/ Generally the issue seems to be that when a Fluentbit INPUT is configured
+	/ with a Tag other than the standard "kube.*", the additional kubernetes
+	/ metadata is not added to the log record, we then fail to get all of the
+	/ key value information. However, there have also been cases where the log
+	/ string values were formatted incorrectly and it couldn't be unmarshalled.
+	/
+	/ Being able to quickly inspect the raw payloads can help us identify issues
```
keiko713: I think this is really useful info but not sure if it belongs here.
It'd be actually nicer to have this in https://pganalyze.com/docs/log-insights/setup/opentelemetry, with a new section of "debugging" or "troubleshooting", then have the typical failing pattern and how to test it.

Actually, now I'm reading what `--debug-log` means, I think this usage is a bit off:

```
"Outputs all log analysis that would be sent, doesn't send any other data (use for debugging only)"
```

How about... uhmm do this with `opts.TestRunLogs`? Ugh, we don't really use this flag to do something like this either though, but might be more matching than `DebugLogs`. Thoughts?

ryanbooz: Totally fair. And no, we don't have a good option for doing something like this, but without the ability, some of these things become so challenging to debug with clients.

I can work with `TestRunLogs`. 😉

input/system/selfhosted/otel_handler.go
ryanbooz: It emits at startup as is. Might not be worth it, but it was helpful for me as I was moving this forward. Happy to just remove it.

image

input/system/selfhosted/otel_handler.go
ryanbooz: I've pulled out the comments and will update the docs. However, after looking at the options that we do have, I think `DebugLogs` is the best option at this point, at least given one other usage here: https://github.com/pganalyze/collector/blob/multi-otel-endpoints/runner/logs.go#L342

In this case, we're not testing that the logs work (generating a logging event and submitting it to pganalyze) and `TestRunLogs` may have worked just fine, I think. But rather (likely) seeing that the logs are not being submitted as expected and we need to debug the incoming log format.

If neither seem to fit, I guess we could try and think of something better and add it, but I think that feels a bit too much for this.

Any other opinions?

input/system/selfhosted/otel_handler.go
keiko713: Interesting, here is what I see when I test it with super simple setup, two servers, one with `db_log_otel_server = bad-address`, the other with `0.0.0.0:4319`.
Note that I added `prefixedLogger.PrintInfo("(before http.ListenAndServe) Registered OpenTelemetry log handler on %s", otelLogServer)` line before `http.ListenAndServe` to make it clear that part is showing up (but not "Registered" one).

```
$ ./pganalyze-collector
2025/07/22 19:39:20 W [server1] Checking log_line_prefix: database (%d) and user (%u) not found: pganalyze will not be able to correctly classify some log lines
2025/07/22 19:39:20 W [server2] Checking log_line_prefix: database (%d) and user (%u) not found: pganalyze will not be able to correctly classify some log lines
2025/07/22 19:39:20 I (before http.ListenAndServe) Registered OpenTelemetry log handler on bad-address
2025/07/22 19:39:20 I Registered OpenTelemetry log handler on bad-address
2025/07/22 19:39:20 E Error starting server on bad-address: listen tcp: address bad-address: missing port in address
2025/07/22 19:39:20 I (before http.ListenAndServe) Registered OpenTelemetry log handler on 0.0.0.0:4319
2025/07/22 19:40:00 I [server2] Submitted compact snapshots successfully: 3 activity
2025/07/22 19:40:00 I [server1] Submitted compact snapshots successfully: 3 activity
```

I also noticed that you have "Starting OpenTelemetry log handler on 0.0.0.0:4318" log line, which I don't think in the current code. Maybe you're seeing that log with some different version of your code?
`http.ListenAndServe` is a blocking call so it really shouldn't emit "Registered OpenTelemetry log handler..." log unless it errored like above `bad-adress` case 🤔 
https://www.reddit.com/r/golang/comments/inlgto/help_httplistenandserve_is_blocking_or_am_i_doing/

input/system/selfhosted/otel_handler.go
keiko713: > at least given one other usage here: https://github.com/pganalyze/collector/blob/multi-otel-endpoints/runner/logs.go#L342

I think this usage is exactly what the usage is describing, "Outputs all log analysis that would be sent" - so this is focused on debugging if whatever sending to the pganalyze making sense (e.g. if the log parsing is doing correctly, etc.).

What you want to debug here is also exactly what you mentioned, "debug the incoming log format". We actually have several places that can benefit of having this ability, like showing received logs from Azure Event Hub or Google Pub/Sub. From that point, introducing a new flag could be one option. The other option is, to update the description of `--debug-logs` and say something like the following:

```
Outputs all log analysis that would be sent, doesn't send any other data. For some providers, it also outputs incoming logs from the source (use for debugging only)
```

Sorry for all the back and forth, but I'd like to clarify other things a bit too.
In your comment (that you had previously):

> Generally the issue seems to be that when a Fluentbit INPUT is configured with a Tag other than the standard "kube.*", the additional kubernetes metadata is not added to the log record, we then fail to get all of the key value information.

You are trying to accomplish this with below two lines, correct?

```
prefixedLogger.PrintInfo("Received OpenTelemetry log data in the following format:\n")
prefixedLogger.PrintInfo(string(jsonData))
```

And for this comment:

> However, there have also been cases where the log string values were formatted incorrectly and it couldn't be unmarshalled.

This is about below?

```
jsonData, err := json.MarshalIndent(logsData, "", "  ")
if err != nil {
	prefixedLogger.PrintError("Failed to convert protobuf to JSON: %v", err)
}
```

For this one, can you please explain a bit more? Maybe you can share some examples? I'm especially curious what the difference between this marshall and the marshall above, `proto.Unmarshal(b, logsData)`.

ryanbooz: I like the option, currently, of updating the description for `debug-logs`. As we find more reasons to use it, maybe a new option will become clear.

ryanbooz:
Pulling this out to a separate comment, @keiko713.
---
> Sorry for all the back and forth, but I'd like to clarify other things a bit too. In your comment (that you had previously):
> 
> > Generally the issue seems to be that when a Fluentbit INPUT is configured with a Tag other than the standard "kube.*", the additional kubernetes metadata is not added to the log record, we then fail to get all of the key value information.
> 
> You are trying to accomplish this with below two lines, correct?
> 
> ```
> prefixedLogger.PrintInfo("Received OpenTelemetry log data in the following format:\n")
> prefixedLogger.PrintInfo(string(jsonData))
> ```
> 

So, yes. By putting the logging here, we are assuming that we did receive a payload and that it did unmarshal somehow into the `logsData` format (which is the Otel V1/logs format). But, even when that happens, a few recent customer issues were challenging to debug simply because we couldn't easily get the logs that were being sent (which don't look exactly like the lines in the container logs). I wanted to provide a "next step" opportunity to assist in debugging, but recognize that the process we go through here to identify errors at various stages could probably be improved some more after some additional experience. 

> And for this comment:
> 
> > However, there have also been cases where the log string values were formatted incorrectly and it couldn't be unmarshalled.
> 
> This is about below?
> 
> ```
> jsonData, err := json.MarshalIndent(logsData, "", "  ")
> if err != nil {
> 	prefixedLogger.PrintError("Failed to convert protobuf to JSON: %v", err)
> }
> ```
> 
> For this one, can you please explain a bit more? Maybe you can share some examples? I'm especially curious what the difference between this marshall and the marshall above, `proto.Unmarshal(b, logsData)`.

The first one existed previously and takes the protobuf message and maps it to the `logsData` structure to be used later. When that mapping works but the message doesn't contain everything we expect, it doesn't error, we just get an incomplete object as it were. Taking that object and converting it into JSON as something that is human readable is the reason for the Marshal. So, one converts the protobuf to an object which we use in the code, the other, if requested, takes the object and maps it into something readable.

Does that make more sense?

input/system/selfhosted/otel_handler.go
```diff
@@ -105,73 +108,103 @@ func skipDueToK8sFilter(kubernetes *common.KeyValueList, server *state.Server, p
...
+
+	if err != nil {
+		prefixedLogger.PrintError("Could not read otel body")
+	}
+	logsData := &otlpLogs.LogsData{}
+	if err := proto.Unmarshal(b, logsData); err != nil {
+		prefixedLogger.PrintError("Could not unmarshal otel body")
+	}
+
+	if opts.DebugLogs {
```
lfittl: Just talked about this a bit with @keiko713 and others in PR review.

It may make sense to introduce a new "verbose verbose" (`-vv`) mode in the collector that emits this information, whilst also emitting snapshots. The current `--debug-logs` flag isn't that useful for customer environments since you would have to stop the collector, start it with that flag on, and when the flag is on no snapshots get sent (i.e. customer would experience it as missing data).

input/system/selfhosted/otel_handler.go
```diff
@@ -15,6 +16,9 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+// A list of all OpenTelemetry servers that we have started to prevent
+var otelServers []string
```
lfittl: This may not work as expected because global state doesn't get reset on reloads - can we put this into a function somewhere (or a different state struct?)

input/system/selfhosted/otel_handler.go
ryanbooz: Sounds good. I'll add the option and implement here. Other locations can be dealt with separately as needed.

runner/logs.go
```diff
@@ -538,12 +538,7 @@ func testOtelLog(ctx context.Context, wg *sync.WaitGroup, server *state.Server,
 	logTestSucceeded := make(chan bool, 1)
 	parsedLogStream := setupLogStreamer(ctx, wg, opts, logger, []*state.Server{server}, logTestSucceeded, stream.LogTestCollectorIdentify)
 
-	err := selfhosted.SetupOtelHandlerForServer(ctx, wg, opts, logger, server, parsedLogStream)
```
keiko713, line 541: memo: I was wondering why this error was removed, but looks like this error was never really used (error was never really returned) anyways, so should be no effect of removing

input/system/selfhosted/logs.go
```diff
@@ -132,13 +132,32 @@ func SetupLogTailForServer(ctx context.Context, wg *sync.WaitGroup, opts state.C
...
+
+		prefixedLogger := logger.WithPrefix(server.Config.SectionName)
+
+		if util.SliceContains(otelServers, otelLogServer) {
+			prefixedLogger.PrintInfo("OpenTelemetry log handler on %s already registered, skipping. Check your configuration for duplicate 'db_log_otel_server' entries.", otelLogServer)
+			continue
+		}
+
+		if opts.DebugLogs || opts.TestRun {
+			logger.PrintInfo("Setting up OTLP HTTP server receiving logs with %s", server.Config.LogOtelServer)
```
keiko713, line 153: Uhmm, I'm still not so sure about having this line and below line (in otel_hanlder.go) separately.

```
prefixedLogger.PrintVerbose("Registered OpenTelemetry log handler on %s", otelLogServer)
```

How about removing here, then change above line to PrintInfo, and having the condition of printing it out with `opts.DebugLogs || opts.TestRun || opts.VeryVerbose`?
Or, I think it's also okay to simply change the log level to PrintInfo and log all the time too.

ryanbooz: Fair enough. I honestly didn't notice this preexisting message when we were refactoring the PR and added the verbose message as it seemed like we don't have consistency on what we log at startup - but I like your suggestion to keep it here.

PR #720, opened 2025-07-24
Support specifying multiple ports for log OTEL servers

keiko713:
Currently, when using a helm chart,  you can only expose one port for the [db_log_otel_server](https://pganalyze.com/docs/collector/settings#self-managed-servers).
With https://github.com/pganalyze/collector/pull/719, we should be able to create and expose more than one servers, and this helm chart update ensures that specifying to expose multiple ports of the collector is possible using a helm chart.

### DIFF
--- a/contrib/helm/pganalyze-collector/README.md
+++ b/contrib/helm/pganalyze-collector/README.md
@@ -1,6 +1,6 @@
 # pganalyze-collector
 
-![Version: 0.63.0](https://img.shields.io/badge/Version-0.63.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.63.0](https://img.shields.io/badge/AppVersion-v0.63.0-informational?style=flat-square)
+![Version: 0.66.2](https://img.shields.io/badge/Version-0.66.2-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.66.2](https://img.shields.io/badge/AppVersion-v0.66.2-informational?style=flat-square)
 
 pganalyze statistics collector
 
@@ -50,6 +50,7 @@ pganalyze statistics collector
 | service.create | bool | `false` | Specifies whether a service should be created for receiving logs via OpenTelemtry This service is used when Postgres is running within the cluster and Postgres logs are sent out to the collector using log collectors like Fluent Bit |
 | service.name | string | `"pganalyze-collector-otel-service"` | The name of the service to use. If not set and create is true, a name is generated using the fullname template. This is the name referenced by the log sender like Fluent Bit |
 | service.port | int | `4318` | The port of service. This is the port referenced by the log sender like Fluent Bit |
+| service.ports | list | `[{"name":"otel1","port":4318,"targetPort":4318},{"name":"otel2","port":4319,"targetPort":4319}]` | The list of port and target ports for OTEL logging When this is specified, above port and targetPort will be ignored. If you need to have multiple log OTEL servers, use this. |
 | service.targetPort | int | `4318` | The target port of the log OTEL server port. This should match to the port number specified with db_log_otel_server |
 | service.type | string | `"ClusterIP"` | The type of service to create. |
 | serviceAccount.annotations | object | `{}` | Annotations to add to the service account |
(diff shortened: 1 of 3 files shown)


contrib/helm/pganalyze-collector/README.md
```diff
@@ -50,6 +50,7 @@ pganalyze statistics collector
 | service.create | bool | `false` | Specifies whether a service should be created for receiving logs via OpenTelemtry This service is used when Postgres is running within the cluster and Postgres logs are sent out to the collector using log collectors like Fluent Bit |
 | service.name | string | `"pganalyze-collector-otel-service"` | The name of the service to use. If not set and create is true, a name is generated using the fullname template. This is the name referenced by the log sender like Fluent Bit |
 | service.port | int | `4318` | The port of service. This is the port referenced by the log sender like Fluent Bit |
+| service.ports | list | `[{"name":"otel1","port":4318,"targetPort":4318},{"name":"otel2","port":4319,"targetPort":4319}]` | The list of port and target ports for OTEL logging When this is specified, above port and targetPort will be ignored. If you need to have multiple log OTEL servers, use this. |
 | service.targetPort | int | `4318` | The target port of the log OTEL server port. This should match to the port number specified with db_log_otel_server |
```
ryanbooz, line 54: I wonder if there's a way to make this a little more clear. The `ports` is in between `port` and `targetPort` which need to be set together if only using one port. Should we just remove "port/targetPort" and force folks to use the `service.ports` instead? It can still be just one port, correct?

keiko713: Yeah I thought about having only `ports`, though I wanted to keep `port` + `targetPort` for the backward compatibility, as retiring `port` + `targetPort` will be a breaking change.
It doesn't mean that we can't make a breaking change, though 🤔

ryanbooz: Oh... good point. 

Since I'm not suggesting that we (necessarily) change the templates for now, they could still be backwards compatible, we'd just be pulling the "older" single port style out of documentation for anyone that comes along, right? That is, existing configs would still work if we leave your template changes as-is? (maybe I'm not thinking about this correctly)

keiko713: > we'd just be pulling the "older" single port style out of documentation for anyone that comes along

Do you mean our docs? (the one you created, esp. this part https://github.com/pganalyze/pganalyze-docs/blob/4155aaa49a2701bc3184125fcf38cc5610c93d77/log-insights/setup/opentelemetry.mdx#L72-L79)
If so, +1, I think that's totally fine to do.

For this README.md, it's auto-generated based on the values.yaml, so if we want to remove it from here, we'll need to remove it from values.yaml. It might be confusing to completely remove the description (e.g. I'd see the latest docs and like "what, why my setting is working even though it's not documented"), so I'd like to keep it somewhere.

lfittl: I discussed this with Keiko a bit just now, specifically re: whether we should deprecate `port/targetPort` and only recommend `ports` going forward in the `values.yml`.

I think the argument for keeping both styles (and not deprecating `port/targetPort`) is that when you're running the collector with a single server configured (as is often the case with Kubernetes deploys), it feels slightly better to me to keep the existing syntax, since we can document it without the (optional) name we document for the `ports` option. i.e. the documentation flows better if we say: If you have a single server with Otel integration, use `port/targetPort`, if you have multiple servers configured with the collector, use `ports` and give each port the name of the section you used in `CONFIG_CONTENTS` (to understand which Otel port maps to which monitored server).

lfittl: So consider this a +0.5 on keeping the PR as is (happy to discuss further), but I'd still shift the `service.ports` field to be after `service.targetPort` here.

keiko713: > I'd still shift the `service.ports` field to be after `service.targetPort` here.

Ahhhh, I wasn't sure what you were talking about previously, but sadly, this is auto-generated doc and the order is alphabetical, so it's not possible to tweak the order (unless I change `ports` to something like... `xports` - well that'll move down further below the `type`)

lfittl: Ah, that's unfortunate, but I guess we can't change that.

lfittl (approved):

PR #721, opened 2025-08-05
Use --very-verbose flag for logging received data from Azure EventHub, Google Pub/Sub, syslog

keiko713:
With https://github.com/pganalyze/collector/pull/719, we introduced a new flag `--very-verbose`, which enables emitting the incoming logs with Otel.
With this PR, expand the usage of `--very-verbose`, to enable the logging of incoming logs with Azure EventHub, Google Pub/Sub, and syslog.

See how the logs will look like in the following gist:
https://gist.github.com/keiko713/ebc0c8dbfcfd6ecb297fb1c1f3229e50

Internal memo: for https://github.com/pganalyze/pganalyze/issues/5073

### DIFF
--- a/input/system/google_cloudsql/logs.go
+++ b/input/system/google_cloudsql/logs.go
@@ -42,29 +42,30 @@ type LogStreamItem struct {
 	Content               string
 }
 
-func setupPubSubSubscriber(ctx context.Context, wg *sync.WaitGroup, logger *util.Logger, config config.ServerConfig, gcpLogStream chan LogStreamItem) error {
+func setupPubSubSubscriber(ctx context.Context, wg *sync.WaitGroup, logger *util.Logger, config config.ServerConfig, gcpLogStream chan LogStreamItem, opts state.CollectionOpts) error {
 	if strings.Count(config.GcpPubsubSubscription, "/") != 3 {
-		return fmt.Errorf("Unsupported subscription format - must be \"projects/PROJECT_NAME/subscriptions/SUBSCRIPTION_NAME\", got: %s", config.GcpPubsubSubscription)
+		return fmt.Errorf("unsupported subscription format - must be \"projects/PROJECT_NAME/subscriptions/SUBSCRIPTION_NAME\", got: %s", config.GcpPubsubSubscription)
 	}
 	idParts := strings.SplitN(config.GcpPubsubSubscription, "/", 4)
 	projectID := idParts[1]
 	subID := idParts[3]
 
-	var opts []option.ClientOption
+	var clientOpts []option.ClientOption
 	if config.GcpCredentialsFile != "" {
 		logger.PrintVerbose("Using GCP credentials file located at: %s", config.GcpCredentialsFile)
-		opts = append(opts, option.WithCredentialsFile(config.GcpCredentialsFile))
+		clientOpts = append(clientOpts, option.WithCredentialsFile(config.GcpCredentialsFile))
 	} else {
 		logger.PrintVerbose("No GCP credentials file provided; assuming GKE workload identity or VM-associated service account")
 	}
-	client, err := pubsub.NewClient(ctx, projectID, opts...)
+	client, err := pubsub.NewClient(ctx, projectID, clientOpts...)
 	if err != nil {
-		return fmt.Errorf("Failed to create Google PubSub client: %v", err)
+		return fmt.Errorf("failed to create Google PubSub client: %v", err)
 	}
 
 	sub := client.Subscription(subID)
+	wg.Add(1)
 	go func(ctx context.Context, wg *sync.WaitGroup, logger *util.Logger, sub *pubsub.Subscription) {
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -81,7 +81,7 @@ func main() {
 	flag.BoolVar(&reload, "reload", false, "Reloads the collector daemon that's running on the host")
 	flag.BoolVar(&noReload, "no-reload", false, "Disables automatic config reloading during a test run")
 	flag.BoolVarP(&logger.Verbose, "verbose", "v", false, "Outputs additional debugging information, use this if you're encountering errors or other problems")
-	flag.BoolVarP(&veryVerbose, "very-verbose", "", false, "Enable very verbose logging (will also enable verbose logging)")
+	flag.BoolVar(&veryVerbose, "very-verbose", false, "Enable very verbose logging (will also enable verbose logging)")
```
keiko713, line 84: I was trying to put `vv` as a shorthand, but looks like this module only support a single letter as a shorthand, so I tweaked this to use the simpler version `BoolVar` instead of `BoolVarP`

input/system/google_cloudsql/logs.go
```diff
@@ -78,7 +79,17 @@ func setupPubSubSubscriber(ctx context.Context, wg *sync.WaitGroup, logger *util
...
+				if opts.VeryVerbose {
+					jsonData, err := json.MarshalIndent(msg, "", "  ")
+					if err != nil {
+						logger.PrintVerbose("Failed to convert googleLogMessage struct to JSON: %v", err)
+					}
+					logger.PrintVerbose("Received Google Pub/Sub log data in the following format:\n")
+					logger.PrintVerbose(string(jsonData))
+				}
+
+				switch msg.Resource.ResourceType {
```
keiko713, line 91: In the VS Code, it was showing some warning here and I also found it easier to use with `switch`, so I updated here

input/system/google_cloudsql/logs.go
```diff
@@ -42,29 +42,30 @@ type LogStreamItem struct {
...
 	}
-	client, err := pubsub.NewClient(ctx, projectID, opts...)
+	client, err := pubsub.NewClient(ctx, projectID, clientOpts...)
 	if err != nil {
-		return fmt.Errorf("Failed to create Google PubSub client: %v", err)
+		return fmt.Errorf("failed to create Google PubSub client: %v", err)
 	}
 
 	sub := client.Subscription(subID)
+	wg.Add(1)
```
keiko713, line 66: This part was raising some warning in VS Code so I tweaked it.

input/system/selfhosted/syslog_handler.go
```diff
@@ -129,7 +140,7 @@ func setupSyslogHandler(ctx context.Context, config config.ServerConfig, out cha
 				// and disambiguate based on logParts["client"]
 			case <-ctx.Done():
 				server.Kill()
-				break
+				break outer
```
keiko713: VS Code was also complaining this, which makes sense - the break will only break from the `select`, not with `for`. Tweaked this to ensure that this correctly breaks.

lfittl: Alternatively I think a `return` statement could work here as well? (since we're in an anonymous function)

input/system/selfhosted/syslog_handler.go
```diff
@@ -86,9 +88,18 @@ func setupSyslogHandler(ctx context.Context, config config.ServerConfig, out cha
...
 
 	go func(ctx context.Context, server *syslog.Server, channel syslog.LogPartsChannel) {
+	outer:
 		for {
 			select {
 			case logParts := <-channel:
+				if opts.VeryVerbose {
+					jsonData, err := json.MarshalIndent(logParts, "", "  ")
+					if err != nil {
+						prefixedLogger.PrintVerbose("Failed to convert LogParts struct to JSON: %v", err)
```
lfittl, line 97: Minor detail, but wouldn't `logParts` be more consistent here with the other messages?

input/system/google_cloudsql/logs.go
lfittl: Good catch, that one was actually a (minor) concurrency bug, which could cause a caller to shut things down incorrectly since the wait group was theoretically be able to be completed before the Go routine was scheduled.

input/system/selfhosted/syslog_handler.go
keiko713: Sooo not really. I'm using the name of the struct here for other cases too. Here, the `logParts` is using the `format.LogParts`.

With google, the name of struct is `googleLogMessage` (and var name is `msg`), and with azure, it's `AzureEventHubData`.

... and looking into the otel case, looks like we're doing "Failed to convert protobuf to JSON" so it's not really consistent there 🙈 but I think it's indeed protobuf which is a bit different from other cases so I'd say that it's okay to leave it as is.

PR #724, opened 2025-08-13
Install staticcheck in CI, add staticcheck config

lfittl:


setup/main.go
```diff
@@ -63,7 +62,7 @@ func main() {
 		steps.ConfirmEmitTestExplain,
 	}
 
-	var setupState state.SetupState
+	var setupState s.SetupState
```
seanlinsley: I'd slightly prefer removing `s` and using `state` everywhere. Yes it's longer, but it makes it easier to figure out where the type is defined

lfittl: Yeah, I think that's reasonable, let me adjust.

lfittl: Done - I did end up renaming an existing `state` variable to `s` in most files, since it avoids the name conflict, and I think its reasonable to use short-hand names for variables.

staticcheck.conf
```diff
@@ -0,0 +1,8 @@
+checks = [
+    "all",
+    "-S1000", # Off because of https://github.com/dominikh/go-tools/issues/503
+    "-S1007", # Off because regular expression escaping is complicated, and its not worth forcing raw string literals
+    "-ST1005", # TODO: We should reformat our error strings, but this will require some effort
+    "-SA1019", # TODO: Investigate deprecation warnings
+    "-SA9003", "-ST1000", "-ST1003", "-ST1016", "-ST1020", "-ST1021", "-ST1022", "-ST1023" # Off in default config
+]
```
msakrejda, line 8: Nice

PR #725, opened 2025-08-14
Add Oracle Linux as a supported OS to the install script

keiko713:
Tested manually with the Oracle Linux 9 instance and it worked. Given that Oracle Linux is 100% application binary compatible with RHEL, we should be okay adding it to the supported OS list.


lfittl (approved):
:+1:

PR #728, opened 2025-09-19
GCP Pub/Sub: Allow multiple collectors subscribing to the same topic

lfittl:
This avoids different collectors stealing each others messages by only acknowledging a Pub/Sub message after it has been confirmed the collector is configured for that server, or it is a message that no collector would care about. Additionally, if a message is older than 1 hour, always acknowledge it.


input/system/google_cloudsql/logs.go
```diff
@@ -91,49 +88,92 @@ func setupPubSubSubscriber(ctx context.Context, wg *sync.WaitGroup, logger *util
...
+					var server *state.Server
+					for _, s := range servers {
+						if parts[0] == s.Config.GcpProjectID && parts[1] != "" && parts[1] == s.Config.GcpCloudSQLInstanceID {
+							server = s
+						}
+					}
+
+					if server == nil {
+						// Return message to be processed by a different collector if its recent.
+						if t.Before(time.Now().Add(time.Hour * -1)) {
```
lfittl: It might be worth considering making this dependent on a config setting and defaulting to "0" (ack right away if its not our message)

lfittl: Per PR review on 10/28, +1 on making this configurable and defaulting to 0 (always ack right away).

input/system/google_cloudsql/logs.go
```diff
@@ -34,22 +34,30 @@ type googleLogMessage struct {
...
 	if strings.Count(config.GcpPubsubSubscription, "/") != 3 {
 		return fmt.Errorf("unsupported subscription format - must be \"projects/PROJECT_NAME/subscriptions/SUBSCRIPTION_NAME\", got: %s", config.GcpPubsubSubscription)
 	}
 	idParts := strings.SplitN(config.GcpPubsubSubscription, "/", 4)
 	projectID := idParts[1]
 	subID := idParts[3]
 
+	var maxAge time.Duration
+	if config.GcpPubsubMaxAge != "" {
+		maxAge, err := time.ParseDuration(config.GcpPubsubMaxAge)
```
lfittl: This is the first setting where we're using [time.ParseDuration](https://pkg.go.dev/time#ParseDuration). I think this is better than calling the setting something like `gcp_pubsub_max_age_mins`, but happy to keep this as a simple integer too.

msakrejda: I think it's fine to use `ParseDuration`, but maybe we should be doing that in `config/read.go` rather than here? This seems like it should live in `preprocessConfig`.

lfittl: Yep, good point, moved it there.

PR #730, opened 2025-10-21
Avoid skipping query stats in case of slow query text/schema collection

lfittl:
Previously the high-frequency stats collection (which occurs every minute, except at the 10 minute mark) shared a mutex with the full snapshot collection, in order to record query stats. This effectively meant that query stats would be missed if the full snapshot took longer than 1 minute, which was observed in practice in the case of very slow query text file loading (up to 2 minutes by default, higher with changed settings), or when collecting schema data from many databases (up to 8 minutes total).

Instead, give the high frequency collection its own persisted state struct and mutex, and move data from that struct during the full snapshot. Additionally, to improve correctness of statistics, start each full snapshot by calling the exact same logic that otherwise runs once a minute. Perform the query text collection afterwards, without recording statistics at that time.

This effectively means that we are doing an additional pg_stat_statements access (without query text), but since that only looks at shared memory and holds the pgss lock very briefly, it should not be a problem, and avoids incorrect statistics when query text collection is slow (since pg_stat_statements looks at the statistics after accessing the query text file).

In passing, improve the correctness of the next query stats collection after a reset (it had the wrong reference point before), and add a missing mutex to the state file write (which now also writes out the separate high frequency state).

An existing short-coming is left for a future improvement (but with an added note), which is that state file updates only get written every 10 minutes, missing high-frequency updates inbetween in case of crashes.

### DIFF
--- a/runner/full_1min.go
+++ b/runner/full_1min.go
@@ -101,9 +101,9 @@ func Gather1minStatsFromAllServers(ctx context.Context, servers []*state.Server,
 	// (which holds the state mutex that the state file write also wants to acquire). That means
 	// in case of collector crashes we may have an incorrect reference point on a subsequent start.
 	//
-	// XXX: Can we address this by using a read/write mutex that is held in read mode during the
-	// collection, and only elevated to write (i.e. blocking other readers) once we swap out the
-	// state stored on the server struct?
+	// We could potentially address this by using a read/write mutex that is held in read mode
+	// during the collection, and only elevated to write (i.e. blocking other readers) once we swap
+	// out the state stored on the server struct.
 
 	wg.Wait()
 }


input/full.go
```diff
@@ -35,18 +35,41 @@ func CollectFull(ctx context.Context, server *state.Server, connection *sql.DB,
...
+	newHighFreqState, err := CollectAndDiff1minStats(ctx, c, connection, ps.CollectedAt, server.HighFreqPrevState)
+	if err != nil {
+		logger.PrintError("Could not collect high frequency statistics for server: %s", err)
+		err = nil
+	} else {
+		// Move over previously collected high frequency stats to be submitted with this full snapshot
+		//
+		// We don't want to delay this longer in case query text or schema metadata collection takes time.
+		// Doing this later could mean that additional high frequency query stats are being collected and
+		// included in the submission, whose query texts were not yet picked up with this snapshot.
```
seanlinsley: This context makes sense when reviewing the PR, but I think it raises unnecessary questions when reading the code in the future. So I'd think this comment should be removed or at least slimmed down, because it's obvious that in the `err == nil` case we'd want to store the stats that were just generated.

input/full.go
```diff
@@ -35,18 +35,41 @@ func CollectFull(ctx context.Context, server *state.Server, connection *sql.DB,
...
+		ts.ServerIoStats = newHighFreqState.QueuedServerIoStats
+		newHighFreqState.UnidentifiedStatementStats = make(state.HistoricStatementStatsMap)
+		newHighFreqState.UnidentifiedPlanStats = make(state.HistoricPlanStatsMap)
+		newHighFreqState.QueuedServerIoStats = make(state.HistoricPostgresServerIoStatsMap)
+		server.HighFreqPrevState = newHighFreqState
+	}
+	server.HighFreqStateMutex.Unlock()
 	bufferCacheReady := make(chan state.BufferCache)
 	go func() {
 		postgres.GetBufferCache(ctx, c, server, opts, bufferCacheReady)
```
seanlinsley: This should be kept at the top of the function (before `CollectAndDiff1minStats`) to reduce delays in submitting the snapshot since buffer cache collection can take a long time.

input/full.go
```diff
@@ -35,18 +35,41 @@ func CollectFull(ctx context.Context, server *state.Server, connection *sql.DB,
...
 	}()
 
-	ps.LastStatementStatsAt = time.Now()
 	err = postgres.SetQueryTextStatementTimeout(ctx, connection, logger, server)
 	if err != nil {
 		logger.PrintError("Error setting query text timeout: %s", err)
 		return
 	}
-	ts.Statements, ts.StatementTexts, ps.StatementStats, err = postgres.GetStatements(ctx, c, connection, true)
+	ts.Statements, ts.StatementTexts, err = postgres.GetStatementTexts(ctx, c, connection)
```
seanlinsley, line 101: I wonder if this should be moved to a goroutine that's spun up at the start of the snapshot, similar to `GetBufferCache`, so other work can be done while waiting on Postgres to read the query text file. One downside is that we'd potentially use more memory that way, since other collection tasks could be performed while `GetStatementTexts` has received the text and is doing work on it.

seanlinsley (commented):
This looks good overall, though more people should probably review it. It could also make sense to deploy this branch to production before merging so we can confirm it works as expected on a busy system.

input/full.go
lfittl: Yeah, agreed, I actually have a related change locally that moved it to the very top, but somehow when splitting apart the diff this got moved too far down. Will fix.

input/postgres/statements.go
```diff
@@ -75,16 +77,175 @@ func ResetStatements(ctx context.Context, c *Collection, db *sql.DB) error {
...
+		select {
+		// Since normalizing can take time, explicitly check for cancellations
+		case <-ctx.Done():
+			return nil, nil, ctx.Err()
+		default:
+			fingerprintAndNormalize(c, key, query, statements, statementTextsByFp, ignoreIoTiming)
+		}
+	}
+
+	c.SelfTest.MarkCollectionAspectOk(state.CollectionAspectPgStatStatements)
```
msakrejda, line 238: Both this and the above are marking the same collection aspect: I thought that could be a problem, but it looks like the aspect status bookkeeping accounts for that.

input/postgres/statements.go
```diff
@@ -116,35 +277,15 @@ func GetStatements(ctx context.Context, c *Collection, db *sql.DB, showtext bool
...
 		c.SelfTest.MarkCollectionAspectError(state.CollectionAspectPgStatStatements, "extension version not supported (1.%d installed, 1.3+ supported)", foundExtMinorVersion)
-		return nil, nil, nil, fmt.Errorf("pg_stat_statements extension not supported (1.%d installed, 1.3+ supported). To update run `ALTER EXTENSION pg_stat_statements UPDATE`", foundExtMinorVersion)
+		return "", 0, fmt.Errorf("pg_stat_statements extension not supported (1.%d installed, 1.3+ supported). To update run `ALTER EXTENSION pg_stat_statements UPDATE`", foundExtMinorVersion)
 	}
 
 	if c.GlobalOpts.TestRun && foundExtMinorVersion < extMinorVersion {
 		pgssMsg := fmt.Sprintf("extension outdated (1.%d installed, 1.%d available)", foundExtMinorVersion, extMinorVersion)
 		c.Logger.PrintInfo("pg_stat_statements %s. To update run `ALTER EXTENSION pg_stat_statements UPDATE`", pgssMsg)
-		if extMinorVersion >= 9 {
+		if extMinorVersion >= 9 && foundExtMinorVersion < 9 {
```
msakrejda, line 283: Why this additional check? [This check](https://github.com/pganalyze/collector/pull/730/files#diff-80669d27bdbdad6405bf3819eef438baa8a7ba60c75f1e4ef2acb0107fb3a702R285) on (new) line 285 should cover this, no?

input/full.go
seanlinsley: Oh, some customers might not want us to use another database connection. Instead we could have a single goroutine that calls both `GetBufferCache` and `GetStatementTexts`. But running those two expensive tasks serially might take snapshot generation slower. We could feasibly get into a situation where everything else has been collected, and then we stall waiting for these two to finish.

seanlinsley: Another option would be to have three goroutines: buffer cache, statement texts, and everything else. A max of two goroutines would actually run in parallel, to reduce database connection usage. Though I'm not sure how to implement that in Go.

input/postgres/statements.go
keiko713: I was wondering why we have two places to say similar things (here and a new line 277), and admittedly, I'm the one who added both of them:
https://github.com/pganalyze/collector/pull/472
https://github.com/pganalyze/collector/pull/613

I'd propose to drop this if statement entirely actually, and move the comment up to the new line 277 - below is just logging at the error level during the test run, and line 277 is causing the test to mark as an error state - so line 277 one is a stronger message that we should keep.

input/postgres/server_stats.go
```diff
@@ -111,10 +111,6 @@ func GetServerStats(ctx context.Context, c *Collection, db *sql.DB, ps state.Per
 	if err != nil {
 		return ps, ts, err
 	}
-	ps.ServerIoStats, err = GetPgStatIo(ctx, c, db)
```
keiko713, line 114: Memo: I was wondering what this is - it's introduced back in March in https://github.com/pganalyze/collector/pull/553

This is one of the stats that we collect once a min too (hence makes sense to move it to input/full_1min.go), and looks like we are not using in the app side yet.

runner/full_1min.go
```diff
@@ -140,5 +96,14 @@ func Gather1minStatsFromAllServers(ctx context.Context, servers []*state.Server,
 		}(servers[idx])
 	}
 
+	// TODO: We currently do not write out the state file here, since we do not want to block
+	// subsequent high-frequency stats collections in case the full snapshot is still running
+	// (which holds the state mutex that the state file write also wants to acquire). That means
+	// in case of collector crashes we may have an incorrect reference point on a subsequent start.
+	//
+	// XXX: Can we address this by using a read/write mutex that is held in read mode during the
```
keiko713: Is this... a memo for the future iteration?

keiko713 (commented):
Overall, looking good to me. +1 on Sean re: trying this out by ourself first

input/postgres/server_stats.go
lfittl: Yeah, this is the `pg_stat_io` data - we'll put this to use soon :)

runner/full_1min.go
lfittl: Yeah, basically I was just documenting my understanding in code, because this is a problem, just not a new one. I can drop this too if this is confusing (or at least the `XXX:` part?)

input/full.go
lfittl: Whilst I agree making query text collection be parallel with other tasks is worth exploring (if we can figure out the connection use), I don't think we need to tie it to this change - the additional query stats collection introduced in this PR should be very fast, and we already had the schema collection wait on query text collection previously.

However, whilst considering this, I realized that we may gain a minor correctness improvement by gathering some more statistics before we capture the query texts (basically capturing metrics earlier is good, so they stay closer to the reported `collected_at` timestamp, and don't get skewed by slow query texts). I'll move up the replication/server/backend counts/system stats accordingly.

input/full.go
lfittl: The point of the comment is to explain why this is not happening in the runner code, where it would be more natural to perform (since its not really an "input" activity per se). I'll shorten it a bit, but I think its helpful for clarity.

input/postgres/statements.go
lfittl: Yeah, I think its okay to keep this the same aspect, since they should not diverge (they're reporting on the same internal state, just twice).

input/full.go
seanlinsley: I'm not concerned that the additional query stats collection will cause a slowdown, but we know query text collection is already a bottleneck. But yes it makes sense for that to be a separate PR.

input/postgres/statements.go
lfittl: Yeah, I reworked this a bit but kept the logger calls (vs only relying on the aspect/summary reporting), since I'm slightly worried some people only look at the log output.

runner/full_1min.go
lfittl: I've reworded the last sentence here to drop the XXX prefix, but otherwise kept it in place per discussion on Slack that was supportive of keeping this note.

PR #731, opened 2025-11-07
Full snapshot: Improve buffer cache connection handling, run system state collection in Go routine

lfittl:


input/full.go
```diff
@@ -29,6 +29,24 @@ import (
 func CollectFull(ctx context.Context, server *state.Server, connection *sql.DB, opts state.CollectionOpts, logger *util.Logger) (ps state.PersistedState, ts state.TransientState, err error) {
 	ps.CollectedAt = time.Now()
 
+	bufferCacheReady := make(chan state.BufferCache)
+	go func() {
+		if server.Config.MaxBufferCacheMonitoringGB > 0 {
```
lfittl, line 34: This is technically a new condition, but I think its okay to add, and avoids an unnecessary database connection if buffer cache collection is turned off.

input/full.go
```diff
@@ -29,6 +29,24 @@ import (
...
+			bufferCacheReady <- postgres.GetBufferCache(ctx, server, logger, opts)
```
keiko713, line 35: This is indeed better than the current one (as the right side will be evaluated first and `db.Close()` will happen), but this still can cause a goroutine leak I think.
There is no other logic after this, but if there was, that logic will be never processed until `bufferCacheReady` is read (which never happens if cancelled). Well, also this goroutine will keep alive waiting for a channel to be read, meaning GC can't reclaim the bufferCache value returned from `postgres.GetBufferCache`, can cause the memory leak too.

As we know that there will at most one value send to the channel, I think it'd be safer to create a buffered channel with size 1 (aka `bufferCacheReady := make(chan state.BufferCache, 1)`). The same for below `systemStateReady` channel.
In that way, goroutine will finish (aka no leaking) after writing the result to the buffer (don't wait for to be read), then the buffer will be GC-ed when it becomes unreachable (aka this `CollectFull` func is completed).

PR #734, opened 2025-11-07
Require WebSocket connection to pganalyze server by default

seanlinsley:
Ahead of deprecating the legacy HTTP API, this PR requires that the collector use the WebSocket API by default. If users run into connection issues they can temporarily restore the old behavior by setting `--require-websocket=false` while we look into the source of the issue.

In passing, this changes the WebSocket handler to attempt reconnection more often, while also reconnecting less often if the server responds with a 4xx status code (e.g. 401 unauthorized or 429 rate limit).

### DIFF
--- a/runner/websocket.go
+++ b/runner/websocket.go
@@ -84,7 +84,9 @@ func connect(ctx context.Context, server *state.Server, opts state.CollectionOpt
 	}
 	headers["User-Agent"] = []string{util.CollectorNameAndVersion}
 	conn, response, err := dialer.DialContext(connCtx, url.String(), headers)
-	connectStatus = response.StatusCode
+	if response != nil {
+		connectStatus = response.StatusCode
+	}
 	if err != nil {
 		cancelConn()
 		logger.PrintWarning("Error starting websocket: %s %v", err, response)


main.go
```diff
@@ -109,6 +110,7 @@ func main() {
 	flag.StringVar(&stateFilename, "statefile", defaultStateFile, "Specify alternative path for state file")
 	flag.StringVar(&pidFilename, "pidfile", "", "Specifies a path that a pidfile should be written to (default is no pidfile being written)")
 	flag.BoolVar(&benchmark, "benchmark", false, "Runs collector in benchmark mode (skip submitting the statistics to the server)")
+	flag.BoolVar(&requireWebsocket, "require-websocket", true, "Require WebSocket connection to the pganalyze server")
```
lfittl, line 113: Maybe:

```suggestion
	flag.BoolVar(&requireWebsocket, "require-websocket", true, "Require WebSocket connection to the pganalyze server (don't allow HTTP fallback)")
```

lfittl: Also I wonder if it would be better to invert the meaning, so its `--allow-http-fallback`. Having a boolean flag that takes a value (with the default being true) feels a bit unusual to me.

output/full.go
```diff
@@ -85,6 +86,8 @@ func submitFull(ctx context.Context, s snapshot.FullSnapshot, server *state.Serv
 		server.SnapshotStream <- compressedData.Bytes()
 		logger.PrintInfo("Submitted full snapshot successfully")
 		return nil
+	} else if collectionOpts.RequireWebsocket {
```
lfittl, line 89: One problem with this change as-is is that we are more likely to throw away snapshots if there is an interim networking problem (i.e. imagine a spurious TCP reset that causes the websocket to close, that may take up to 60 seconds to reconnect), where before we would have fallen back to HTTP in that situation.

I feel like we ideally need a dedicated retry mechanism for Websocket connections that models what we're doing for HTTP since #190 - i.e. keep retrying for a little bit without losing the upload.

However, for now, maybe we could do something simpler, that is, before we throw the snapshot away, force at least one reconnect on the websocket (if its not up).

If you'd like I can give inventing a mechanism here a try - I think the interaction with the existing retry loop makes that a bit finicky.

seanlinsley: >before we throw the snapshot away, force at least one reconnect on the websocket

I could see the argument for doing that for full snapshots, but am wary of retrying on activity snapshots. It would also require restructuring the code so that a snapshot runner can itself trigger a reconnect. Right now that's handled out of band and the snapshot runner simply checks to see if a websocket is currently active.

One way to implement that without restructuring the code might be to track the timestamp of the last successfully submitted snapshot. If it was within the last 5 minutes, shorten the reconnection delay to 10 seconds so we wouldn't be likely to lose more than a single activity snapshot.

lfittl: > I could see the argument for doing that for full snapshots, but am wary of retrying on activity snapshots. It would also require restructuring the code so that a snapshot runner can itself trigger a reconnect. Right now that's handled out of band and the snapshot runner simply checks to see if a websocket is currently active.

Yeah, that's fair - I don't think losing an activity snapshot is a big deal, but losing a full snapshot is much more noticeable.

> One way to implement that without restructuring the code might be to track the timestamp of the last successfully submitted snapshot. If it was within the last 5 minutes, shorten the reconnection delay to 10 seconds so we wouldn't be likely to lose more than a single activity snapshot.

Hmm, yeah, that could help. Alternatively, I wonder if we should behave differently when the connect fails, vs when our connection errors unexpectedly (in the reader, if I followed the code correctly). Maybe its reasonable to retry a new connect right away if the reader loop exists due to an error.

seanlinsley: The read and write loops probably deserve the same retry priority, unless I'm missing something. And I'd hesitate to retry immediately because there's the potential it could get caught in a retry loop, putting extra load on the API.

It could make sense to delay retries when the initial connection fails, since things like network configuration or invalid auth tokens make the initial connection less likely to be successful on retry. But I think that scenario is more or less handled by my previous suggestion, of reconnecting sooner if a snapshot was recently submitted.

lfittl:
@seanlinsley I've evolved my thinking on this PR a bit since last we spoke, here are my current thoughts:

First, I think we should make this a collector config setting, not just a command line argument. Command line arguments are actually pretty frustrating to modify (e.g. with systemd you need to modify the unit file directly, make sure its synced up, etc).

Second, I'm wondering whether we should start with making this an error in collector test run only to start with, and only default to requiring it for regular starts in a later release. If someone auto-updates the collector (typical with package-based installs, or when using the `stable` tag for the container) the update may break the reporting of data, and since we don't notify about that issue today, that could be quite frustrating when noticing it late. A test command is always interactive with higher likelihood that someone notices it.

PR #736, opened 2025-11-28
Rework legacy HTTP API and WebSocket handling, fix incorrect pausing of collectors when using skip_if_replica

lfittl:
See individual commits.

Intended to support the work done in #734 (and takes the reconnect interval changes from there) as well as fix a current bug related to the `skip_if_replica` flag (see last commit message).

### DIFF
--- a/runner/activity.go
+++ b/runner/activity.go
@@ -28,18 +28,15 @@ func processActivityForServer(ctx context.Context, server *state.Server, opts st
 	}
 
 	if server.Config.SkipIfReplica {
+		// This connection gets reused for the activity snapshot collection later
 		connection, err = postgres.EstablishConnection(ctx, server, logger, opts, "")
 		if err != nil {
 			return newState, false, errors.Wrap(err, "failed to connect to database")
 		}
 		defer connection.Close()
-		var isReplica bool
-		isReplica, err = postgres.GetIsReplica(ctx, logger, connection)
+		err = checkReplicaCollectionDisabledWithConn(ctx, server, logger, connection)
 		if err != nil {
 			return newState, false, err
-		}
-		if isReplica {
-			return newState, false, state.ErrReplicaCollectionDisabled
 		}
 	}
 
@@ -58,6 +55,7 @@ func processActivityForServer(ctx context.Context, server *state.Server, opts st
 		}
 		return newState, false, nil
 	}
+
 	// N.B.: Without the SkipIfReplica flag, we wait to establish the connection to avoid opening
 	// and closing it for no reason when the grant EnableActivity is not set (e.g., production plans)
 	if connection == nil {
@@ -133,25 +131,12 @@ func CollectActivityFromAllServers(ctx context.Context, servers []*state.Server,
 				server.ActivityStateMutex.Unlock()
 				server.SelfTest.MarkCollectionAspectError(state.CollectionAspectActivity, "%s", err.Error())
 
-				server.CollectionStatusMutex.Lock()
-				isIgnoredReplica := err == state.ErrReplicaCollectionDisabled
-				if isIgnoredReplica {
-					reason := err.Error()
-					server.CollectionStatus = state.CollectionStatus{
-						CollectionDisabled:        true,
-						CollectionDisabledReason:  reason,
-						LogSnapshotDisabled:       true,
-						LogSnapshotDisabledReason: reason,
-					}
-				}
-				server.CollectionStatusMutex.Unlock()
-
-				if isIgnoredReplica {
+				if err == state.ErrReplicaCollectionDisabled {
 					prefixedLogger.PrintVerbose("All monitoring suspended while server is replica")
 				} else {
 					allSuccessful = false
 					prefixedLogger.PrintError("Could not collect activity for server: %s", err)
-					if !isIgnoredReplica && server.Config.ErrorCallback != "" {
+					if server.Config.ErrorCallback != "" {
 						go runCompletionCallback("error", server.Config.ErrorCallback, server.Config.SectionName, "activity", err, prefixedLogger)
 					}
 				}
--- a/runner/util.go
+++ b/runner/util.go
@@ -1,0 +1,59 @@
+package runner
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/pganalyze/collector/input/postgres"
+	"github.com/pganalyze/collector/state"
+	"github.com/pganalyze/collector/util"
+)
+
+func checkReplicaCollectionDisabled(ctx context.Context, server *state.Server, opts state.CollectionOpts, logger *util.Logger) error {
+	if !server.Config.SkipIfReplica {
+		return nil
+	}
+
+	connection, err := postgres.EstablishConnection(ctx, server, logger, opts, "")
+	if err != nil {
+		return fmt.Errorf("Failed to connect to database: %s", err)
+	}
+	defer connection.Close()
+
+	return checkReplicaCollectionDisabledWithConn(ctx, server, logger, connection)
+}
+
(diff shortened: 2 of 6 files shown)


output/grant.go
```diff
@@ -4,59 +4,100 @@ import (
...
-func GetGrant(ctx context.Context, server *state.Server, opts state.CollectionOpts, logger *util.Logger) (state.Grant, error) {
-	grant := state.Grant{Config: pganalyze_collector.ServerMessage_Config{Features: &pganalyze_collector.ServerMessage_Features{}}}
-	req, err := http.NewRequestWithContext(ctx, "GET", server.Config.APIBaseURL+"/v2/snapshots/grant", nil)
+// EnsureGrant - Ensures the server has a valid grant stored from either WebSocket or HTTP-based grant API
+func EnsureGrant(ctx context.Context, server *state.Server, opts state.CollectionOpts, reuseGrant bool) error {
+	if opts.ForceEmptyGrant {
+		return nil
+	}
+
+	// Accept existing grant data if we're using a WebSocket connection and its
```
msakrejda:
```suggestion
	// Accept existing grant data if we're using a WebSocket connection and it's
```

output/upload.go
```diff
@@ -2,100 +2,108 @@ package output
...
+							if i < len(keys)-1 {
+								details += ", "
+							}
+						}
+						if len(details) > 0 {
+							logger.PrintInfo("Submitted compact snapshots successfully: " + details)
+						}
+						compactLogTime = time.Now().Truncate(time.Minute)
+						compactLogStats = make(map[string]uint8)
+					}
```
msakrejda: Would it make sense to extract this to a helper? It's a lot of code inline for just logging.

output/upload_http_legacy.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+
+	if resp.StatusCode != http.StatusCreated {
+		return "", fmt.Errorf("Bad S3 upload return code %s (should be 201 Created), body: %s", resp.Status, body)
```
msakrejda:
```suggestion
		return "", fmt.Errorf("Bad S3 upload return code %s (expected 201 Created), body: %s", resp.Status, body)
```

runner/activity.go
```diff
@@ -29,42 +28,34 @@ func processActivityForServer(ctx context.Context, server *state.Server, opts st
...
 	if server.Config.SkipIfReplica {
+		// This connection gets reused for the activity snapshot collection later
 		connection, err = postgres.EstablishConnection(ctx, server, logger, opts, "")
 		if err != nil {
 			return newState, false, errors.Wrap(err, "failed to connect to database")
 		}
 		defer connection.Close()
-		var isReplica bool
-		isReplica, err = postgres.GetIsReplica(ctx, logger, connection)
+		err = checkReplicaCollectionDisabledWithConn(ctx, server, logger, connection)
```
msakrejda, line 37: Are we only using the connection we established above for this check? Should we just call the variant without `WithConn` here?

runner/util.go
```diff
@@ -0,0 +1,51 @@
...
+
+func checkReplicaCollectionDisabledWithConn(ctx context.Context, server *state.Server, logger *util.Logger, connection *sql.DB) error {
+	isReplica, err := postgres.GetIsReplica(ctx, logger, connection)
+	if err != nil {
+		return fmt.Errorf("Error checking replication status")
+	}
+	if isReplica {
+		// Shut down websocket so another server that has become primary can collect without
+		// being considered the older collector (which could cause the server to pause it)
+		server.WebSocket.Disconnect()
```
msakrejda, line 35: I assume this is idempotent?

util/reconnecting_socket.go
```diff
@@ -0,0 +1,174 @@
...
+			case <-w.start:
+				if !w.Connected() && w.requested.Load() && time.Now().After(skipConnectUntil) {
+					connectStatus, err := w.connect(ctx)
+					if connectStatus >= 400 && connectStatus < 500 {
+						skipConnectUntil = time.Now().Add(clientErrorTimeout) // Delay reconnect when server responds with 4xx errors
+					}
+					w.startWait <- err
+				} else {
+					w.startWait <- nil
+				}
```
msakrejda: So if we call `Connect()` but we've had a 4xx error, `Connect()` will return a nil error but not actually connect. Is that right? What does that mean for the snapshot flow?

util/reconnecting_socket.go
```diff
@@ -0,0 +1,174 @@
...
+			}
+		}
+	}()
+	// Reader goroutine
+	go func() {
+		for {
+			_, data, err := conn.ReadMessage()
+			if err != nil {
+				serverClosed := websocket.IsCloseError(err, websocket.CloseNoStatusReceived) // The server shutdown the websocket
```
msakrejda:
```suggestion
				serverClosed := websocket.IsCloseError(err, websocket.CloseNoStatusReceived) // The server shut down the websocket
```

util/reconnecting_socket_test.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	time.Sleep(2 * time.Second)
+
+	cancel()
+
+	if atomic.LoadInt32(&connectsReceived) < 2 {
+		t.Errorf("TestSocketReconnect: expected: 2+ connects; actual: %d", connectsReceived)
+	}
+}
```
msakrejda, line 68: Nice.

msakrejda (approved):
A few questions and minor comments, but looks great. Channels are a nice solution here.

runner/util.go
msakrejda: Ah yes, I saw later that it is.

output/upload.go
lfittl: Yeah, that's fair. Its existing code that I moved around, but not opposed to refactoring it into a helper.

util/reconnecting_socket.go
lfittl: From a snapshot flow perspective we should fall back to HTTP snapshots ([reference](https://github.com/pganalyze/collector/blob/rework-upload-grant-websockets/output/upload.go#L98)), because the websocket is not connected.

The one thing that may not fully work correctly (when looking at this again), is that the grant code that checks the return value of `Connect()` will fall through to the Grant.Valid check ([reference](https://github.com/pganalyze/collector/blob/rework-upload-grant-websockets/output/grant.go#L36)). That may accept an outdated HTTP grant that does not have a valid upload location.

Maybe we should just special case the situation where we're skipping connecting because of the past error, and return an error in `Connect()` in that case too?

msakrejda: Hmm. I think that might be clearer, yeah.

runner/activity.go
lfittl: The whole reason its two separate methods is because for activity snapshots we want to reuse the connection for the actual snapshot (matters for activity snapshots because they're so frequent, and otherwise we'd open two connections), see comment a few lines above.

output/upload.go
lfittl: I've put parts of this logic in a helper - feedback welcome whether that's a good balance.

util/reconnecting_socket.go
lfittl: I've implemented that with a new ErrorConnectRateLimited error.

runner/activity.go
msakrejda: Right, okay. I missed the comment and was not looking at the collapsed sections. This is a very long method (but that's not this patch's problem).

output/upload.go
msakrejda: Nice, that's much clearer.

PR #739, opened 2025-12-02
Add support for capturing plan statistics with pg_stat_plans

lfittl:


input/postgres/plans.go
```diff
@@ -25,39 +25,70 @@ FROM
...
 
-	computePlanIdEnabled, err := GetPostgresSetting(ctx, db, "aurora_compute_plan_id")
-	if err != nil {
-		if c.GlobalOpts.TestRun {
-			c.Logger.PrintInfo("Function aurora_stat_plans() is not supported because Aurora version is too old. Upgrade to Aurora PostgreSQL version 14.10, 15.5, or later versions to collect query plans and stats.")
+	if c.PostgresVersion.IsAwsAurora {
+		computePlanIdEnabled, err := GetPostgresSetting(ctx, db, "aurora_compute_plan_id")
+		if err != nil {
+			if c.GlobalOpts.TestRun {
+				c.Logger.PrintInfo("Function aurora_stat_plans() is not supported because Aurora version is too old. Upgrade to Aurora PostgreSQL version 14.10, 15.5, or later versions to collect query plans and stats.")
```
msakrejda, line 56: I know this is existing code, but is it wise to log that the version is too old when encountering _any_ error here?

input/postgres/plans.go
```diff
@@ -25,39 +25,70 @@ FROM
...
+			ioTimeFields = statementSQLIoTimeFieldsDefault
+		}
+		querySql = fmt.Sprintf(auroraPlanSQL, ioTimeFields, showtext)
 	} else {
-		ioTimeFields = statementSQLIoTimeFieldsDefault
+		var extSchema string
+		err = db.QueryRowContext(ctx, QueryMarkerSQL+plansExtensionSQL).Scan(&extSchema)
+		if err == sql.ErrNoRows {
+			// In case the extension does not exist we return without an error here - because pg_stat_plans
+			// is not widely available yet, we don't want to confuse the test output.
```
msakrejda, line 83: Should we print something in test runs, like we do for Aurora?

msakrejda (approved):

input/postgres/plans.go
lfittl: I think the situation on a non-Aurora install is different because we don't know if the `pg_stat_plans` extension is even installed or not. Whilst we could theoretically suggest to install the extension if its a self-managed system, in practice this does not seem like a good idea, because the extension is relatively new and we'd have to make it conditional on Postgres 16+, and not using self-managed whilst actually being on a managed service provider.

I was considering adding something if the extension is added to `shared_preload_libraries` but not yet created on the database, because that would be an easy fix - but I think for now its okay to not add that and come back to this.

input/postgres/plans.go
lfittl: Yeah, fair point, reworked that a bit.

PR #740, opened 2025-12-02
Disable buffer cache collection for Aurora Serverless

msakrejda:
We've had a report of segfaults caused by scanning the buffer cache
during Aurora Serverless scaling.

Disable buffer cache collection for Serverless instances to avoid
causing instability with the collector's monitoring.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,6 @@ jobs:
 
     - name: Lint (go fmt)
       run: |
-        go fmt ./...
         test $(go fmt ./... | wc -l) -eq 0
 
     - name: Lint (staticcheck)


msakrejda:
I'm just starting to test this, but it's otherwise ready for review.

One question I had: should we allow users to opt out of this behavior with something like `aws_unsafe_force_serverless_buffer_cache_collection`? My feeling is no until someone complains, but I don't feel strongly and have not given it that much thought.

input/full.go
```diff
@@ -29,21 +29,27 @@ import (
 func CollectFull(ctx context.Context, server *state.Server, connection *sql.DB, opts state.CollectionOpts, logger *util.Logger) (ps state.PersistedState, ts state.TransientState, err error) {
 	ps.CollectedAt = time.Now()
 
+	// These were previously run in separate goroutines, but because buffer cache checks
+	// during Aurora Serverless scaling can cause segfaults, we now run them in the same
+	// goroutine so that we can recognize Serverless nodes from the API (there does not
+	// seem to be a way to check whether Aurora is Serverless in SQL).
```
lfittl: I'd reword this a bit, I don't think the "previously ..." style makes sense for comments (I'd usually only use that in commit messages), but I'd instead describe why the code works the way it does. Maybe:

```suggestion
    // Buffer cache statistics and system information are technically separate, but there is
    // an edge case we care about: We intentionally turn off use of pg_buffercache data
    // when Amazon Aurora Serverless is in use, because it currently has a bug that causes
    // a server crash during a scaling event, if pg_buffercache is queried concurrently.
```

msakrejda:
```suggestion
    // Buffer cache statistics and system information are separate, but querying pg_buffercache during
    // an Amazon Aurora Serverless scaling event can cause a server crash (which seems to be due to
    // a Serverless bug), so we first get system information and skip the buffer cache if necessary.
```
Yeah, good point. Reworded a bit to note that the system information check is to determine if we should skip the buffer cache scan (and for conciseness).

msakrejda:
Manual testing confirms (we'll need to make it clearer what's going on on the app side, but that's separate).

msakrejda:
Huh. I don't get that lint failure locally. Looking.

msakrejda:
CI seems to pass with additional debugging output but fails if I remove that.

PR #741, opened 2025-12-03
Use consistent pattern for temp files; prune stray files on startup

msakrejda:


main.go
```diff
@@ -369,3 +372,45 @@ func Reload(logger *util.Logger) {
...
+		if !strings.HasPrefix(name, "pga_collector") {
+			continue
+		}
+		fi, err := entry.Info()
+		if err != nil {
+			logger.PrintWarning("Could not check info for file %s in temp dir %s: %s\n", name, tmpdir, err)
+			continue
+		}
+		if stat, ok := fi.Sys().(*syscall.Stat_t); ok {
+			if stat.Uid == uid32 {
```
msakrejda: Not sure if we want to keep the ownership check, or just rely on the name pattern...

config/read.go
```diff
@@ -545,7 +545,7 @@ func CreateOTelTracingProvider(ctx context.Context, conf ServerConfig) (*sdktrac
 }
 
 func writeValueToTempfile(value string) (string, error) {
-	file, err := os.CreateTemp("", "")
+	file, err := os.CreateTemp("", "pga_collector")
```
lfittl: Is there a benefit to keeping this shorter? Otherwise I'd argue for using the full "pganalyze_collector" for clarity.

main.go
```diff
@@ -369,3 +372,45 @@ func Reload(logger *util.Logger) {
...
+	uid32 := uint32(uid)
+	tmpdir := os.TempDir()
+	entries, err := os.ReadDir(tmpdir)
+	if err != nil {
+		logger.PrintWarning("Could not open temp directory to prune temp files: %s\n", err)
+		return
+	}
+	for _, entry := range entries {
+		name := entry.Name()
+		if !strings.HasPrefix(name, "pga_collector") {
```
lfittl: Should we put "pga_collector" name (or possibly "pganalyze_collector") in a constant somewhere? (maybe move this utility + the constant into a file in `util/`, so it can be referenced from input/config?)

main.go
```diff
@@ -279,6 +280,8 @@ func main() {
 	sigs := make(chan os.Signal, 1)
 	signal.Notify(sigs, os.Interrupt, syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP)
 
+	pruneTempFiles(logger)
```
lfittl: We should probably not run this in test runs, so a test run can't delete a file that the collector running in the background still uses.

main.go
lfittl: Hmhm. How does this work with temp files - can a different user who is not root delete someone else's temp file?

With the pattern matching it feels safe-ish to not accidentally delete the wrong files, but if we can be sure we can only delete pganalyze user owned files anyway because of permissions (in typical scenarios) that'd make me sure we can skip the ownership check.

config/read.go
msakrejda: Okay.

main.go
msakrejda: Sure.

main.go
msakrejda: I think standard UNIX permissions apply. On my machine:
```
maciek@whimbrel:~$ touch /tmp/foo
maciek@whimbrel:~$ ls -l /tmp/foo
-rw-rw-r-- 1 maciek maciek 0 Dec  3 17:11 /tmp/foo
```
That file is owned by me but is group-writeable, so anyone in the `maciek` group can delete it.

With the additional check here we're only checking that the file was actually created by the collector user, not just that it has permissions to delete it.

lfittl: Given that I would say we should be okay without the permission check.

lfittl (approved):
:+1: Works as expected!

PR #745, opened 2025-12-08
Add new configuration setting to require WebSockets, default off for now

lfittl:
This introduces a new setting "api_require_websocket" / "API_REQUIRE_WEBSOCKET" that requires the Snapshot API connection to go over the newer WebSocket API instead of the legacy HTTPS API.

The WebSocket API is geared towards long-lived connections (lower overhead) and allows the server side to push down requests, like for running EXPLAIN on-demand as part of the opt-in collector workflow for Workbooks.

We are currently planning to change the default value of this to "on" in a future collector release. Currently this defaults to "off" which is the same behaviour as before (try WebSocket, but allow HTTPS fallback).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -241,6 +241,9 @@ type ServerConfig struct {
 	HTTPClient          *http.Client
 	HTTPClientWithRetry *http.Client
 
+	// Require WebSocket connection (if false, will try WebSocket but fall back to legacy HTTPS API)
+	APIRequireWebsocket bool `ini:"api_require_websocket"`
+
 	// WebSocket URL to be used for API WebSocket connection
 	WebSocketUrl string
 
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -241,6 +241,9 @@ type ServerConfig struct {
 	HTTPClient          *http.Client
 	HTTPClientWithRetry *http.Client
 
+	// Require WebSocket connection (if false, will try WebSocket but fall back to legacy HTTPS API)
+	APIRequireWebsocket bool `ini:"api_require_websocket"`
```
lfittl, line 245: I've confirmed in a quick test that in order to flip the default we can simply update `getDefaultConfig` to have this set to `true`.

PR #747, opened 2025-12-09
OTel Log Server: Support receiving JSON log format in addition to Protobuf

lfittl:
Rebased version of https://github.com/pganalyze/collector/pull/711 with an additional fix in a separate commit.

### DIFF
--- a/input/system/selfhosted/otel_handler.go
+++ b/input/system/selfhosted/otel_handler.go
@@ -191,7 +191,6 @@ func otelV1LogHandler(w http.ResponseWriter, r *http.Request, server *state.Serv
 			}
 		}
 	}
-	return
 }
 
 func setupOtelHandler(ctx context.Context, server *state.Server, rawLogStream chan<- SelfHostedLogStreamItem, parsedLogStream chan state.ParsedLogStreamItem, prefixedLogger *util.Logger, opts state.CollectionOpts) {


input/system/selfhosted/otel_handler.go
```diff
@@ -105,16 +106,28 @@ func skipDueToK8sFilter(kubernetes *common.KeyValueList, server *state.Server, p
 	return false
 }
 
-func otelV1LogHandler(w http.ResponseWriter, r *http.Request, server *state.Server, rawLogStream chan<- SelfHostedLogStreamItem, parsedLogStream chan state.ParsedLogStreamItem, prefixedLogger *util.Logger, opts state.CollectionOpts) http.ResponseWriter {
```
lfittl, line 108: The `http.ResponseWriter` return here was unused (and not clear why we did that?), and I added a bunch more early returns so figured I can remove it at the same time.

keiko713: sooo +1 for removing `w` from the return value of this func, but I think technically we should do the following two things:

* when everything is going well (aka at the end of this func), we should technically write 204 to response writer, like `w.WriteHeader(http.StatusNoContent)`
* when things didn't go well (aka some early return pattern), we should technically write 4xx to response writer

The OTEL client is sending something to us, so it's not really nice of us for not... responding. I think introducing `w.WriteHeader(http.StatusNoContent)` doesn't hurt, but returning 4xx may introduce some unnecessary retries on OTEL client side, so I don't think we need to do that without more testing.

lfittl: Ah, that's a good point. In the interest of getting the release shipped, let me address that in a follow-up PR.

lfittl: It seems the OTLP spec is actually pretty specific here: https://github.com/open-telemetry/opentelemetry-specification/blob/84a5fa3d1ef42caae8f99f1b431504d7cdb9fffe/specification/protocol/otlp.md#otlphttp-response

Cleaned this up in a follow-up PR: #751

PR #749, opened 2025-12-11
feat: add support for PlanetScale logs

mattrobenolt:
👋 

I would like to toss this up to solicit feedback and see how we can work together on this a tad!

So right now, this is mostly written as a best effort, "hope it works".

Right now, I'm blocked on 2 things:

- [x] The API to get the signed URL for fetching logs from our logs service is an internal API. (This is on us, we will expose this.)
- [x] I can't fully end to end test this with pganalyze since the collector, for the best of my knowledge, requires a pganalyze API key. I'd like to be able to see the logs fully on PgAnalyze side and make sure things are working as intended.

I have tested this implementation from _our_ side, basically the code in `input/system/planetscale/logs.go`, but writing a little `pscale_logs` program that uses this to just fetch and tail logs from us.

Is there a better way we can work together on getting this shipped? We can likely figure out other better integration points, but logging support was the first obvious step that was missing.

cc @lfittl :)

### DIFF
--- a/runner/logs.go
+++ b/runner/logs.go
@@ -27,10 +27,8 @@ import (
 	"github.com/pkg/errors"
 )
 
-const (
-	LogDownloadInterval  time.Duration = 30 * time.Second
-	LogStreamingInterval time.Duration = 10 * time.Second
-)
+const LogDownloadInterval time.Duration = 30 * time.Second
+const LogStreamingInterval time.Duration = 10 * time.Second
 
 // SetupLogCollection - Starts streaming or scheduled downloads for logs of the specified servers
 func SetupLogCollection(ctx context.Context, wg *sync.WaitGroup, servers []*state.Server, opts state.CollectionOpts, logger *util.Logger, hasAnyHeroku bool, hasAnyGoogleCloudSQL bool, hasAnyAzureDatabase bool, hasAnyTembo bool) {
@@ -147,12 +145,9 @@ func downloadLogsForServerWithLocksAndCallbacks(ctx context.Context, wg *sync.Wa
 }
 
 func downloadLogsForServer(ctx context.Context, server *state.Server, opts state.CollectionOpts, logger *util.Logger) (state.PersistedLogState, bool, error) {
-	var err error
-	if !opts.ForceEmptyGrant && !opts.DebugLogs {
-		err = output.EnsureGrant(ctx, server, opts, logger, false)
-		if err != nil || !server.Grant.Load().ValidConfig {
-			return server.LogPrevState, false, err
-		}
+	err := output.EnsureGrant(ctx, server, opts, logger, false)
+	if err != nil || !server.Grant.Load().ValidConfig {
+		return server.LogPrevState, false, err
 	}
 	transientLogState := state.TransientLogState{CollectedAt: time.Now()}
 


mattrobenolt:
@lfittl it's the `planetscale` org. If you can extend the trial, I can get this tested with some real data.

runner/logs.go
```diff
@@ -145,9 +147,11 @@ func downloadLogsForServerWithLocksAndCallbacks(ctx context.Context, wg *sync.Wa
...
 
 func downloadLogsForServer(ctx context.Context, server *state.Server, opts state.CollectionOpts, logger *util.Logger) (state.PersistedLogState, bool, error) {
-	err := output.EnsureGrant(ctx, server, opts, logger, false)
-	if err != nil || !server.Grant.Load().ValidConfig {
-		return server.LogPrevState, false, err
+	if !opts.ForceEmptyGrant {
+		err := output.EnsureGrant(ctx, server, opts, logger, false)
+		if err != nil || !server.Grant.Load().ValidConfig {
+			return server.LogPrevState, false, err
+		}
```
lfittl: I assume you needed this for testing since you didn't have an API key to use - it seems this change might make sense in general to enable `--dry-run-logs` to work correctly, but we'll need to add a `var err error` line above, to make the code below continue to work as expected:


```suggestion
    var err error
	if !opts.ForceEmptyGrant {
		err = output.EnsureGrant(ctx, server, opts, logger, false)
		if err != nil || !server.Grant.Load().ValidConfig {
			return server.LogPrevState, false, err
		}
```

input/system/planetscale/logs.go
```diff
@@ -0,0 +1,373 @@
...
+	if e.Body != "" {
+		return fmt.Sprintf("HTTP %d: %s", e.StatusCode, e.Body)
+	}
+	return fmt.Sprintf("HTTP %d", e.StatusCode)
+}
+
+type authCacheKey = struct{ org, db, branch string }
+
+// serverAuthCache maps server identifiers to their cached auth data
+var serverAuthCache = map[authCacheKey]cachedAuth{}
```
lfittl: We try to avoid globals that keep state in the collector, since its supposed to be able to handle complete config changes via SIGHUP, including removing servers (e.g. someone could remove all Planetscale servers, and we don't want to leak this then).

Could we move this under the `state.Server` struct? We might even be able to not have a key for it (i.e. just the cachedAuth), since the config (`authCacheKey`) won't change without a reload.

config/identify_system.go
```diff
@@ -90,7 +90,7 @@ func identifySystem(config ServerConfig) (systemID string, systemType string, sy
 	} else if (config.PlanetScaleOrg != "") || systemType == "planetscale" {
 		systemType = "planetscale"
 		if systemID == "" {
-			systemID = config.PlanetScaleDatabase + "/" + config.PlanetScaleBranch
+			systemID = config.PlanetScaleOrg + "/" + config.PlanetScaleDatabase + "/" + config.PlanetScaleBranch
```
lfittl, line 93: As discussed on Slack, I would leave this as `database / branch` - because we have the org already in the scope, its not needed here.

Alternatively, we could leave the scope empty and do this here, but that would be different than we do for other providers (where we put the account ID/etc in the scope and leave it out from the ID) - but open to that if y'all think that makes more sense.

In practice the only difference here is the default display name in the pganalyze server list (it defaults to the ID without the scope). That display name can always be changed via the UI later.

input/system/planetscale/system.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// GetSystemState returns system information for a PlanetScale Postgres instance.
+// PlanetScale doesn't expose detailed system metrics via this API, so we return
+// minimal information.
+func GetSystemState(ctx context.Context, server *state.Server, logger *util.Logger) state.SystemState {
+	system := state.SystemState{}
+	system.Info.Type = state.PlanetScaleSystem
+	system.Info.SystemID = server.Config.SystemID
+	system.Info.SystemScope = server.Config.SystemScope
```
lfittl, line 17: Setting SystemID/SystemScope here is technically not necessary since the calling general purpuse `GetSystemState` function will set them after it calls into the provider specific logic.

config/identify_system.go
lfittl: Per Slack discussion, you preferred to keep the org name in the ID - I think its the same from a pganalyze perspective, so I'm good to do it as you have it here.

runner/logs.go
```diff
@@ -145,9 +147,12 @@ func downloadLogsForServerWithLocksAndCallbacks(ctx context.Context, wg *sync.Wa
 }
 
 func downloadLogsForServer(ctx context.Context, server *state.Server, opts state.CollectionOpts, logger *util.Logger) (state.PersistedLogState, bool, error) {
-	err := output.EnsureGrant(ctx, server, opts, logger, false)
-	if err != nil || !server.Grant.Load().ValidConfig {
-		return server.LogPrevState, false, err
+	var err error
+	if !opts.ForceEmptyGrant && !opts.DebugLogs {
```
lfittl: Looking at this again, `EnsureGrant` already checks for ForceEmptyGrant, so that should not be necessary.

For DebugLogs, I think its better if we just always have that do ForceEmptyGrant - we can do that here: https://github.com/pganalyze/collector/blob/main/main.go#L176

lfittl (approved):
Looks good, thanks for working on this! 🥳

PR #751, opened 2025-12-13
Rework OTel logs handler to be spec compliant, refactor for clarity 

lfittl:
Plus some minor prep work to help with that, see individual commits.

Follow-up from #747.


util/http_server.go
```diff
@@ -0,0 +1,47 @@
...
+func GoServeHTTP(ctx context.Context, logger *Logger, addr string, serveMux *http.ServeMux) {
+	s := &http.Server{
+		BaseContext: func(net.Listener) context.Context { return ctx },
+		Addr:        addr,
+		Handler:     serveMux,
+	}
+	lc := net.ListenConfig{}
+	l, err := lc.Listen(ctx, "tcp", s.Addr)
+	if err != nil {
+		logger.PrintError("Error starting HTTP server on %s: %v\n", addr, err)
```
keiko713, line 19: One thing it's slightly missing from the previous version is, this error message is too generic and sometimes it'll be hard to tell which part didn't go well. I like that we have this util func so I think it's an acceptable tradeoff, also I can't come up with a better alternative without making things messy, but just figured to mention.

input/system/selfhosted/otel_handler.go
```diff
@@ -3,28 +3,178 @@ package selfhosted
...
+func handleOtlpLogsRequestProtobuf(b []byte, server *state.Server, rawLogStream chan<- SelfHostedLogStreamItem, parsedLogStream chan state.ParsedLogStreamItem, prefixedLogger *util.Logger, veryVerbose bool) (resp []byte, err error) {
+	logsData := &otlpLogs.LogsData{}
+	if err = proto.Unmarshal(b, logsData); err != nil {
+		prefixedLogger.PrintError("OTel log server could not unmarshal request body, expected binary OTLP Protobuf format: %s", err)
+		err = fmt.Errorf("Could not unmarshal Protobuf request body")
+		return
+	}
+
+	if veryVerbose {
+		jsonData, err := json.Marshal(logsData)
```
keiko713: Looks like you dropped the indent thingy (`json.MarshalIndent(logsData, "", "  ")`), why? Maybe you thought that it'll be too many lines...?

input/system/selfhosted/otel_handler.go
```diff
@@ -3,28 +3,178 @@ package selfhosted
...
+	logsData := &otlpLogs.LogsData{}
+	if err = protojson.Unmarshal(b, logsData); err != nil {
+		prefixedLogger.PrintError("OTel log server could not unmarshal request body, JSON does not match expected format: %s\n  received body: %s", err, string(b))
+		err = fmt.Errorf("Could not unmarshal JSON request body")
+		return
+	}
+
+	if veryVerbose {
+		prefixedLogger.PrintVerbose("OTel log server received JSON log data in the following format:\n")
+		prefixedLogger.PrintVerbose(string(b))
```
keiko713: A bit similar to above, this was previously pretty JSON but now it's a compact one. Is this intentional?

input/system/selfhosted/otel_handler.go
```diff
@@ -3,28 +3,178 @@ package selfhosted
...
+		if err != nil {
+			prefixedLogger.PrintError("OTel log server could not read request body: %s", err)
+			w.WriteHeader(http.StatusBadRequest)
+			w.Write([]byte("Could not read request body"))
+			return
+		}
+
+		var resp []byte
+		switch r.Header.Get("Content-Type") {
+		case "application/x-protobuf":
```
keiko713, line 46: I think you'll also want to specify the content type for the write, which maybe you can do here by doing something like the following?:

```
w.Header().Set("Content-Type", "application/x-protobuf")
```

input/system/selfhosted/otel_handler.go
```diff
@@ -3,28 +3,178 @@ package selfhosted
...
+			resp, err = handleOtlpLogsRequestProtobuf(b, server, rawLogStream, parsedLogStream, prefixedLogger, opts.VeryVerbose)
+		case "application/json":
+			resp, err = handleOtlpLogsRequestJson(b, server, rawLogStream, parsedLogStream, prefixedLogger, opts.VeryVerbose)
+		default:
+			err = fmt.Errorf("Unsupported Content-Type, must be application/x-protobuf or application/json")
+		}
+
+		if err != nil {
+			w.WriteHeader(http.StatusBadRequest)
+			w.Write([]byte(err.Error()))
```
keiko713: According to the specification, we should technically:

* use [Status](https://pkg.go.dev/google.golang.org/genproto/googleapis/rpc/status#Status) as a reply
  * we can omit code, but will need message
* then marshall Status with either `proto` or `protojson`, depending on which content type we received

https://opentelemetry.io/docs/specs/otlp/#failures-1

keiko713 (commented):
Some small comments but overall, looking good!

input/system/selfhosted/otel_handler.go
lfittl: So I've actually found the pretty-printed version harder to work with in practice - to actually analyze it you'd typically want to copy it out, and the pretty printing makes that harder to do.

That said, I think I'll revert to pretty printing, since it seems that's what we're doing for the other `--very-verbose` cases, and its better to be consistent here. I can also see the pretty printing working better when we're on a customer call to review a problem.

util/http_server.go
lfittl: Yeah, I thought about that, but short of adding a special field or error callback, this is hard to improve (because one of the two errors happens in the Go routine).

lfittl:
@keiko713 This is ready for re-review! (your feedback has been addressed, and I also added a test for some additional coverage of the logic)

PR #753, opened 2025-12-19
Add PlanetScale server and system type, config settings

lfittl:
Takes the non-log related changes from #749 and does the necessary protobuf changes for the system type. Also adds collector info for the significant config information.


protobuf/full_snapshot.proto
```diff
@@ -374,6 +374,10 @@ message CollectorConfig {
   bool azure_has_ad_certificate = 59;
   string azure_subscription_id = 60;
 
+  string planetscale_org = 64;
+  string planetscale_database = 65;
+  string planetscale_branch = 66;
```
lfittl, line 379: We got ourselves a bit in a tight situation here with the numbering, but I think fitting this in between Azure and GCP makes most sense.

Matt added some additional settings over in https://github.com/pganalyze/collector/pull/749/changes#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R135, but I don't think we need to report those to the server (definitely wouldn't want to do it for the API credentials).

keiko713 (approved):

PR #787, opened 2026-03-12
Clean up unused log state

seanlinsley:
Logs text has been submitted inline with log snapshots (as opposed to uploaded separately to S3) since version 0.60.0, but there was some leftover state that this PR removes.


protobuf/compact_log_snapshot.proto
```diff
@@ -14,9 +14,7 @@ message CompactLogSnapshot {
 
 message LogFileReference {
   string uuid = 1;
-  string s3_location = 2;
-  string s3_cek_algo = 3;
-  string s3_cmk_key_id = 4;
+  // 2-4 are deprecated and shouldn't be used
```
seanlinsley, line 17: This changes which fields are accessible to the pganalyze server when processing snapshots, so I've marked this PR as a draft until we're ready to commit to removing server-side support.

lfittl (approved):

PR #797, opened 2026-03-25
Track nested query stats using toplevel field

seanlinsley:
The `pg_stat_statements.toplevel` field tracks whether a query was called as a statement (true) or from inside of a function (false). Historically the collector has merged the query stats of both versions and reported them as a single entry, but this PR changes that to report the toplevel field to pganalyze, making it possible to distinguish between top-level calls and nested calls.


protobuf/shared.proto
```diff
@@ -62,6 +62,7 @@ message QueryReference {
   int32 database_idx = 1;
   int32 role_idx = 2;
   bytes fingerprint = 3;
+  optional bool toplevel = 4;
```
seanlinsley, line 65: This is the first use of `optional` in the repo; was there a reason we were avoiding it?

I chose to use `optional` so on the Rust side it's `Option<bool>` and can be `toplevel.unwrap_or(true)` to set the default value for old collectors. I felt that was preferable to having a collector version check. Another option would be to rename it from `toplevel` to `nested`, so the default false value works automatically.

lfittl: Seems reasonable, I think we mainly don't use it much today because we initially developed against proto2 which didn't support optional fields.

output/transform/postgres_statements.go
lfittl: This comments needs updating

output/transform/util.go
lfittl, line 80: I'd add a note here that states that we intentionally don't set toplevel because we don't know if the query in question was at the top-level or not.

lfittl (approved):

PR #802, opened 2026-04-09
WIP: Replace prepare statements with plain queries

caius:
We're forced to run through a transaction based connection pool, which doesn't support prepared statements due to connections being swapped out between the collector and the database server.

Replace all the explicit prepare/query calls with plain query calls instead.

Not expecting this to be accepted upstream, but its the version we're running internally that's working for us.

### DIFF
--- a/input/postgres/backend_counts.go
+++ b/input/postgres/backend_counts.go
@@ -27,14 +27,7 @@ func GetBackendCounts(ctx context.Context, c *Collection, db *sql.DB) ([]state.P
 		sourceTable = "pg_catalog.pg_stat_activity"
 	}
 
-	stmt, err := db.PrepareContext(ctx, QueryMarkerSQL+fmt.Sprintf(backendCountsSQL, sourceTable))
-	if err != nil {
-		return nil, err
-	}
-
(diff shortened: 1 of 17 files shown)


lfittl:
> We're forced to run through a transaction based connection pool, which doesn't support prepared statements due to connections being swapped out between the collector and the database server.
> 
> Replace all the explicit prepare/query calls with plain query calls instead.
> 
> Not expecting this to be accepted upstream, but its the version we're running internally that's working for us.

Thanks for sharing!

Clearly my assumptions in https://github.com/pganalyze/collector/pull/793 were incorrect, since we explicitly call the functions that produce prepared statements.

We'll discuss internally how to go about this to address it one way or the other. For context, we do generally discourage running a connection pooler between the pganalyze collector and the database, but understand that in certain deployments this is hard to avoid.

keiko713:
Thanks Caius for this PR! Closing this as we merged the needed changes with https://github.com/pganalyze/collector/pull/810
Planning to cut a new release with this soon. Thanks again for the PR, as well as confirming that this works with your setup.

PR #803, opened 2026-04-15
Check if pganalyze server supports breaking changes

seanlinsley:
#797 required server-side changes to properly support, so this PR introduces a `CompatFlags` server message allowing the server to tell the collector which breaking changes it supports. This allows new collector versions to continue working with old pganalyze Enterprise server versions.


output/transform/postgres_statements.go
```diff
@@ -28,7 +28,9 @@ func groupStatements(statements state.PostgresStatementMap, statsMap state.Diffe
 			databaseOid: sKey.DatabaseOid,
 			userOid:     sKey.UserOid,
 			fingerprint: statement.Fingerprint,
-			toplevel:    sKey.Toplevel,
+		}
+		if server.Grant.Load().Config.GetCompatFlags().GetToplevel() {
```
seanlinsley: Using the generated `Get` functions is necessary here to avoid nil pointer errors, since this struct isn't present when connected to an old server.

protobuf/server_message.proto
```diff
@@ -24,6 +24,9 @@ message ServerMessage {
     bool enable_logs = 6;
     // Maximum number of tables that can be monitored per server
     int32 schema_table_limit = 7;
+    // Server-side support of breaking changes, so new collector versions fall back to
+    // the original behavior when connected to old Enterprise servers
+    CompatFlags compat_flags = 8;
```
seanlinsley: Other possible names are `Capabilities`, `ServerSupports`, etc

msakrejda: I think I prefer `Capabilities` but maybe just because it's a full word. No strong feelings here.

output/transform/postgres_statements.go
```diff
@@ -28,7 +28,9 @@ func groupStatements(statements state.PostgresStatementMap, statsMap state.Diffe
...
+			key.toplevel = sKey.Toplevel
```
msakrejda, line 35: Is it enough to just omit the `toplevel` flag? I thought the issue was that an older server would receive both `toplevel` and `!toplevel` versions of the same query, try to store both, and hit unique constraint problems. With this change we're still sending both, just without the `toplevel` flag. Or am I misunderstanding the issue?

seanlinsley: This change is inside of the `groupStatements` function, which groups multiple query stats together if they have the same identity. By not setting `key.toplevel`, we effectively fall back to the original behavior where query stats were grouped by fingerprint instead of fingerprint + toplevel.

protobuf/server_message.proto
seanlinsley: Lukas also preferred capabilities when we initially discussed this. Not a big deal either way, but I slightly prefer something  compat/breaking/support-adjacent to make it clear this signals what breaking changes the server is compatible with. As opposed to capabilities which could be interpreted to mean a generic list of features the server supports.

If we do end up preferring capabilities, does it really make sense for this to be a separate struct in the protobuf or can it be thrown into the existing `Features` struct?

output/transform/postgres_statements.go
msakrejda: Got it; that makes sense.

protobuf/server_message.proto
msakrejda: Fair point. I kinda feel it should be distinct from Features, since that's about what the user has configured to influence collector behavior. This is more about influencing collector behavior based on what the server is capable of processing. That said, I think a field naming convention inside Features could be enough to distinguish these.

lfittl: I'd suggest using `Capabilities` for now to keep it simple, we can always adjust names later whilst keeping the same numeric field number.

lfittl (approved):

protobuf/server_message.proto
```diff
@@ -53,4 +56,8 @@ message ServerMessage {
     repeated string query_parameter_types = 6;
     map<string, string> postgres_settings = 7;
   }
+
+  message CompatFlags {
+    bool toplevel = 1;
```
lfittl, line 64: Might be worth documenting what capability this is (i.e. that its the fact that the collector will send stats separately for toplevel=true vs toplevel=false)

PR #804, opened 2026-04-15
Release 0.70.1

seanlinsley:
Releasing #803 to address breaking changes in #797

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,11 @@
 
 ## 0.70.1      2026-04-15
 
-* Fix compatibility with existing Enterprise Server releases
+* Fix compatibility with existing pganalyze Enterprise Server releases
   - Version 0.70.0 added tracking of nested query statistics using the toplevel field,
-    which broke compatibility with Enterprise Server. This change detects whether the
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 0.70.1      2026-04-15
+
+* Fix compatibility with existing Enterprise Server releases
+  - Version 0.70.0 added tracking of nested query statistics using the toplevel field,
+    which broke compatibility with Enterprise Server. This change detects whether the
```
lfittl: Maybe "which broke compatibility with Enterprise Server when using pg_stat_statements.track = all"? (since that's not the default, so not everybody would be affected)

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 0.70.1      2026-04-15
+
+* Fix compatibility with existing Enterprise Server releases
```
lfittl:
```suggestion
* Fix compatibility with existing pganalyze Enterprise Server releases
```

(just because "Enterprise Server" could also be a Postgres variant instead of our product)

lfittl (approved):

PR #805, opened 2026-04-16
Improve disk size reporting for Aurora

keiko713:
Use the `VolumeBytesUsed` CloudWatch metric (queried by `DBClusterIdentifier`) for Aurora instead of the previously inaccurate OS-level filesystem values or `AllocatedStorage`. This was wrongly reporting a few MB storage usage regardless of the actual cluster storage size.
With the `GetRdsClusterIntMetric` function (to obtain `VolumeBytesUsed` metric), use the 3 hours lookback window to retrieve the data reliably (see below for more detail).

In passing, update `AuroraMaxStorage` from 64TB to 128TB to reflect the current Aurora storage limit.

Several notes (likely worth noting this in the UI side):
* Aurora volume metrics like `VolumeBytesUsed` are [reported infrequently](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.AuroraMonitoring.Metrics.html) ("collected at intervals, not continuously"). In testing, a 10-minute window returned no data at all, and a 1-hour window still produced frequent gaps. A 3-hour window reliably returns data. When data is available, we pick the most recent datapoint.
  * This means that the data can be at max 3 hours old, also there might be still gap (will report 0B in that case)
* `VolumeBytesUsed` is only available as a cluster-level metric (dimension: `DBClusterIdentifier`), not per-instance.
  * This means all instances in the same cluster will report the same disk usage

This server is using this branch:
https://staging.pganalyze.com/servers/336fdc7yijartkopql6yerfisy/system/storage?t=24h

### DIFF
--- a/input/system/rds/system.go
+++ b/input/system/rds/system.go
@@ -100,7 +100,8 @@ func GetSystemState(server *state.Server, logger *util.Logger) (system state.Sys
 	system.Info.AmazonRds.ParameterApplyStatus = *group.ParameterApplyStatus
 
 	dbInstanceID := *instance.DBInstanceIdentifier
-	cloudWatchReader := awsutil.NewRdsCloudWatchReader(sess, logger, dbInstanceID)
+	dbClusterID := util.StringPtrToString(instance.DBClusterIdentifier)
+	cloudWatchReader := awsutil.NewRdsCloudWatchReader(sess, logger, dbInstanceID, dbClusterID)
 
 	system.Disks = make(state.DiskMap)
 	system.Disks["default"] = state.Disk{
@@ -225,7 +226,7 @@ func GetSystemState(server *state.Server, logger *util.Logger) (system state.Sys
 					usedBytes := uint64(diskPartition.Used * 1024)
 					totalBytes := uint64(diskPartition.Total * 1024)
 					if isAurora {
-						auroraVolumeUsed := uint64(cloudWatchReader.GetRdsIntMetric("VolumeBytesUsed", "Bytes"))
+						auroraVolumeUsed := uint64(cloudWatchReader.GetRdsClusterIntMetric("VolumeBytesUsed", "Bytes"))
 						if auroraVolumeUsed > 0 {
 							usedBytes = auroraVolumeUsed
 						}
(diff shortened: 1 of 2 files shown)


input/system/rds/system.go
```diff
@@ -219,16 +221,25 @@ func GetSystemState(server *state.Server, logger *util.Logger) (system state.Sys
...
-					if isAurora {
-						totalBytes = AuroraMaxStorage
+
+				if isAurora {
+					auroraVolumeUsed := cloudWatchReader.GetRdsClusterIntMetric("VolumeBytesUsed", "Bytes")
+					for _, diskPartition := range osSnapshot.FileSystems {
+						system.DiskPartitions[diskPartition.MountPoint] = state.DiskPartition{
+							DiskName:      "default",
+							PartitionName: diskPartition.Name,
+							UsedBytes:     uint64(auroraVolumeUsed),
```
msakrejda: So we're recording the same CloudWatch VolumeBytesUsed number for every partition. Is that right?

input/system/rds/system.go
```diff
@@ -13,9 +13,10 @@ import (
 	"github.com/pganalyze/collector/util/awsutil"
 )
 
-// Aurora storage is automatically extended up until 64TB, so we should always
-// report that limit as the total disk space (to avoid bogus disk space warnings)
-const AuroraMaxStorage = 64 * 1024 * 1024 * 1024 * 1024
+// Aurora storage is automatically extended up to 128TB, so report that as
+// total disk space when we can't determine the actual limit from CloudWatch
```
msakrejda: >... when we can't determine the actual limit from CloudWatch

Are we trying to determine the actual limit from CouldWatch? I don't see that: we seem to just always use this value for Aurora (which is fine, but the comment needs an update).

msakrejda (approved):
Minor comments, but looks good.

input/system/rds/system.go
keiko713: Ah, yeah no, not really. Updated.

input/system/rds/system.go
keiko713: Good point - so we only ever use these UsedBytes or TotalBytes for data dir partition in the pganalyze app side (so throwing other values away), but reporting completely off values for other partitions like /tmp is not nice in case we wanna ever use it in the future in the app side. I tweaked to make sure only override with the data dir partition.

PR #806, opened 2026-04-17
CONTRIBUTING: Improve order of Docker release steps

lfittl:
The steps as written don't actually make sense, since you can't login to docker before installing it. Also we can install the ARM emulation support at the same time as docker itself to make the last step not mix two unrelated actions.


keiko713 (approved):
Thanks for fixing!!

PR #810, opened 2026-05-01
Remove prepared statements from all collector queries

keiko713:
Based on (or mostly the same as) https://github.com/pganalyze/collector/pull/802

- Previously, explicit prepared statements broke when Cloud SQL / AlloyDB is configured with IAM authentication.
- Fix this by removing all PrepareContext usage from the collector queries, and switching the Cloud SQL / AlloyDB IAM pgx driver to the simple query protocol, to ensure that prepared statements aren't used.

With this change, every query is a self-contained round trip. lib/pq sends an unnamed Parse+Bind+Execute, and the pgx/IAM path uses simple protocol. The custom pgx driver stays because we still need its custom dialer for IAM auth, and we set QueryExecModeExec there to also defeat pgx's default named-statement caching.

----

Note: tested with non Cloud SQL / AlloyDB instances, everything worked fine.

### DIFF
--- a/util/pgxdriver/pgxdriver.go
+++ b/util/pgxdriver/pgxdriver.go
@@ -11,9 +11,12 @@ import (
 	"github.com/jackc/pgx/v5/stdlib"
 )
 
-// pgxDriver is a database/sql driver that uses pgx with QueryExecModeExec
-// to avoid creating server-side prepared statements. It wraps a dial function
-// (Cloud SQL or AlloyDB) that handles the actual connection.
+// pgxDriver is a database/sql driver that wraps pgx with a custom dial
+// function for Cloud SQL or AlloyDB IAM authentication. It uses
+// QueryExecModeSimpleProtocol so queries are sent via the simple query
+// protocol, which is required when these managed services sit behind a
+// transaction-mode connection pooler that cannot reuse prepared statements
+// across pooled backend connections.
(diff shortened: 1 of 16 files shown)


util/pgxdriver/pgxdriver.go
```diff
@@ -41,7 +44,7 @@ func (p *pgxDriver) dbURI(name string) (string, error) {
 	}
 	instConnName := config.Config.Host
 	config.Config.Host = "localhost"
-	config.DefaultQueryExecMode = pgx.QueryExecModeExec
+	config.DefaultQueryExecMode = pgx.QueryExecModeSimpleProtocol
```
lfittl: Why is this change needed again? (did we maybe suggest this because we missed the fact that we explicitly did prepares?)

I don't think the extended protocol on its own is a problem, just prepared statements are. Per https://pkg.go.dev/github.com/jackc/pgx/v5#QueryExecMode

(and I do think its nice for the rare cases where we do pass parameters to not rely on string interpolation)

keiko713: Yeah you're right, we need `QueryExecModeExec` (aka we can't use the default `QueryExecModeCacheStatement`), but we don't need `QueryExecModeSimpleProtocol`. I'll revert.

lfittl (approved):

PR #811, opened 2026-05-01
remove warning when aurora_computer_plan_id is not enabled

ryanbooz:
Per recent discussions, removing this informational warning for now. cc: @lfittl

### DIFF
--- a/input/postgres/plans.go
+++ b/input/postgres/plans.go
@@ -62,7 +62,7 @@ func GetPlans(ctx context.Context, c *Collection, db *sql.DB, showtext bool) (st
 		// aurora_compute_plan_id needs to be on to use aurora_stat_plans function
 		// 2026-04-30: For now, we've removed guidance in documentation to enable
 		//   this setting because of potential LWLock:pg_stat_statement locking/wait
-		//   issues on Aurora Serverless on high-concurrency workloads when it's enabled.
+		//   issues on Amazon Aurora on high-concurrency workloads when it's enabled.
 		//   We'll revisit this in the future when we have more information about the underlying
 		//   issue and potential fixes from AWS.
 
--- a/input/postgres/plans.go
+++ b/input/postgres/plans.go
@@ -60,10 +60,16 @@ func GetPlans(ctx context.Context, c *Collection, db *sql.DB, showtext bool) (st
 			return nil, nil, err
 		}
 		// aurora_compute_plan_id needs to be on to use aurora_stat_plans function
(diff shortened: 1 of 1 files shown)


input/postgres/plans.go
```diff
@@ -60,12 +60,18 @@ func GetPlans(ctx context.Context, c *Collection, db *sql.DB, showtext bool) (st
...
+		//   issues on Aurora Serverless on high-concurrency workloads when it's enabled.
+		//   We'll revisit this in the future when we have more information about the underlying
+		//   issue and potential fixes from AWS.
+
+		// if computePlanIdEnabled != "on" {
+		// 	if c.GlobalOpts.TestRun {
+		// 		c.Logger.PrintInfo("Function aurora_stat_plans() is not supported because aurora_compute_plan_id is turned off. Skipping collecting query plans and stats.")
+		// 	}
+		// 	return nil, nil, nil
+		// }
```
lfittl: I think we still need the if condition, otherwise we'll try to access the data and fail:

```suggestion
		if computePlanIdEnabled != "on" {
		// 	if c.GlobalOpts.TestRun {
		// 		c.Logger.PrintInfo("Function aurora_stat_plans() is not supported because aurora_compute_plan_id is turned off. Skipping collecting query plans and stats.")
		// 	}
			return nil, nil, nil
		}
```

input/postgres/plans.go
```diff
@@ -60,12 +60,18 @@ func GetPlans(ctx context.Context, c *Collection, db *sql.DB, showtext bool) (st
...
 		// aurora_compute_plan_id needs to be on to use aurora_stat_plans function
-		if computePlanIdEnabled != "on" {
-			if c.GlobalOpts.TestRun {
-				c.Logger.PrintInfo("Function aurora_stat_plans() is not supported because aurora_compute_plan_id is turned off. Skipping collecting query plans and stats.")
-			}
-			return nil, nil, nil
-		}
+		// 2026-04-30: For now, we've removed guidance in documentation to enable
+		//   this setting because of potential LWLock:pg_stat_statement locking/wait
+		//   issues on Aurora Serverless on high-concurrency workloads when it's enabled.
```
lfittl:
```suggestion
		//   issues on Amazon Aurora on high-concurrency workloads when it's enabled.
```

(I don't think this is specific to Serverless, rather its an issue with how `aurora_stat_plans` is implemented for both regular Aurora and Serverless)

keiko713 (approved):

lfittl (approved):

PR #816, opened 2026-05-07
Log processing: Support log_error_verbosity = verbose

lfittl:
Previously the collector did not support processing any logs when log_error_verbosity was set to verbose, due the log format differing slightly, since SQLSTATE gets added inline to the server log output.

Address by adding special cased parsing for when this is enabled, and drop the error checks. Note that the special casing is not needed for jsonlog/csvlog, since Postgres does not modify the message for them.

Additionally, handle LOCATION/BACKTRACE log lines but don't send them to the server side, since they are not useful outside of Postgres development use cases.


msakrejda (approved):
Looks good.

PR #819, opened 2026-05-18
Track IO worker backends from Postgres 18+

seanlinsley:
For context: https://pganalyze.com/blog/postgres-18-async-io


lfittl (approved):
:+1:

PR #820, opened 2026-05-29
feat / helm chart: allow passing deployment annotations

davidmontoyago:
this is useful to have tools like https://github.com/stakater/reloader automatically restart the collector on changes to configuration


keiko713 (approved):
Looks like a good addition, thanks for the contribution!

PR #821, opened 2026-06-08
Report collector start time in request header

seanlinsley:
This adds a `Pganalyze-Collector-Started-At` HTTP header to all requests to the pganalyze server, allowing us to know if the websocket connection is unreliable because of network issues, or if the collector itself is restarting.

### DIFF
--- a/config/read.go
+++ b/config/read.go
@@ -498,7 +498,7 @@ func CreateWebSocketDialer(conf ServerConfig) websocket.Dialer {
 	}
 }
 
-func APIHeaders(conf ServerConfig, testRun bool) map[string][]string {
+func APIHeaders(conf ServerConfig, testRun bool, startedAt time.Time) map[string][]string {
 	headers := make(map[string][]string)
 	headers["Pganalyze-Api-Key"] = []string{conf.APIKey}
 	headers["Pganalyze-System-Id"] = []string{conf.SystemID}
@@ -507,6 +507,7 @@ func APIHeaders(conf ServerConfig, testRun bool) map[string][]string {
 	headers["Pganalyze-System-Id-Fallback"] = []string{conf.SystemIDFallback}
 	headers["Pganalyze-System-Type-Fallback"] = []string{conf.SystemTypeFallback}
 	headers["Pganalyze-System-Scope-Fallback"] = []string{conf.SystemScopeFallback}
+	headers["Pganalyze-Collector-Started-At"] = []string{startedAt.Format(time.RFC3339)}
 	if testRun {
 		headers["Pganalyze-Test-Run"] = []string{"true"}
 	}
--- a/main.go
+++ b/main.go
@@ -311,6 +311,7 @@ ReadConfigAndRun:
 			cancel()
 			wg.Wait()
 			writeStateFile()
+			opts.StartedAt = time.Now() // Signal to pganalyze API that collector essentially restarted
 			goto ReadConfigAndRun
 		}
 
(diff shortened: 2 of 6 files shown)


config/read.go
```diff
@@ -507,6 +507,7 @@ func APIHeaders(conf ServerConfig, testRun bool) map[string][]string {
 	headers["Pganalyze-System-Id-Fallback"] = []string{conf.SystemIDFallback}
 	headers["Pganalyze-System-Type-Fallback"] = []string{conf.SystemTypeFallback}
 	headers["Pganalyze-System-Scope-Fallback"] = []string{conf.SystemScopeFallback}
+	headers["Pganalyze-Collector-Started-At"] = []string{startedAt.Format(time.RFC3339)}
```
seanlinsley, line 510: Currently this reports the timestamp as a RFC3339 string, with just seconds precision. It could instead use nanosecond precision, though it's unfortunate that'll be truncated to microsecond precision when stored in Postgres.

Alternatively instead of a time string, we could report the number of microseconds since the epoch.

config/read.go
```diff
@@ -498,7 +498,7 @@ func CreateWebSocketDialer(conf ServerConfig) websocket.Dialer {
 	}
 }
 
-func APIHeaders(conf ServerConfig, testRun bool) map[string][]string {
+func APIHeaders(conf ServerConfig, testRun bool, startedAt time.Time) map[string][]string {
```
seanlinsley, line 501: I would've liked to have the function accept `state.CollectionOpts`, but that causes a recursive dependency error.

lfittl: I wonder if its worth moving `state.CollectionOpts` to be in the `config` module instead? They are a kind of configuration really (not a state), just being passed in via the command line instead of the file.

config/read.go
lfittl: I think RFC3339 is reasonable, and a bit easier to work with if one were to manually look at the headers for some reason.

I don't see the need for more than second precision right now, since we care more about "is the collector alive longer than the websocket connection", and that plays out on a minutes to hours timescale, not less than seconds, I think.

lfittl:
Overall this seems reasonable, but I wonder if we shouldn't instead be reporting the time of the last reload (or possibly both?), since reloads will restart the websocket connection, but that's expected and not a problem to be reported.

seanlinsley:
I suspect that no one reloads the collector config often enough that they'd trigger the unstable websocket warning, which only shows if the average connection age is less than an hour. With that assumption, it seems better to stick to just the collector start time.

lfittl:
> I suspect that no one reloads the collector config often enough that they'd trigger the unstable websocket warning, which only shows if the average connection age is less than an hour. With that assumption, it seems better to stick to just the collector start time.

I recall at least one case where someone had automated collector config refreshes and did frequent reloads (every 30mins AFAIK). Not saying its great (it can drop snapshots if you're unlucky), but I don't see a reason why we can't send the reload time as well to help us interpret what's going on.

seanlinsley:
My concern is that it over-complicates the warning logic, and the data layout on the Collector Settings page to track both or even just tracking the last reload time. I'm not sure it's worth the complexity for a user who's reloading the collector config every 30 minutes; I struggle to imagine a valid use case for that.

seanlinsley:
As discussed with Lukas it might make sense to reset `opts.StartedAt` when a reload occurs. If it turns out that timestamp is only relevant to `started_at_mismatched_at` which we no longer use, there may not be any other uses that rely on it so we can change its meaning to be "started time, or most recent reload time"

lfittl (approved):

PR #822, opened 2026-06-09
Retry failed snapshot submission

seanlinsley:
Snapshots are now submitted to a `Queue` that retains recent snapshots if submission fails. This significantly reduces the risk of data loss when there are temporary network connectivity or API availability issues.

### DIFF
--- a/output/upload_http_legacy.go
+++ b/output/upload_http_legacy.go
@@ -3,59 +3,27 @@ package output
 import (
 	"bytes"
 	"context"
-	"encoding/json"
-	"encoding/xml"
 	"fmt"
 	"io"
-	"mime"
 	"mime/multipart"
 	"net/http"
-	"net/url"
-	"os"
-	"path/filepath"
-	"strings"
 	"time"
 
-	"github.com/pganalyze/collector/config"
 	"github.com/pganalyze/collector/state"
 	"github.com/pganalyze/collector/util"
 )
 
-func uploadSnapshot(ctx context.Context, httpClient *http.Client, grant *state.Grant, logger *util.Logger, data []byte, filename string) (string, error) {
-	var err error
-
+func uploadSnapshot(ctx context.Context, httpClient *http.Client, grant *state.Grant, logger *util.Logger, data []byte) error {
 	if !grant.ValidForS3Until.After(time.Now()) {
-		return "", fmt.Errorf("Error - can't upload without valid S3 grant")
-	}
-
-	if grant.S3URL == "" && grant.LocalDir != "" {
-		location := grant.LocalDir + filename
-		err = os.MkdirAll(filepath.Dir(location), 0755)
-		if err != nil {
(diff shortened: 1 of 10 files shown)


output/upload_http_legacy.go
```diff
@@ -3,50 +3,24 @@ package output
...
 
-func uploadSnapshot(ctx context.Context, httpClient *http.Client, grant *state.Grant, logger *util.Logger, data []byte, filename string) (string, error) {
-	var err error
-
+func uploadSnapshot(ctx context.Context, httpClient *http.Client, grant *state.Grant, logger *util.Logger, data []byte) error {
 	if !grant.ValidForS3Until.After(time.Now()) {
-		return "", fmt.Errorf("Error - can't upload without valid S3 grant")
-	}
-
-	if grant.S3URL == "" && grant.LocalDir != "" {
```
seanlinsley, line 31: Is `LocalDir` still in use?

output/upload_http_legacy.go
```diff
@@ -64,111 +38,37 @@ func uploadToS3(ctx context.Context, httpClient *http.Client, S3URL string, S3Fi
...
-	return s3Resp.Key, nil
-}
-
-func submitSnapshot(ctx context.Context, server *state.Server, testRun bool, logger *util.Logger, s3Location string, collectedAt time.Time, compact bool) error {
-	requestURL := server.Config.APIBaseURL + "/v2/snapshots"
-
-	if testRun {
-		requestURL = server.Config.APIBaseURL + "/v2/snapshots/test"
-	} else if compact {
-		requestURL = server.Config.APIBaseURL + "/v2/snapshots/compact"
```
seanlinsley, line 115: I dropped this function primarily so that the snapshot UUID and collected timestamp no longer need to be retained. `/v2/snapshots/compact` is no longer used, but what about `/v2/snapshots/test`?

lfittl:
As discussed, an open question here is whether we should store this in temp files or not - if we can reliably assume that full snapshots are only a handful of MBs in size (which I think is unlikely) it might be fine to stay in memory, otherwise we should utilize temp files (at least when the snapshot is big) to avoid high peak memory use when the server side is temporarily unavailable.

We could also consider having a hard limit of how much memory is used across all servers on the collector for this purpose to avoid OOMs, if there is a good reason to not use temp files.

seanlinsley:
This PR was essentially rewritten. Notable changes:
- All servers now respect a global 200 MB max queue size.
- A single queue per server is used to avoid significant complexity needed to coordinate two separate queues (previously there were separate queues for compact snapshots and full snapshots). As a consequence the max per-server queue size was increased from 10 to 200 to ensure we retain at least one full snapshot for a reasonable amount of time.
- Snapshot submission over the websocket now returns an error, so we're able to correctly retain snapshots in the queue.
- There are now tests to validate the queue's behavior.

PR #823, opened 2026-06-10
Require WebSocket connection by default

seanlinsley:
Followup to #745, this requires that the WebSocket API is always used by default. If after upgrading your collector isn't able to maintain a long-lived WebSocket connection to the pganalyze server, you may have a network proxy whose configuration needs to be updated. As a temporary measure you can set the `api_require_websocket` / `API_REQUIRE_WEBSOCKET` setting to `off`, but a future release will remove the legacy API, so network issues will need to be resolved.


lfittl:
As discussed, we're reviewing whether we can get an email alert for "collector stopped sending snapshots" released first, so in the case of automatic package updates causing breakage, people at least get an email notification so they can take action quickly (e.g. by adjusting the setting). Otherwise this change looks good to ship, since the collector had a note about this in the changelog for the 0.67.0 release (Dec 2025).

PR #825, opened 2026-06-25
Migrate from aws-sdk-go v1 to aws-sdk-go-v2

shreyas-sriram:
aws-sdk-go v1 reached end-of-life on 2025-07-31 and no longer receives security patches or CVE reports. This PR replaces it with the v2 SDK.

## What changed

**`util/awsutil/amazon.go`** — rewritten
- `GetAwsSession() *session.Session` replaced by `GetAwsConfig() aws.Config`
- Added `NewRdsClient`, `NewCloudWatchClient`, `NewCloudWatchLogsClient` helpers that apply per-service custom endpoint overrides (the v1 global session endpoint resolver has no direct equivalent in v2)
- Credential chain: static → env → shared credentials → EC2 IMDS, same precedence as before; STS assume-role and web identity token flows preserved

**`util/awsutil/rds.go`** — updated
- Service clients use `*rds.Client` / `*cloudwatch.Client` (v2 types)
- All API calls take a context; `context.Background()` used where no context was previously threaded through
- CloudWatch `Statistics` and `Dimensions` use typed slice values instead of `[]*string` / `[]*cloudwatch.Dimension`
- `Period` changed from `*int64` to `*int32` per v2 API

**`input/system/rds/system.go`** — updated
- `BackupRetentionPeriod`, `Iops`, `AllocatedStorage`, `MonitoringInterval` are `*int32` in v2 (were `*int64`); uses existing `util.Int32PtrToInt`
- `EnabledCloudwatchLogsExports` is `[]string` in v2 (was `[]*string`)
- `DBParameterGroups` is `[]DBParameterGroupStatus` values in v2 (was `[]*DBParameterGroupStatus`)

**`input/system/rds/logs.go`** — updated
- `getAwsDbInstanceID` and `downloadRdsLogFilePortion` now take a context
- `AdditionalDataPending` dereferenced via `aws.ToBool` (nil-safe)

**`input/postgres/iam_connection.go`** — updated
- `rdsutils.BuildAuthToken` replaced by `feature/rds/auth.BuildAuthToken`

## Testing

`go build ./...` and `go test ./...` pass. The one pre-existing failure (`TestLogParser` in `logs/`) is a timezone mismatch in the test fixture unrelated to this change — it fails identically on unmodified `main`.


shreyas-sriram:
Hey @lfittl, thanks for kicking off the CI checks!

Please let me know if there are any other checks to get this landed!

util/awsutil/amazon.go
```diff
@@ -1,111 +1,107 @@
...
+		))
 	}
 
-	// add default providers
-	providers = append(providers, &credentials.EnvProvider{})
-	providers = append(providers, &credentials.SharedCredentialsProvider{Filename: "", Profile: ""})
-
-	// add the metadata service
-	def := defaults.Get()
-	def.Config.HTTPClient = config.CreateEC2IMDSHTTPClient(cfg)
```
lfittl, line 62: This had a lower timeout (1 second) vs the default (30 seconds), to match upstream behavior for the default client (https://github.com/aws/aws-sdk-go/pull/3066) - I think we'll need to keep that lower timeout by keep using the CreateEC2IMDSHTTPClient method and passing that to `WithHTTPClient`.

util/awsutil/amazon.go
```diff
@@ -1,111 +1,107 @@
...
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/credentials"
+	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatch"
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatchlogs"
+	"github.com/aws/aws-sdk-go-v2/service/rds"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	collectorConfig "github.com/pganalyze/collector/config"
```
lfittl: I think it'd be more idiomatic in this code base to keep this as `config`, since we already alias the AWS `config` to be `awsconfig`.

util/awsutil/amazon.go
```diff
@@ -1,111 +1,107 @@
...
+// NewCloudWatchClient creates a CloudWatch client with any custom endpoint from serverCfg applied.
+func NewCloudWatchClient(awsCfg aws.Config, serverCfg collectorConfig.ServerConfig) *cloudwatch.Client {
+	var opts []func(*cloudwatch.Options)
+	if serverCfg.AwsEndpointCloudwatchURL != "" {
+		url := serverCfg.AwsEndpointCloudwatchURL
+		opts = append(opts, func(o *cloudwatch.Options) { o.BaseEndpoint = &url })
+	}
+	if serverCfg.AwsEndpointSigningRegion != "" {
+		region := serverCfg.AwsEndpointSigningRegion
+		opts = append(opts, func(o *cloudwatch.Options) { o.Region = region })
```
lfittl: I don't think this is correct - the signing region is not the region that the request is sent against (that could already be modified via the `aws_region` setting), but rather the region the request should be signed for. This is only needed in special cases, but we should keep the original intent and map this to the endpoint signing region only.

input/postgres/iam_connection.go
```diff
@@ -26,17 +25,18 @@ type iamConnectionParams struct {
 func getIamConnectionParams(config config.ServerConfig) (driverName string, iamParams iamConnectionParams, err error) {
```
lfittl: I think we can also pass context here (since the single caller has a context variable), instead of utilizing `context.Background()`.

input/postgres/iam_connection.go
```diff
@@ -26,17 +25,18 @@ type iamConnectionParams struct {
...
 	switch config.SystemType {
 	case "amazon_rds":
-		var sess *session.Session
-		sess, err = awsutil.GetAwsSession(config)
-		if err != nil {
+		var awsCfg, cfgErr = awsutil.GetAwsConfig(config)
```
lfittl: I don't think we need to change the pattern here, but can instead keep the explicit declaration of the result variable.

input/system/rds/logs.go
```diff
@@ -56,9 +55,9 @@ func DownloadLogFiles(ctx context.Context, server *state.Server, opts state.Coll
 		FileLastWritten:      &lastWritten,
 	}
 
-	rdsSvc := rds.New(sess)
+	rdsSvc := awsutil.NewRdsClient(awsCfg, server.Config)
```
lfittl, line 58: If we keep the endpoint configuration in the central config as before we can use a direct initialization of `rds.NewFromConfig(awsCfg)` - I think that makes it a bit easier to understand what's going on.

util/awsutil/rds.go
```diff
@@ -1,242 +1,223 @@
...
 	}
-
 	return nil, fmt.Errorf("Failed to find RDS instance using endpoint-based search for host \"%s\" and port %d", host, port)
 }
 
-func FindRdsInstance(config config.ServerConfig, sess *session.Session) (*rds.DBInstance, error) {
-	svc := rds.New(sess)
+// FindRdsInstance finds and returns an RDS DBInstance for the given server config.
+func FindRdsInstance(serverCfg config.ServerConfig, awsCfg aws.Config) (*rdstypes.DBInstance, error) {
+	return findRdsInstanceWithContext(context.Background(), serverCfg, awsCfg)
```
lfittl: I think we should update callers to pass context instead of using `context.Background` here.

lfittl (commented):
Thanks for the PR & your patience!

I'm leaving a few review comments for context, but since you allowed maintainer edits (and because we want to get this into a release) I'm actually going to directly push up commits that address them shortly.

input/system/rds/logs.go
lfittl: Ah, but I didn't realize the new v2 interface discourages this. I think we should do the actual v1 to v2 change separately from the SDK update, since the old style interface is still supported as v1 (but deprecated), and its not strictly required to do now. I think we'll need to do some more testing to correctly map v1 to v2. But lets keep the new helper methods, since we'll need them anyway.

lfittl (approved):

shreyas-sriram:
Hey @lfittl, thanks for the review and merge!

Issue #826, opened 2026-06-30
unable to install rpm on FIPS enabled AL2023


BradBarnich:
so a standard `dnf install` doesn't work.
```
sh-5.2$ dnf install pganalyze-collector
...
Error: Transaction test error:
  package pganalyze-collector-0.70.2-1.x86_64 does not verify: no digest
```

I work around it with:
```
dnf download -y --destdir /tmp pganalyze-collector.x86_64
rpm --upgrade --replacepkgs --nodigest --nofiledigest /tmp/pganalyze-collector-*.x86_64.rpm
```

Here are some more details
```
sh-5.2$ rpm -Kv pganalyze-collector-0.70.2-1.x86_64.rpm
pganalyze-collector-0.70.2-1.x86_64.rpm:
    Header V4 RSA/SHA256 Signature, key ID a2b5f2f9: OK
    Header SHA1 digest: OK
    Payload SHA256 digest: NOTFOUND
    Payload SHA256 ALT digest: NOTFOUND
    V4 RSA/SHA256 Signature, key ID a2b5f2f9: OK
    MD5 digest: NOTFOUND
sh-5.2$ rpm --upgrade --test --replacepkgs pganalyze-collector-0.70.2-1.x86_64.rpm
        package pganalyze-collector-0.70.2-1.x86_64 does not verify: no digest
```

compared with a non-FIPS image:
```
 Header V4 RSA/SHA256 Signature: OK
  Header SHA1 digest: OK
  V4 RSA/SHA256 Signature: OK
  MD5 digest: OK
```

The chatbot consensus is that without FIPS enabled, it was relying the MD5 digest, which is ignored under FIPS mode. The issue may go away when the rpm building image (https://github.com/pganalyze/collector/blob/main/packages/repo/Dockerfile.repo-rpm#L1) is updated, which needs to happen anyway now that al2 is out of support. But I'm not that familiar with rpm packaging to be certain

I could make a PR with this change if that would be helpful

lfittl:
> The chatbot consensus is that without FIPS enabled, it was relying the MD5 digest, which is ignored under FIPS mode. The issue may go away when the rpm building image (https://github.com/pganalyze/collector/blob/main/packages/repo/Dockerfile.repo-rpm#L1) is updated, which needs to happen anyway now that al2 is out of support. But I'm not that familiar with rpm packaging to be certain

[...]

PR #827, opened 2026-07-03
Package RPMs on rocky8 for sha256 payload digest

BradBarnich:
fix for https://github.com/pganalyze/collector/issues/826

continues building RPMs on amazonlinux:2 to keep the glibc requirement at 2.26, but moves packaging to rockylinux:8, which has new enough rpm to produce sha256 payload digests. This is required for new distributions (Fedora 43) as well as FIPS enabled ones (that block the older md5 digest)

Moving building to rockylinux:8 would raise the glibc requirement to 2.28. When amazonlinux2 support is dropped, building and packaging could go back to the same base image.

### DIFF
--- a/packages/src/Dockerfile.build.rpm-systemd
+++ b/packages/src/Dockerfile.build.rpm-systemd
@@ -42,6 +42,9 @@ RUN cp $CODE_DIR/contrib/systemd/pganalyze-collector.service $SOURCE_DIR/etc/sys
 RUN mkdir -p $SOURCE_DIR/usr/share/pganalyze-collector/sslrootcert
 RUN cp $CODE_DIR/contrib/sslrootcert/* $SOURCE_DIR/usr/share/pganalyze-collector/sslrootcert
 
+# RPM 4.11 in Amazon Linux 2 does not emit SHA256 payload digest headers. Use
+# AL2 only for compiling the binaries, and assemble the RPM with AL2023 which uses
+# the newer RPM 4.16 that emits SHA256, but remains installable on older systems.
 FROM amazonlinux:2023 AS package
 
 ARG VERSION
@@ -51,8 +54,6 @@ ENV ROOT_DIR /root
 ENV SOURCE_DIR /source
 ENV NAME pganalyze-collector
 
-# RPM 4.11 in Amazon Linux 2 does not emit SHA256 payload digest headers. Use
-# AL2 only for compiling the binaries, and assemble the RPM with RPM 4.16.
 RUN dnf install -y ruby3.2 ruby3.2-devel gcc make rpm-build tar gzip xz
 
 # FPM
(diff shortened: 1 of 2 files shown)


lfittl:
Thanks for updating! The new test failure seems unrelated, I'll look into that to see what's up.

packages/src/Dockerfile.build.rpm-systemd
```diff
@@ -50,10 +42,31 @@ RUN cp $CODE_DIR/contrib/systemd/pganalyze-collector.service $SOURCE_DIR/etc/sys
...
+# FPM
+RUN ruby3.2-gem install fpm -v 1.17.0
+
+COPY --from=build $SOURCE_DIR $SOURCE_DIR
+COPY --from=build $CODE_DIR/packages $CODE_DIR/packages
+
 # Build the package
 WORKDIR $ROOT_DIR
+# Use xz payload compression so the Rocky-built RPM remains installable on
+# Amazon Linux 2 / RPM 4.11.
```
lfittl:
```suggestion
# Use xz payload compression so the RPM remains installable on Amazon Linux 2 / RPM 4.11.
```

(since we don't have Rocky Linux in the picture here now)

packages/src/Dockerfile.build.rpm-systemd
```diff
@@ -50,10 +42,31 @@ RUN cp $CODE_DIR/contrib/systemd/pganalyze-collector.service $SOURCE_DIR/etc/sys
...
+ARG VERSION
+ENV GOPATH /go
+ENV CODE_DIR $GOPATH/src/github.com/pganalyze/collector
+ENV ROOT_DIR /root
+ENV SOURCE_DIR /source
+ENV NAME pganalyze-collector
+
+# RPM 4.11 in Amazon Linux 2 does not emit SHA256 payload digest headers. Use
+# AL2 only for compiling the binaries, and assemble the RPM with RPM 4.16.
+RUN dnf install -y ruby3.2 ruby3.2-devel gcc make rpm-build tar gzip xz
```
lfittl, line 57: I'd shift this up a bit, and reword slightly:

```suggestion
# RPM 4.11 in Amazon Linux 2 does not emit SHA256 payload digest headers. Use
# AL2 only for compiling the binaries, and assemble the RPM with AL2023 which uses
# the newer RPM 4.16 that emits SHA256, but remains installable on older systems.
FROM amazonlinux:2023 AS package

ARG VERSION
ENV GOPATH /go
ENV CODE_DIR $GOPATH/src/github.com/pganalyze/collector
ENV ROOT_DIR /root
ENV SOURCE_DIR /source
ENV NAME pganalyze-collector

RUN dnf install -y ruby3.2 ruby3.2-devel gcc make rpm-build tar gzip xz
```

lfittl (approved):
Looks good - two minor comment fixes we should do I think, and then this is ready to go :)

PR #828, opened 2026-07-08
Prevent file and goroutine leaks during self-hosted log tailing

seanlinsley:


input/system/selfhosted/logs.go
```diff
@@ -235,8 +235,16 @@ func tailFile(ctx context.Context, path string, out chan<- SelfHostedLogStreamIt
 	TailLoop:
 		for {
 			select {
-			case line := <-t.Lines():
-				out <- SelfHostedLogStreamItem{Line: line.String()}
+			case line, ok := <-t.Lines():
+				if !ok {
+					break TailLoop
```
seanlinsley, line 240: The `Lines` channel could close due to an internal error. In that case we should break out of the loop to free the file descriptor

input/system/selfhosted/logs.go
```diff
@@ -235,8 +235,16 @@ func tailFile(ctx context.Context, path string, out chan<- SelfHostedLogStreamIt
...
+				}
+				select {
+				case out <- SelfHostedLogStreamItem{Line: line.String()}:
+				case <-ctx.Done():
```
seanlinsley, line 244: Check for context cancellation before writing to the `out` channel to avoid a possible deadlock

input/system/selfhosted/logs.go
```diff
@@ -329,6 +337,12 @@ func setupLogLocationTail(ctx context.Context, logLocation string, out chan<- Se
 		return fmt.Errorf("fsnotify new: %s", err)
 	}
 
+	err = watcher.Add(logLocation)
+	if err != nil {
+		watcher.Close()
+		return fmt.Errorf("fsnotify add \"%s\": %s", logLocation, err)
```
seanlinsley, line 343: `watcher.Add` is now before the goroutine launch so Add failures don't orphan the watcher goroutine and its open file tails.

input/system/selfhosted/logs.go
```diff
@@ -358,7 +372,7 @@ func setupLogLocationTail(ctx context.Context, logLocation string, out chan<- Se
 						}
 					}
 				}
-				if event.Op&fsnotify.Remove == fsnotify.Remove || event.Op&fsnotify.Rename == fsnotify.Rename || event.Op&fsnotify.Chmod == fsnotify.Chmod {
+				if event.Op&fsnotify.Remove == fsnotify.Remove || event.Op&fsnotify.Rename == fsnotify.Rename {
```
seanlinsley, line 375: Not related to the rest of the changes, but I'm not sure it makes sense to close the tail if the permissions change. At the very least we can end up losing in-progress logs.

input/system/selfhosted/logs.go
lfittl: Makes sense especially since that channel is currently unbuffered (and it seems beneficial to keep that, or at most use a low capacity, so that we backpressure on the actual file read in go-tail to avoid high memory use with large log lines).

input/system/selfhosted/logs.go
lfittl: Seems reasonable.

lfittl (approved):

vendor/github.com/papertrail/go-tail/follower/follower.go
```diff
@@ -84,6 +79,9 @@ func (t *Follower) Err() error {
 }
 
 func (t *Follower) Close() {
+	if t.file != nil {
+		t.file.Close()
+	}
 	t.closeCh <- struct{}{}
 }
```
seanlinsley, line 86: https://github.com/papertrail/go-tail/pull/15

lfittl: It looks like this change is not actually safe to do, since it causes a data race between Go routines (since we're calling `t.file.Close()` from a different Go routine than the one that actually manages the file), as observed in https://github.com/pganalyze/collector/actions/runs/29125918855/job/86472202020?pr=827

If I recall our conversation correctly, this eager closing was actually not strictly required, since making the channel buffered addresses the Go routine getting stuck. I've dug a bit into this, and revised that change further in https://github.com/pganalyze/go-tail/pull/1 and also made a PR to update to that: https://github.com/pganalyze/collector/pull/836

PR #830, opened 2026-07-09
Fix schema collection ignore_schema_regexp filtering 

ryanbooz:
For large partitioned catalogs, these two queries were not correctly excluding catalog objects that were part of the regexp. 

relation_stats: rewrite the partition-size rollup so it no longer scans the whole pg_class once per relation. Introduces a primary_objects CTE (the collected set) and a filtered_partitions CTE that drives off pg_inherits_parent_index and only sums the sizes of excluded partitions whose parent is actually collected. On a ~500k-row pg_class this took relation_stats from ~112s to sub-second.

column_stats: honor ignore_schema_regexp (it previously collected every column, including ignored schemas, despite the filter) - matching the other schema queries.

### DIFF
--- a/input/postgres/relation_column_stats.go
+++ b/input/postgres/relation_column_stats.go
@@ -14,6 +14,7 @@ const columnStatsSQL = `
 SELECT schemaname, tablename, attname, inherited, null_frac, avg_width, n_distinct, correlation
   FROM %s
  WHERE schemaname NOT IN ('pg_catalog', 'information_schema')
+   AND ($1 = '' OR (schemaname || '.' || tablename) !~* $1)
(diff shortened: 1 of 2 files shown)


ryanbooz:
Also noting that we have tested and verified these changes with a customer that has these very large catalog counts which was causing extreme delays during the full snapshot. Verification of the SQL results and timings have been shared internally.

lfittl (approved):
Looks good - @msakrejda noted he also wanted to take one more look today for correctness, would be good to wait for that before merging.

msakrejda (approved):
Looks good. Thanks for the fix.

PR #831, opened 2026-07-09
Skip user-defined types that match ignore_schema_regexp

seanlinsley:
Similar to #830, this filters out user-defined types that match `ignore_schema_regexp`. This is unlikely to improve performance because the number of types should be relatively small, but it does improve correctness.


seanlinsley:
As discussed, this may be surprising to users, and our documentation does not imply that user defined types would be filtered out:
>`ignore_schema_regexp`: Skip collecting metadata for all matching tables, schemas, or functions; match is checked against schema-qualified object names (e.g. to ignore table “foo” only in the public schema, set to ^public.foo$)

PR #834, opened 2026-07-09
Fix RDS/Aurora cluster ID collision between accounts

bisonhubert:
Fixes a bug where `cluster_id` for RDS/Aurora is only unique within an AWS account. Because `DBClusterIdentifier` is scoped per account, two accounts with the same `DBClusterIdentifier` are treated as the same cluster.

The fix prepends the AWS account number to the ID by parsing `instance.DBInstanceArn` per [AWS RDS ARN construction](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_Tagging.ARN.Constructing.html).

> `arn:aws:rds:<region>:<account number>:<resourcetype>:<name>`

On the next snapshot after upgrade, RDS/Aurora servers move to a new `ServerGroup` with the scoped `cluster_id`. The old `ServerGroup` is treated as stale with no recent statistics data.


input/system/rds/system.go
```diff
@@ -39,7 +39,14 @@ func GetSystemState(server *state.Server, logger *util.Logger) (system state.Sys
 		return
 	}
 
-	system.Info.ClusterID = util.StringPtrToString(instance.DBClusterIdentifier)
+	clusterID := util.StringPtrToString(instance.DBClusterIdentifier)
+	if clusterID != "" {
+		arnParts := strings.Split(util.StringPtrToString(instance.DBInstanceArn), ":")
```
lfittl, line 45: Maybe worth adding a comment for clarity, like:

```suggestion
        // Include AWS Account ID in cluster ID to avoid incorrectly merging clusters across account boundaries when they belong to the same organization
		arnParts := strings.Split(util.StringPtrToString(instance.DBInstanceArn), ":")
```

(mainly since its not obvious that `arnParts[4]` is the account ID portion)

lfittl (approved):

PR #835, opened 2026-07-10
Fix unbound memory issue in multi-line log stitching

ryanbooz:
I've been working for multiple weeks with a customer that had consistent, ongoing crash loops of the collector in their environment. Part of the problem was tackled in #830 - the full snapshot was taking 2-3 minutes in most cases which increased memory usage and delayed data.

However, there was still periodic crash looping that couldn't be tied to a specific time period (full snapshot, time of day). In one testing run the process (running separately and not bound by `systemctl` `MaxMemory`) climbed to over 11GB in the span of a few minutes and had to be killed.

We ended up doing multiple remote sessions with different debug builds to get the `.mprof` dump when memory was climbing. Ultimately, we identified that they were getting significantly large JSON EXPLAIN plans (4MB or more) and the line-continuation path was using the inefficient `+=` on a string which reallocates the memory each time over thousands and thousands of lines. In a short 3-minute test run, a few EXPLAIN plans generated over 30GB of internal memory work on these reallocations.

After changing this section of code, the results on the next test run were immediate and dramatic. There was a contributing factor of a very stale and large state file that needs to be discussed separately. For now, the very old, stale file was deleted to free up the resources on next restart.

| Metric | Before | After |
|---|---|---|
| `relation_stats` query | ~112 s | ~0.5 s |
| Full snapshot | > 2 min | ~30 s |
| Collector live heap (`inuse_space`) | 4,364 MB | **56 MB** |
| Allocation churn (`alloc_space`) | 44.2 GB | 2.2 GB |
| Log-parse allocation churn | 38.6 GB | 0.44 GB |
| State-file decode at startup | 1.69 GB / 27M objects | ~16.8 MB |
| Stability | crash-loop, no full snapshots (weeks) | 5+ hours |

Aside from using a StringBuilder, there is currently an added cap of 10MB for any single log line. In this case we only pull 10MB from AWS so that single line cap shouldn't be reached, but the log size and content size were not actually bound together and, in theory, the log line could exceed the current cap if not specifically tested for. This will cause a warning message and the additional log line data to be dropped. Open to discussion.

Note: Customer is running this over the weekend and we are monitoring it. However, I wanted to get this PR opened as we potentially approach a new collector release to begin discussions. I can provide `.mprof` files if helpful.

### DIFF
--- a/logs/parse.go
+++ b/logs/parse.go
@@ -462,8 +462,7 @@ func ParseAndAnalyzeBuffer(logStream LineReader, linesNewerThan time.Time, serve
 	}
 
 	// Continuation lines (lines that don't parse as a new log line) are stitched onto
-	// the previous kept line. We accumulate them in a strings.Builder and materialize
-	// once, so a single entry spanning many lines is O(n) rather than O(n^2) copies.
+	// the previous kept line.
 	var additionalLines strings.Builder
 	parentLine := -1
 	additionalLinesExceeded := false
--- a/logs/parse.go
+++ b/logs/parse.go
@@ -447,12 +447,33 @@ type LineReader interface {
 	ReadString(delim byte) (string, error)
 }
 
+// maxAdditionalLinesBytes bounds how much continuation content we append to a single
+// log line. Multi-line entries (e.g. auto_explain JSON plans) are stitched onto their
+// primary line; without a bound, a single very large entry can grow one Content string
+// without limit and OOM the collector. It is a var so tests can lower it.
+var maxAdditionalLinesBytes = 10 * 1024 * 1024
+
 func ParseAndAnalyzeBuffer(logStream LineReader, linesNewerThan time.Time, server *state.Server, opts state.CollectionOpts, logger *util.Logger) ([]state.LogLine, []state.PostgresQuerySample) {
 	var logLines []state.LogLine
 	var currentByteStart int64 = 0
 	parser := server.GetLogParser()
 	if parser == nil {
 		return []state.LogLine{}, []state.PostgresQuerySample{}
+	}
+
+	// Continuation lines (lines that don't parse as a new log line) are stitched onto
+	// the previous kept line. We accumulate them in a strings.Builder and materialize
+	// once, so a single entry spanning many lines is O(n) rather than O(n^2) copies.
+	var additionalLines strings.Builder
+	parentLine := -1
+	additionalLinesExceeded := false
+	flushAdditionalLines := func() {
+		if parentLine >= 0 && additionalLines.Len() > 0 {
+			logLines[parentLine].Content += additionalLines.String()
+		}
+		additionalLines.Reset()
+		parentLine = -1
+		additionalLinesExceeded = false
 	}
 
 	for {
@@ -474,8 +495,21 @@ func ParseAndAnalyzeBuffer(logStream LineReader, linesNewerThan time.Time, serve
 			// Assume that a parsing error in a follow-on line means that we actually
 			// got additional data for the previous line
 			if len(logLines) > 0 && logLine.Content != "" {
-				logLines[len(logLines)-1].Content += logLine.Content
-				logLines[len(logLines)-1].ByteEnd += int64(len(logLine.Content))
+				lastLine := len(logLines) - 1
+				if parentLine != lastLine {
+					// The last kept line changed since we last stitched; finalize it.
+					flushAdditionalLines()
+					parentLine = lastLine
+				}
+				logLines[lastLine].ByteEnd += int64(len(logLine.Content))
+				if additionalLines.Len()+len(logLine.Content) <= maxAdditionalLinesBytes {
+					additionalLines.WriteString(logLine.Content)
+				} else if !additionalLinesExceeded {
+					additionalLinesExceeded = true
+					logger.PrintWarning("Log line continuation exceeded %d bytes and was truncated; "+
+						"a single log entry (e.g. an auto_explain JSON plan) is larger than the "+
+						"collector will stitch together", maxAdditionalLinesBytes)
+				}
 			}
 			continue
 		}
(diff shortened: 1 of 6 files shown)


logs/parse.go
```diff
@@ -455,6 +461,21 @@ func ParseAndAnalyzeBuffer(logStream LineReader, linesNewerThan time.Time, serve
 		return []state.LogLine{}, []state.PostgresQuerySample{}
 	}
 
+	// Continuation lines (lines that don't parse as a new log line) are stitched onto
+	// the previous kept line. We accumulate them in a strings.Builder and materialize
+	// once, so a single entry spanning many lines is O(n) rather than O(n^2) copies.
```
msakrejda:
```suggestion
	// the previous kept line.
```
The first part of the comment is fine, but the second just describes what a strings.Builder is for, so I'd suggest dropping it.

logs/parse_stitch_test.go
```diff
@@ -0,0 +1,71 @@
...
+		"  \"Node\": \"a\",\n" +
+		"  \"Node\": \"b\",\n" +
+		"  \"Node\": \"c\"\n"
+
+	logLines := parseStitchBuffer(input)
+
+	if len(logLines) != 1 {
+		t.Fatalf("expected 1 stitched log line, got %d", len(logLines))
+	}
+	want := "duration: 1.000 ms  plan:\n  \"Node\": \"a\",\n  \"Node\": \"b\",\n  \"Node\": \"c\"\n"
```
msakrejda: This might be more readable with '`' strings rather to avoid all the escapes...

logs/parse_stitch_test.go
```diff
@@ -0,0 +1,71 @@
...
+// Content without bound; it is capped at maxAdditionalLinesBytes.
+func TestParseAndAnalyzeBufferCapsRunawayStitch(t *testing.T) {
+	orig := maxAdditionalLinesBytes
+	maxAdditionalLinesBytes = 4096
+	defer func() { maxAdditionalLinesBytes = orig }()
+
+	var b strings.Builder
+	b.WriteString(stitchTestPrefix + "duration: 1.000 ms  plan:\n")
+	contLine := strings.Repeat("x", 1000) + "\n"
+	for i := 0; i < 200; i++ { // ~200 KB of continuation, far above the 4 KB cap
```
msakrejda: Why not just barely exceed the cap?

msakrejda (approved):
I left some minor comments, but looks great.

ryanbooz:
The earlier memory fix only covered the file-tail path (ParseAndAnalyzeBuffer).
The streaming path (Azure/GCP/Heroku/Tembo/OTel) stitches continuation lines in
its own stitchLogLines, which had no cap — so the same OOM was still reachable
there. Pulled the cap into a shared StitchBuffer so both paths use it.

The stream path already used a Builder, so this just adds the size bound (incl.
capping Grow so a huge input can't spike an allocation before the cap hits).

cc: @msakrejda @lfittl - slightly larger update now and needs another pass.

logs/stitch.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/pganalyze/collector/util"
+)
+
+// MaxStitchedContentBytes bounds how much continuation content is stitched onto a single
+// log line. Multi-line entries (e.g. auto_explain JSON plans) are stitched onto their
+// primary line, and a log_line_prefix mismatch can likewise make every follow-on line
+// look like a continuation. Without a bound, a single Content string can grow without
+// limit and OOM the collector. It is a var so tests can lower it.
+var MaxStitchedContentBytes = 10 * 1024 * 1024
```
msakrejda, line 14: This may be something we want to expose a knob for, but let's give it a shot as a constant and see if there's demand.

logs/stitch.go
msakrejda, line 1: Nice, I like this refactor.

msakrejda (approved):
Re-reviewed. Looks good.

lfittl:
@ryanbooz @msakrejda I've pushed up a refactoring commit (hope you don't mind me changing this branch directly) to avoid use of anonymous local functions, since we typically don't do it in the collector code base, and it jumped out to me at review. I've also used the opportunity to make the download and streaming cases more similar, so its easier to understand what's going on.

Further, I don't think the explicit Grow call for streaming is necessary (it was originally added by me in f6c4a5db623969bfeb04225a4e6b82d206d0c806, but I think the string builder is the more significant change there), and removing it lets us keep the code more readable.

I'd like to get this into the collector release, so would be good if @msakrejda can do a quick pass on that refactor before we merge. Otherwise the code looks good, and great work figuring this out and confirming the impact.

lfittl (approved):

logs/parse.go
```diff
@@ -485,14 +479,8 @@ func ParseAndAnalyzeBuffer(logStream LineReader, linesNewerThan time.Time, serve
...
-			if len(logLines) > 0 && logLine.Content != "" {
-				lastLine := len(logLines) - 1
-				if parentLine != lastLine {
-					// The last kept line changed since we last stitched; finalize it.
-					flushAdditionalLines()
-					parentLine = lastLine
-				}
-				logLines[lastLine].ByteEnd += int64(len(logLine.Content))
+			if pending != nil && logLine.Content != "" {
+				pending.ByteEnd += int64(len(logLine.Content))
```
msakrejda: I know the code you refactored was doing the same thing and I approved that, but should we be bumping `ByteEnd` here? It seems like the more logical place to do that is when we actually append `pending.Content += additionalLines.String()` below (we'd need to break it out into a local variable to get the `additionalLines` content and its length, but I think that's worth it).

msakrejda (approved):
This works and is maybe a little easier to reason about, though we should consider moving the `ByteEnd` handling for clarity.

logs/parse.go
lfittl: Yeah, the `ByteEnd` handling also did jump out to me in the initial code review, but I left that untouched for refactoring since I wasn't fully sure which direction to take. Specifically, do we want to have `ByteEnd` include the bytes that we skipped because we exceeded the stitch limit, or should it only reflect what we actually kept?

Looking at it again, on the backend we mainly use `ByteEnd` for quota checks (plus some other minor things), and we do seem to not advance it for other cases where we skip lines (e.g. ignored lines from log_statement = all), so I suppose we shouldn't do it here either for anything that exceeds the limit.

I've pushed up a commit to make that change based on the existing `Len()` method on `StitchBuffer` (which just goes to the underlying string builder slice length, which is O(1), so cheap to get).

PR #839, opened 2026-07-21
Support Neon logs over OTLP

ryanbooz:
Neon ships Postgres logs as gzipped, RFC5424-syslog-framed OTLP, which the collector received but silently dropped. This adds Neon as a first-class system type and makes that log path work end to end. The gzip and syslog handling are generic, so any syslog-over-OTLP source benefits too.


config/identify_system.go
```diff
@@ -92,6 +93,20 @@ func identifySystem(config ServerConfig) (systemID string, systemType string, sy
...
+		}
+		if systemTypeFallback == "" {
+			systemTypeFallback = "self_hosted"
+		}
+		if systemIDFallback == "" {
+			systemIDFallback = selfManagedSystemID(config)
+		}
+		if systemScopeFallback == "" {
+			systemScopeFallback = selfManagedSystemScope(config)
+		}
```
ryanbooz, line 109: Not sure this is really necessary. In development I was getting the 503 responses because the app doesn't yet know about this `systemType`. Tried this as a workaround seeing the `aiven` setup above. If we have any Neon customers ATM, they would be marked as `self_hosted` and likely produce an error after upgrading the collector I think, which is what this is trying to deal with.

input/system/neon/neon.go
ryanbooz, line 1: Added based on precedence of other systems - although it doesn't provide much value yet.

input/system/neon/neon.go
```diff
@@ -0,0 +1,13 @@
...
+import "github.com/pganalyze/collector/config"
+
+// LogDatabaseFallback returns the database to attribute log lines to when Neon's
+// log_line_prefix omits %d (it only ever serves the configured database).
+// Returns "" for non-Neon servers.
+func LogDatabaseFallback(config config.ServerConfig) string {
+	if config.SystemType != "neon" {
+		return ""
+	}
+	return config.GetDbName()
```
ryanbooz, line 12: I noticed that since we cannot set `log_line_prefix` in Neon, there is no database log lines. I was torn between letting everything at the server level, or associating with the single `neondb` database which most users would be focused on. Not sure if we have precedence for this??

input/system/selfhosted/logs.go
```diff
@@ -477,6 +478,12 @@ func setupLogTransformer(ctx context.Context, wg *sync.WaitGroup, server *state.
 					logLine.LogLineNumberChunk = item.LogLineNumberChunk
 				}
 
+				if logLine.Database == "" {
+					if db := neon.LogDatabaseFallback(server.Config); db != "" {
```
ryanbooz, line 482: See above...

config/identify_system.go
keiko713: Do you want to try shipping the app side first so that you can test better?
To give you some pointer, you'll need to make a collector PR similar to this first: https://github.com/pganalyze/collector/pull/753
Then, make a PR in the app side like https://github.com/pganalyze/pganalyze/pull/5398 , using the collector version with above PR merged.
After that, you can start implementing the logs part (well, I meant you already did here), and you should be able to test it more smoothly.

ryanbooz: This is good feedback and worth doing more deliberately. I'm going to close this PR to avoid any confusion and move it to two smaller chunks in the recommended order.

ryanbooz:
Closing in favor of two smaller PRs that separate this work for clarity and based on feedback.

PR #840, opened 2026-07-22
Add Neon as a supported system type

ryanbooz:
Adding the NeonDB system type so that we can support neon specific features like log collection  once database instances can be identified correctly. Once this is merged and incorporated into pganalyze, a second PR will add the necessary logging updates to support OTLP data from Neon projects.


config/identify_system.go
```diff
@@ -92,6 +93,20 @@ func identifySystem(config ServerConfig) (systemID string, systemType string, sy
 		if systemID == "" {
 			systemID = config.PlanetScaleOrg + "/" + config.PlanetScaleDatabase + "/" + config.PlanetScaleBranch
 		}
+	} else if strings.HasSuffix(config.GetDbHost(), ".neon.tech") || systemType == "neon" {
```
ryanbooz, line 96: There's no specific API work to be done at this time which would require Neon-specific configuration parameters. Each branch gets it's own unique connection string so branch and project naming have no specific value in connecting to the DB and getting metrics.

Happy to discuss an alternative approach.

keiko713 (approved):

PR #841, opened 2026-07-23
Add Supabase as a supported system type

ryanbooz:
We've had recent requests for better log integration on both platforms ([Neon](https://github.com/pganalyze/collector/pull/840) and Supabase), which requires some separate, additional work on OTLP. Adding both now to allow easier testing with system identification.


config/managed_platforms.go
ryanbooz, line 1: I decided to go this route rather than polluting the main identity file with multiple string logic functions. Happy to discuss other options.

In some cases, these platforms may provide better integration later which allows us to identify them specific configuration like other platforms.

msakrejda: Yeah, I think the abstraction is worthwhile.

msakrejda (approved):
Looks good.

PR #842, opened 2026-07-23
Add log lines from (devel) Postgres make check output

msakrejda:
See individual commits:
- Improve log line matching for existing categories
- Add SKIPPING_MAINTENANCE_PERMISSION_DENIED
- Add VALUE_OUT_OF_RANGE
- Add OBJECT_DOES_NOT_EXIST
- Expand CHECK_CONSTRAINT_VIOLATION to catch some partition errors
- Add ROW_LEVEL_SECURITY_VIOLATION
- Add CANNOT_MODIFY_VIEW and WITH_CHECK_OPTION_VIOLATION
- Add SQL_JSON_ERROR
- Add PARTITION_ERROR
- Add OBJECT_ALREADY_EXISTS and WRONG_OBJECT_TYPE
- Recategorize a bit: prefer specific categories

Note that SQL_JSON_ERROR and PARTITION_ERROR are coarser-grained than a lot of the existing event types, but I think this is preferable to both keeping these as UNKNOWN and a proliferation of event types.


keiko713 (approved):
Nice, it was easy to follow per commit. Just to double check - we'll need to create a follow up PR in docs to explain about new/updated categories?

msakrejda:
@keiko713 yeah, and in the app to actually process them. Thanks for the review!

protobuf/compact_log_snapshot.proto
```diff
@@ -202,6 +203,15 @@ message LogLineInformation {
     COULD_NOT_SERIALIZE_REPEATABLE_READ = 138; // "could not serialize access due to concurrent update"
     COULD_NOT_SERIALIZE_SERIALIZABLE = 139; // "could not serialize access due to read/write dependencies among transactions"
     INCONSISTENT_RANGE_BOUNDS = 140; // "range lower bound must be less than or equal to range upper bound"
+    VALUE_OUT_OF_RANGE = 141; // "... out of range", "... field value out of range", "numeric field overflow" (for non-integer types; integers use INTEGER_OUT_OF_RANGE)
+    OBJECT_DOES_NOT_EXIST = 142; // "<object type> ... does not exist" for object types without a dedicated classification (role, type, schema, index, collation, extension, ...); relations/columns/functions/operators use their own
+    ROW_LEVEL_SECURITY_VIOLATION = 143; // "new row violates row-level security policy for table ...", "query would be affected by row-level security policy for table ..." (errcode 42501)
+    WITH_CHECK_OPTION_VIOLATION = 144; // "new row violates check option for view ..." (errcode 44000)
```
lfittl, line 209: As discussed, these could maybe be considered constraint violations of sorts, but the fact that the SQL error code is not in that group means that maybe doesn't make sense. For now seems best to keep it here in the application errors group.

protobuf/compact_log_snapshot.proto
```diff
@@ -139,6 +139,7 @@ message LogLineInformation {
     AUTOANALYZE_COMPLETED = 66; // "automatic analyze of table"
     SKIPPING_VACUUM_LOCK_NOT_AVAILABLE = 67; // "skipping vacuum of ... --- lock not available"
     SKIPPING_ANALYZE_LOCK_NOT_AVAILABLE = 68; // "skipping analyze of ... --- lock not available"
+    SKIPPING_MAINTENANCE_PERMISSION_DENIED = 69; // "permission denied to {vacuum,analyze,execute CLUSTER on,execute REPACK on} ..., skipping it"
```
lfittl, line 142: We should consider renaming the category into "Maintenance activity" or similar, so that REPACK/etc fits more naturally in here, and this fits better here (since its a manual command that emits that kind of error). 

Even the existing ANALYZE is a bit odd, though the autovacuum daemon does kick that one off.

logs/analyze.go
```diff
@@ -922,22 +1000,137 @@ var operatorDoesNotExist = analyzeGroup{
 	classification: pganalyze_collector.LogLineInformation_OPERATOR_DOES_NOT_EXIST,
 	primary: match{
 		prefixes: []string{"operator does not exist: "},
-		regexp:   regexp.MustCompile(`^operator does not exist: (\w+) ([` + regexp.QuoteMeta("+*/<>=~!@#%^&|`?-") + `]+) (\w+)(?: at character \d+)?`),
-		secrets:  []state.LogSecretKind{0, 0, 0},
+		// The remainder is a Postgres-generated operator/type description (e.g. "boolean || boolean",
+		// "@#@ integer", "time with time zone + time with time zone"); it contains no user data, and
+		// the previous, more specific pattern missed unary operators and multi-word type names.
+		regexp:  regexp.MustCompile(`^operator does not exist: .+`),
```
lfittl: It might be good to make this more restrictive (if we can) so that an incorrect log stitching doesn't pull in data by accident (or at least there is less of a chance of it doing so).

lfittl: Should also consider operators that have an explicit namespace set (which probably shows up in the error message?)

logs/analyze.go
```diff
@@ -922,22 +1000,137 @@ var operatorDoesNotExist = analyzeGroup{
...
+			`role|type|index|collation|conversion|cast|extension|language|` +
+			`foreign-data wrapper|foreign server|foreign table|server|` +
+			`user mapping|mapping for token type|event trigger|trigger|` +
+			`subscription|publication|statistics object|large object|tablespace|` +
+			`access method|operator class|operator family|operator \d+|` +
+			`rule|policy|constraint|domain|transform|` +
+			`text search (?:parser|dictionary|template|configuration)|` +
+			`token type|property graph|property|label|schema|database|cursor|` +
+			`prepared statement|savepoint|snapshot|directory|tablesample method` +
+			`)\b.*does not exist.*`),
```
lfittl: Do we need that trailing `.*`?

logs/analyze.go
```diff
@@ -922,22 +1000,137 @@ var operatorDoesNotExist = analyzeGroup{
...
+			"procedure", "prepared statement", "domain", "property graph", "alias",
+		},
+		regexp: regexp.MustCompile(`^(?:` +
+			`role|type|constraint|column|trigger|relation|schema|database|sequence|index|` +
+			`materialized view|foreign table|view|foreign-data wrapper|foreign server|server|` +
+			`user mapping|event trigger|subscription|publication|statistics object|tablespace|` +
+			`operator|rule|policy|(?:default )?conversion|collation|cast|language|extension|` +
+			`text search (?:parser|dictionary|template|configuration)|enum label|function|aggregate|` +
+			`procedure|prepared statement|domain|property graph|alias` +
+			`)\b.*already exists.*`),
```
lfittl: Same here with the trailing `.*`

logs/analyze.go
```diff
@@ -922,22 +1000,137 @@ var operatorDoesNotExist = analyzeGroup{
...
+// excluded by enumerating only object-kind words.
+var wrongObjectType = analyzeGroup{
+	classification: pganalyze_collector.LogLineInformation_WRONG_OBJECT_TYPE,
+	primary: match{
+		prefixes: []string{"\"", "relation", "ALTER action"},
+		regexp: regexp.MustCompile(`^(?:` +
+			`"[^"]*" is (?:not )?an? (?:table or materialized view|materialized view|foreign table|` +
+			`BRIN index|unique index|index for table "[^"]*"|table|view|sequence|index|` +
+			`domain|property graph)` +
+			`|"[^"]*" is (?:not |already )?(?:an? )?.*partition.*` +
```
lfittl: And this trailing `.*`.

logs/analyze.go
```diff
@@ -1106,6 +1346,141 @@ var inconsistentRangeBounds = analyzeGroup{
...
+			"set-returning functions are not allowed in partition", "functions in partition key expression",
+			"UNIQUE constraint on partitioned", "PRIMARY KEY constraint on partitioned",
+			"EXCLUDE constraint on partitioned", "unsupported UNIQUE constraint", "unsupported PRIMARY KEY constraint",
+			"not-null constraint", "identity columns are not supported on partitions", "removing partition",
+			"upper bound of partition", "lower bound of partition", "new partition", "list of new partitions",
+			"list of partitions to be merged", "moving row to another partition",
+			"ROW triggers with transition tables", "TO must specify", "FROM must specify",
+		},
+		regexp: regexp.MustCompile(`^(?:` +
+			`partitions?\b.+` +
```
lfittl: These also have a bunch of trailing `.*` that we can maybe avoid?

logs/analyze.go
```diff
@@ -1106,6 +1346,141 @@ var inconsistentRangeBounds = analyzeGroup{
...
+		},
+		regexp: regexp.MustCompile(`^(?:` +
+			`jsonpath (?:item method \.\w+\(\)|(?:wildcard )?member accessor|(?:wildcard )?array accessor|array subscript) .+` +
+			`|(?:(?:left|right) operand of|operand of unary) jsonpath operator .+` +
+			`|(?:precision|scale|time precision|field position) of jsonpath item method \.\w+\(\) .+` +
+			`|NaN or Infinity is not allowed for jsonpath item method \.\w+\(\)` +
+			`|could not find jsonpath variable "[^"]*"` +
+			`|syntax error at end of jsonpath input(?: at character \d+)?` +
+			`|no SQL/JSON item found for specified path(?: of column "[^"]*")?` +
+			`|JSON path expression .+` +
```
lfittl, line 1473: Same in this set of expressions

msakrejda (commented):
I've updated most of the introduced expressions to avoid a trailing `.*` or `.+`, and rewrote the `match` struct comment to note our philosophy about this (another set of eyes would be welcome there to make sure I captured that correctly).

Three log event types introduced in this PR are hard to tighten: `permissionDenied`, `partitionError`, and `sqlJsonError`. They cover a lot of ground. We could drop them or let them slide for now. These are only an issue if an unrelated log line containing PII is somehow stitched to the end of the line being captured _without a joining newline_, but that's not impossible if we have a bad enough stitching issue.

In addition, `cannotDrop` and `connectionAuthorized` are existing cases of the same over-broad capture. I can follow up on them separately.

PR #844, opened 2026-07-27
Add schema version tracking

seanlinsley:
This adds a `schema_version` to snapshots, allowing us to signal when the schema structure has significantly changed such that the pganalyze server needs to force a re-sync of the data (like https://github.com/pganalyze/collector/pull/196 needed) or otherwise handle it specially (other than checking for optional fields, like normally done).

Since we haven't been using `snapshot_version_major` and `snapshot_version_minor`, and we don't foresee a need for minor version tracking, this PR renames them to `snapshot_version` and `schema_version` to avoid needing to add a new field.


output/full.go
```diff
@@ -48,8 +48,8 @@ func submitFull(ctx context.Context, s *snapshot.FullSnapshot, server *state.Ser
 	}
 
 	s.CollectorErrors = logger.ErrorMessages
-	s.SnapshotVersionMajor = 1
-	s.SnapshotVersionMinor = 0
+	s.SnapshotVersion = 1
+	s.SchemaVersion = 1
```
keiko713, line 52: This PR still doesn't include the part of "allowing us to signal when the schema structure has significantly changed", right? (as simply putting 1 here)

seanlinsley: The internal PR paired with this starts using schema version as part of schema table change detection logic. Was there something else you had in mind?

keiko713: Sorry, I completely misunderstood that part, never mind 🤦

lfittl (approved):

PR #846, opened 2026-07-29
Improve stats diffing on restarts and promotions

msakrejda:
See individual commits:

- Guard statistics interval calculations against bad durations
- Don't diff Postgres statistics across two different instances


msakrejda:
I've removed the DNS changes from this PR, since I think those are going to be more controversial. We could follow up with that separately, but let's start with the simpler changes here.

state/instance_identity.go
```diff
@@ -0,0 +1,73 @@
...
+	return i.PostmasterStartTime.IsZero() && i.ServerAddr == ""
+}
+
+// Matches - Returns whether two identities refer to the same postmaster on the
+// same instance.
+//
+// A zero identity matches anything. We can fail to determine the identity (an
+// old Postgres version, a permission problem, a transient error), and in that
+// case we prefer to keep diffing as before rather than discard data on the
+// basis of information we do not have.
```
msakrejda:
```suggestion
// A zero identity matches anything. We can fail to determine the identity (an
// old Postgres version, a permission problem, a transient error), and in that
// case we avoid using it.
```

state/instance_identity.go
```diff
@@ -0,0 +1,73 @@
...
+// A restart of the same instance is treated as a change too, which is a
+// deliberate choice rather than a necessity. A clean shutdown persists
+// cumulative statistics to pg_stat/ and reloads them at startup - for
+// pg_stat_statements that is governed by pg_stat_statements.save, on by default
+// - so the counters often do survive and the diff would have been valid. But we
+// cannot tell from the start time alone whether the shutdown was clean, whether
+// saving was enabled, or, where instances share a storage volume as Aurora
+// replicas do, whether the file the new postmaster loaded was written by this
+// instance at all. Losing one interval costs less than diffing against a
+// reference point that turns out not to belong to us.
```
msakrejda:
```suggestion
// A restart of the same instance is treated as a change too. A clean shutdown
// persists cumulative statistics and reloads them at startup (for
// pg_stat_statements that is governed by pg_stat_statements.save, on by default)
// so the counters often do survive and the diff would have been valid. But we
// cannot tell from the start time alone whether the shutdown was clean, whether
// saving was enabled, or, where instances share a storage volume as Aurora
// replicas do, whether the file the new postmaster loaded was written by this
// instance at all. Losing one interval is preferable to diffing against the wrong
// reference point.
```

state/instance_identity.go
```diff
@@ -0,0 +1,73 @@
...
+type PostgresInstanceIdentity struct {
+	// Start time of the postmaster we're connected to; changes when it restarts
```
msakrejda:
```suggestion
	// Start time of the postmaster we're connected to
```

state/instance_identity.go
```diff
@@ -0,0 +1,73 @@
...
+	PostmasterStartTime time.Time
+	// Address of the instance this connection landed on, from inet_server_addr().
+	// Empty for Unix socket connections, where it is not available.
```
msakrejda:
```suggestion
	// Empty for Unix socket connections, where it is not available (but also not
	// necessary, since a local connection string is unambiguous).
```

state/instance_identity.go
```diff
@@ -0,0 +1,73 @@
...
+// old Postgres version, a permission problem, a transient error), and in that
+// case we prefer to keep diffing as before rather than discard data on the
+// basis of information we do not have.
+func (i PostgresInstanceIdentity) Matches(other PostgresInstanceIdentity) bool {
+	if i.IsZero() || other.IsZero() {
+		return true
+	}
+	// Compare the timestamps with Equal rather than ==, since the driver hands us
+	// a separately allocated *time.Location for each row it scans, which makes ==
+	// report two readings of the same start time as different
```
msakrejda:
```suggestion
```

state/instance_identity_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+func TestPostgresInstanceIdentityMatches(t *testing.T) {
+	for _, test := range instanceIdentityMatchTests {
+		actual := test.prev.Matches(test.curr)
+		if actual != test.expected {
+			t.Errorf("%s: Matches(%+v, %+v)\nexpected %t\nactual %t\n", test.name, test.prev, test.curr, test.expected, actual)
+		}
+
+		// The check is used in both directions across the codebase, so it must be
+		// symmetric
```
msakrejda:
```suggestion
```

state/state.go
```diff
@@ -25,6 +25,10 @@ type SchemaStats struct {
 type PersistedState struct {
 	CollectedAt time.Time
 
+	// The instance this state was collected from, so a later run can tell whether
+	// its own reference point came from the same instance
```
msakrejda:
```suggestion
```

state/state.go
```diff
@@ -41,6 +45,10 @@ type PersistedHighFreqState struct {
 	PlanStats      PostgresPlanStatsMap
 	ServerIoStats  PostgresServerIoStatsMap
 
+	// The instance the above stats were collected from, so a later run can tell
+	// whether its own reference point came from the same instance
```
msakrejda:
```suggestion
```

runner/full.go
```diff
@@ -47,12 +47,35 @@ func collectDiffAndSubmit(ctx context.Context, server *state.Server, opts state.
...
+		// consecutive runs, and to keep the uint32 conversion from wrapping if the
+		// clock moved backwards
+	}
+
+	// Relation, index and database statistics are cumulative counters from one
+	// instance's memory, and we may have reached a different instance than the
+	// snapshot we would diff against - see PostgresInstanceIdentity. Diff against
+	// an empty reference point instead, which is how the first snapshot after
+	// collector startup behaves: absolute values are still reported, but no
+	// activity is attributed to this interval.
```
msakrejda:
```suggestion
	// Relation, index and database statistics are cumulative counters from one
	// instance's memory, and we may have reached a different instance than the
	// snapshot we would diff against. Diff against an empty reference point
	// instead: absolute values are still reported, but no activity is attributed
	// to this interval.
```

input/full_1min.go
```diff
@@ -36,9 +37,28 @@ func CollectAndDiff1minStats(ctx context.Context, c *postgres.Collection, connec
 		return newState, nil
 	}
 
+	// These are all cumulative counters from one instance's memory, and we may
+	// have reached a different instance than the run we would diff against - see
+	// PostgresInstanceIdentity. Skip this interval and let the state we just
+	// collected become the new reference point; diffing anyway would attribute
+	// each counter's entire lifetime to this one interval.
```
msakrejda:
```suggestion
	// These are all cumulative counters from one instance's memory, and we may
	// have reached a different instance than the run we would diff against. Skip
	// this interval and let the state we just collected become the new reference point.
```

msakrejda (commented):
Trim comments.

state/postgres_instance_identity.go
keiko713, line 1: The file name will be longer but maybe `postgres_instance_identity.go` is better here? Matching to the other state file names, as well as the struct name.

state/postgres_instance_identity.go
```diff
@@ -0,0 +1,69 @@
...
+//
+// A zero identity matches anything. We can fail to determine the identity (an
+// old Postgres version, a permission problem, a transient error), and in that
+// case we avoid using it.
+func (i PostgresInstanceIdentity) Matches(other PostgresInstanceIdentity) bool {
+	if i.IsZero() || other.IsZero() {
+		return true
+	}
+	return i.PostmasterStartTime.Equal(other.PostmasterStartTime) &&
+		i.ServerAddr == other.ServerAddr
```
keiko713, line 57: Soooooo this `ServerAddr` is a good info to have, esp. during the debug (you can tell which instance it's swapped to, or you can tell if it was a server restart, etc.), but I'm not sure if we should use this to check the match.
The postmaster start time is a microsecond precision, so I can't really think of the case where the postmaster start time is the same but the `ServerAddr` is different.

That said, I don't think I have a strong argument for _not having this check_, so I guess it's okay to keep as is.

state/postgres_instance_identity.go
```diff
@@ -0,0 +1,69 @@
...
+// A restart of the same instance is treated as a change too. A clean shutdown
+// persists cumulative statistics and reloads them at startup (for
+// pg_stat_statements that is governed by pg_stat_statements.save, on by default)
+// so the counters often do survive and the diff would have been valid. But we
+// cannot tell from the start time alone whether the shutdown was clean, whether
+// saving was enabled, or, where instances share a storage volume as Aurora
+// replicas do, whether the file the new postmaster loaded was written by this
+// instance at all. Losing one interval is preferable to diffing against the wrong
+// reference point.
+type PostgresInstanceIdentity struct {
```
keiko713, line 31: Not a scope of this PR, but it would be actually useful to store this info in postgres_server_stats in the future (e.g. we have `pg_stat_statements_reset` in ServerStatistic message)

keiko713 (approved):

state/postgres_instance_identity.go
msakrejda: So we do have microsecond precision on the start time, but that's from two independent clocks that could be skewed. The chance of a collision is still pretty tiny, but I don't think including `ServerAddr` hurts anything, so I'd rather keep it.

state/postgres_instance_identity.go
msakrejda: Yeah, good point. We can do that in a follow-up.

PR #850, opened 2026-07-30
Improve pg_stat_statements version mismatch handling

msakrejda:
We currently check for pg_stat_statements version mismatches, but we
do so with a hardcoded version list that can grow stale (it is
currently stale: it does not handle 18).

Instead, check the pg_available_extensions view for the latest
available version, and add a hint to update to that if
getStatementSource fails.

### DIFF
--- a/input/postgres/statements.go
+++ b/input/postgres/statements.go
@@ -57,6 +57,15 @@ SELECT nspname,
  WHERE pge.extname = 'pg_stat_statements'
 `
 
+// getStatementExtensionVersions returns the schema pg_stat_statements is installed in,
+// the minor version of the extension installed in the current database, and the minor
+// version the extension can be updated to (which is not set if the server doesn't
+// report an available version we understand)
+func getStatementExtensionVersions(ctx context.Context, db *sql.DB) (extSchema string, installedMinorVersion int16, availableMinorVersion null.String, err error) {
+	err = db.QueryRowContext(ctx, QueryMarkerSQL+statementExtensionVersionSQL).Scan(&extSchema, &installedMinorVersion, &availableMinorVersion)
+	return
+}
+
 func collectorStatement(query string) bool {
 	return strings.HasPrefix(query, QueryMarkerSQL)
 }
@@ -320,7 +329,7 @@ func getStatementSource(ctx context.Context, c *Collection, db *sql.DB, showtext
 		bundledExtMinorVersion = 3
 	}
 
-	err = db.QueryRowContext(ctx, QueryMarkerSQL+statementExtensionVersionSQL).Scan(&extSchema, &foundExtMinorVersion, &defaultExtMinorVersion)
+	extSchema, foundExtMinorVersion, defaultExtMinorVersion, err = getStatementExtensionVersions(ctx, db)
 	if err != nil && err != sql.ErrNoRows {
 		return statementSource{}, err
 	}
--- a/input/postgres/statements.go
+++ b/input/postgres/statements.go
@@ -49,9 +49,11 @@ SELECT dbid, userid, queryid, %s, query
 
 const statementExtensionVersionSQL string = `
 SELECT nspname,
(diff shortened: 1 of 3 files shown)


msakrejda:
The CI failure looks like an independent race. Restarting to se if that resolves it.

input/postgres/statements.go
```diff
@@ -265,26 +278,60 @@ func GetStatementTexts(ctx context.Context, c *Collection, db *sql.DB) (statemen
...
-	if c.PostgresVersion.Numeric >= state.PostgresVersion17 {
-		extMinorVersion = 11
+	var defaultExtMinorVersion null.String
+
+	// Version of the extension that ships with this Postgres version, used as a
+	// fallback when the server does not tell us which version is available
+	if c.PostgresVersion.Numeric >= state.PostgresVersion18 {
+		bundledExtMinorVersion = 12
+	} else if c.PostgresVersion.Numeric >= state.PostgresVersion17 {
+		bundledExtMinorVersion = 11
```
keiko713, line 324: If we're adding these nicely, maybe we should also add below (15 and 16 are using 1.10):

```
	} else if c.PostgresVersion.Numeric >= state.PostgresVersion15 {
		bundledExtMinorVersion = 10
```

input/postgres/statements.go
```diff
@@ -49,12 +49,23 @@ SELECT dbid, userid, queryid, %s, query
...
   FROM pg_extension pge
  INNER JOIN pg_namespace pgn ON pge.extnamespace = pgn.oid
+  LEFT JOIN pg_available_extensions pae ON pae.name = pge.extname
  WHERE pge.extname = 'pg_stat_statements'
 `
 
+// getStatementExtensionVersions returns the schema pg_stat_statements is installed in,
+// the minor version of the extension installed in the current database, and the minor
+// version the extension can be updated to (which is not set if the server doesn't
+// report an available version we understand)
```
keiko713: This "which is not set if the server doesn't report an available version we understand" is kinda lie. If the `pae.default_version` returns something like "1.10-myvendor", `availableMinorVersion` will be returned as `10-myvendor`, even though it's not the shape we understand.
Uhmm, but I guess what you meant here was more like `pae.default_version` returns something like "doesn't-even-have-a-dot"?

input/postgres/statements.go
```diff
@@ -293,27 +340,42 @@ func getStatementSource(ctx context.Context, c *Collection, db *sql.DB, showtext
...
 		extSchema = "public"
-		foundExtMinorVersion = extMinorVersion
+		foundExtMinorVersion = bundledExtMinorVersion
+	}
+
+	// Prefer the version the server reports as available, since that's what
+	// `ALTER EXTENSION pg_stat_statements UPDATE` updates to, and it keeps working for
+	// Postgres versions that are newer than this collector release. Fall back to the
+	// bundled version if the server doesn't report one, or reports one we can't parse
+	// (some providers add suffixes to the extension version).
```
keiko713: Is this true that some providers add suffixes to the extension version?
I think we've been successfully parsing `split_part(extversion, '.', 2)` at least to int16, meaning so far, at least `pg_extensions` won't return something with suffixes. It is possible that `defaultExtMinorVersion` does have something different here as it's a new table to query from (`pg_available_extensions`), but I'm curious if you found out the example version with some suffixes.

keiko713 (approved):

input/postgres/statements.go
msakrejda: Yeah, fair point. I've updated the comment for clarity.

input/postgres/statements.go
msakrejda: Technically the version doesn't have to be a number per Postgres extension contract conventions, so it doesn't hurt to handle that. E.g., Citus has some extension versions like "12.1-1" (not for pg_stat_statements). I have not seen it in the wild for pg_stat_statements itself.

PR #851, opened 2026-07-30
Allow adjusting log download interval (AWS, PlanetScale, self-hosted)

lfittl:
This introduces a new config setting "log_download_interval" / LOG_DOWNLOAD_INTERVAL, which controls how often logs are downloaded for platforms where streaming is not used. The main motivation for supporting this is to avoid rate limiting issues with providers like AWS, that implement an account-wide per-minute rate limit on certain API calls (like rds:DownloadDBLogFilePortion) that can be problematic when hosting 100+ database instances in the same account.

The log interval can be increased to up to 10 minutes on a per-server basis, causing delayed reporting of log events in pganalyze, but reducing rate limit problems for servers that have less activity.

For highly active servers this setting should be kept at the default (30s) to avoid losing data. The log download limit is increased from 10MB up to 50MB for longer intervals.


config/read.go
```diff
@@ -814,6 +823,14 @@ func preprocessConfig(config *ServerConfig) (*ServerConfig, error) {
 		}
 	}
 
+	if config.LogDownloadInterval == 0 {
+		// Not set at all (e.g. for a config that was constructed programmatically,
+		// instead of coming from getDefaultConfig)
+		config.LogDownloadInterval = MinLogDownloadInterval
+	} else if config.LogDownloadInterval < MinLogDownloadInterval || config.LogDownloadInterval > MaxLogDownloadInterval {
```
keiko713, line 832: If we failed to `strconv.ParseInt`, the error message will become "log_download_interval must be between 30 and 600 seconds, but is set to -1", which can be misleading. I think we have "not valid" messages other places so we can have that message when `config.LogDownloadInterval` is -1?

config/config_test.go
```diff
@@ -30,6 +31,46 @@ func TestGetDbURLRedacted(t *testing.T) {
...
+	}
+
+	const mb = 1024 * 1024
+	tests := []testItem{
+		// Unset (e.g. a programmatically constructed config) behaves like the default
+		{0, 30 * time.Second, 2*time.Minute + 30*time.Second, 10 * mb},
+		{30, 30 * time.Second, 2*time.Minute + 30*time.Second, 10 * mb},
+		{60, 1 * time.Minute, 3 * time.Minute, 20 * mb},
+		{120, 2 * time.Minute, 4 * time.Minute, 40 * mb},
+		// Partial intervals round down, and the parsing size is capped
```
keiko713: Uhmm, what do you mean by this "partial intervals round down"? I _think_ you're talking about the behavior that the partial interval can round down the max parsing size, but these tests are not really testing such case.
Maybe instead, adding the case like the following? It will be `int(100 secs / 30 secs) = 3`, then `3 * 10MB` which becomes 30MB, instead of something like 33.3MB.

```
{100, 100 * time.Second, 3*time.Minute + 40*time.Second, 30 * mb},
```

input/system/rds/logs.go
```diff
@@ -45,10 +45,11 @@ func DownloadLogFiles(ctx context.Context, server *state.Server, opts state.Coll
 		return server.LogPrevState, nil, nil, err
 	}
 
-	// Retrieve all possibly matching logfiles in the last two minutes, assuming
-	// the collector's scheduler that runs more frequently than that
-	linesNewerThan := time.Now().Add(-2 * time.Minute)
+	// Retrieve all logfiles that may have been written to since the previous log
+	// download, based on the configured log download interval
+	linesNewerThan := time.Now().Add(-server.Config.LogDownloadWindow())
```
keiko713, line 40: So, not only here but also other places, this `linesNewerThan` change from the 2 mins fixed window to the `interval + 2 mins`. This works okay with the shorter interval, but doesn't work well for the longer interval, especially any interval longer than 2 mins.

Let's think about the tolerance of failed/missed tick (e.g. AWS API rate limit, network blips, pganalyze API hiccups, the collector restarts - or actually collector reload too, as it'll "reset" the tick).

For the default, the interval is 30 seconds, and it goes back up to 2 mins 30 seconds, so even if it misses a few ticks, still the next successful one can go back up to 2 mins 30 seconds and likely read all the missed logs.
For the longest case, the interval is 10 mins, and it goes back up to 12 mins. If it misses the tick of 10:00 (which would have covered logs since the 9:50 tick), the next tick happens at 10:10, trying to read the log of 9:58-10:10, so it'll miss logs of 9:50-9:58. If it misses two ticks, the gap will be wider.

I don't think we need to make the tolerance of multiple missed tick (esp. for longer interval case), but I think at least we should try to aim this to survive one missed tick (e.g. `2*interval + 2 mins` instead of `1*interval + 2 mins`)?

input/system/planetscale/logs.go
```diff
@@ -26,14 +26,14 @@ const (
...
-	// This avoids an OOM when extremely large volumes of log data are returned,
-	// e.g. when starting the collector or during a burst of log activity.
-	maxLogParsingSize = 10 * 1024 * 1024
 )
 
+// Analyze and submit at most the trailing portion of the retrieved log data, as
+// determined by ServerConfig.MaxLogParsingSize()
+//
+// This avoids an OOM when extremely large volumes of log data are returned,
+// e.g. when starting the collector or during a burst of log activity.
```
keiko713: I think this comment now actually can go away, but we can mention about OOM in config.go as it's still good thing to comment. We might want to update the comment on RDS side as well, but it's also okay to leave it there I guess. This one stands out more because this comment is really not attached anywhere and feel out of the place.

keiko713 (approved):
Nice, looking good! Though maybe worth checking about `LogDownloadWindow`.

config/read.go
lfittl: Good point, added a special case "log_download_interval is invalid / not an integer" error message if its -1.

input/system/rds/logs.go
lfittl: Yeah, that's a good point. How about `2 * interval + 1 min`? That keeps the lookback window the same as today for the default (2 minutes), but scales it up more for longer intervals. The `+ 1 min` is really more meant to cover scheduling delays, and the 2x covers the missed ticks scenario.

config/config_test.go
lfittl: ~~That was more referring to the download window, but I can see how that's confusing. I'll clarify the comment and add your test case.~~

Actually, wait I got confused. Re-reading this.

lfittl: Added your test case and clarifying the comment (which is all about the parsing size, as you noted).

input/system/planetscale/logs.go
lfittl: I think I can actually move this down to the comment in front of `DownloadLogFiles` (to match RDS), since that had a stale 30 second reference.

keiko713:
Note: might be in your TODO already, but you should update the docs for this.

PR #853, opened 2026-08-04
Fix log line ordering in log analysis

msakrejda:
Our log line handling inadvertently reorders log lines during
analysis, but we rely on order to calculate byte offsets and log file
sizes. We use file sizes during snapshot processing.

See individual commits:
 - Add test showing that log line order is not preserved in analysis
 - Add test showing byte offset handling can lead to wrong log file size
 - Fix log line ordering issues by resorting

### DIFF
--- a/logs/stream/stream.go
+++ b/logs/stream/stream.go
@@ -178,27 +178,6 @@ func createLogFile(readyLogLines []state.LogLine, logger *util.Logger) (state.Lo
 	return logFile, nil
 }
 
-// handleLogAnalysis - Performs log analysis on submitted log lines on a per-backend basis
-func handleLogAnalysis(analyzableLogLines []state.LogLine) ([]state.LogLine, []state.PostgresQuerySample) {
-	// Split log lines by backend to ensure we have the right context
-	backendLogLines := make(map[int32][]state.LogLine)
-
-	for _, logLine := range analyzableLogLines {
-		backendLogLines[logLine.BackendPid] = append(backendLogLines[logLine.BackendPid], logLine)
-	}
-
-	var logLinesOut []state.LogLine
-	var querySamples []state.PostgresQuerySample
-
-	for _, analyzableLogLines := range backendLogLines {
-		backendLogLinesOut, backendSamples := logs.AnalyzeBackendLogLines(analyzableLogLines)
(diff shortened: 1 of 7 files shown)


logs/stream/stream.go
```diff
@@ -306,7 +285,7 @@ func AnalyzeStreamInGroups(logLines []state.LogLine, now time.Time, server *stat
 	}
 
 	logState := state.TransientLogState{CollectedAt: now}
-	logFile.LogLines, logState.QuerySamples = handleLogAnalysis(analyzableLogLines)
```
msakrejda, line 309: The above originally included some special-casing, but it has since converged with `logs.AnalyzeLogLines`, so we can just call that.

logs/stream/stream_test.go
```diff
@@ -370,12 +369,6 @@ func TestAnalyzeStreamInGroups(t *testing.T) {
 		if diff := cfg.Compare(pair.TransientLogState, TransientLogState); diff != "" {
 			t.Errorf("For %v: log state diff: (-want +got)\n%s", pair.TransientLogState, diff)
 		}
-		sort.SliceStable(pair.logFile.LogLines, func(i, j int) bool {
-			return pair.logFile.LogLines[i].ByteStart < pair.logFile.LogLines[j].ByteStart
-		})
-		sort.SliceStable(logFile.LogLines, func(i, j int) bool {
-			return logFile.LogLines[i].ByteStart < logFile.LogLines[j].ByteStart
-		})
```
msakrejda, line 378: These are no longer necessary; in hindsight they should have been a red flag.

runner/logs.go
```diff
@@ -334,7 +334,7 @@ func postprocessAndSendLogs(ctx context.Context, server *state.Server, opts stat
 		logFile := &transientLogState.LogFiles[idx]
 		if len(logFile.LogLines) > 0 {
 			logsExist = true
-			logFile.ByteSize = int64(logFile.LogLines[len(logFile.LogLines)-1].ByteEnd)
+			logFile.UpdateByteSize()
```
lfittl, line 337: I was wondering why we move this to its own method, but answering my own question, I can see its used in a test.

lfittl (approved):
:+1: Change makes sense to me.

keiko713 (approved):
Looks good to me too!

PR #854, opened 2026-08-04
Add support for both Neon and Supabase OTel log collection

ryanbooz:
Adds OTLP-based Postgres log ingestion for Neon and Supabase (both push logs to
the collector's OTLP receiver), plus the fixes needed to make Supabase work end to end.

### Log ingestion
- Handle gzip-encoded OTLP submissions (needed by both the Neon and Supabase drains).
- Parse Neon's OTLP Postgres logs.
- Parse Supabase's log-drain format (`input/system/supabase`): Postgres csvlog fields
  arrive under `metadata.parsed`, with the log message in the record's `EventName`.
  Supavisor pooler application logs (no `parsed`) are skipped.

### Supabase pooler support
- Strip the Supavisor `<role>.<project-ref>` username suffix when resolving the
  connected role, so `pg_monitor` is detected correctly and the helper-SQL `GRANT`s
  target the real role. (Also fixes the same latent issue for PlanetScale.)

### Log test
- Make the logs test timeout configurable (`db_log_test_timeout` / `LOG_TEST_TIMEOUT`).
  Default stays 10s; 30s for Supabase, whose log drain is batched and arrives with a consistent delay beyond 10s.

Tested against a live Supabase project over the session pooler: system classification,
replication, `pg_stat_statements`, and an auto_explain plan all flow through. Unit tests
cover the OTLP parsing and the timeout logic.

Requires a pganalyze build that includes the merged PR for identifying Neon and Supabase (https://github.com/pganalyze/pganalyze/pull/6072).


input/system/selfhosted/otel_handler.go
```diff
@@ -140,9 +175,20 @@ func handleOtlpLogsRequest(logsData *otlpLogs.LogsData, servers []*state.Server,
 	for _, r := range logsData.ResourceLogs {
 		for _, s := range r.ScopeLogs {
 			for _, l := range s.LogRecords {
-				if l.Body.GetKvlistValue() != nil {
+				if kv := l.Body.GetKvlistValue(); kv != nil {
+					if parsed := supabase.ParsedFields(kv); parsed != nil {
```
keiko713: It feels weird to do this check on top of the loop now. We can count if there is any `SystemType == "supabase"` server in servers,  then if such case, we can do this parse checking and stream.

input/system/supabase/logs.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+	}
+	return nil
+}
+
+// LogLineFrom maps a Supabase log drain Postgres record to a LogLine. The message -
+// including the collector identify marker - comes from the record's EventName (the
+// drain's event_message), not from "parsed", whose message/query fields are usually
+// empty. The timestamp comes from the record's TimeUnixNano.
+func LogLineFrom(l *otlpLogs.LogRecord, parsed *common.KeyValueList) state.LogLine {
```
keiko713: This one can be potentially merged together with the existing `logLineFromJsonlog` func in otel_handler.go (well, maybe rename is needed in that case as it's not only from the Jsonlog).

This one is missing the `detail` key handling (well, not sure if Supabase sends one but I feel like it should be sending), and I think the difference (`Content` and `OccurredAt`) can be handled over there, depending on if the `l` is passed or not (or if these fields are blank or not) - that would also silence the GH bot, as it's properly handled over there.

input/system/selfhosted/logs.go
```diff
@@ -477,6 +478,12 @@ func setupLogTransformer(ctx context.Context, wg *sync.WaitGroup, server *state.
 					logLine.LogLineNumberChunk = item.LogLineNumberChunk
 				}
 
+				if logLine.Database == "" {
+					if db := neon.LogDatabaseFallback(server.Config); db != "" {
+						logLine.Database = db
+					}
```
keiko713: This can simply be:

```suggestion
					logLine.Database = neon.LogDatabaseFallback(server.Config)
```

or we could also do something like:

```go
// once, before the for loop (next to linesNewerThan)
fallbackDatabase := neon.LogDatabaseFallback(server.Config)
// in the loop
if logLine.Database == "" {
      logLine.Database = fallbackDatabase
}
```

config/managed_platforms.go
```diff
@@ -31,3 +31,14 @@ func supabaseSystemID(config ServerConfig) string {
 	}
 	return host
 }
+
+// extractSupabaseUsername returns the Postgres role from a Supabase username. On
+// the Supavisor pooler the username carries the project ref as <role>.<project-ref>
+// for tenant routing, but we need just the role name part. Direct connections have no
+// suffix and are returned unchanged.
+func extractSupabaseUsername(username string) string {
+	if idx := strings.LastIndexByte(username, '.'); idx > 0 {
```
keiko713, line 49: This one is doing `strings.LastIndexByte` but above in `supabaseSystemID` func, it's doing `strings.Cut`, maybe better to unify this?

keiko713 (commented):
Several small comments but overall, it's looking good 🎉

input/system/supabase/logs.go
ryanbooz: Yeah - this was a decision on my part to not muddy the other stuff and pull (what seemed to be) Supabase specific stuff into it's on package. I have work in a separate branch to possibly enable getting logs over their Management API which makes this package more relevant. For now, I'll keep it somewhat because of convention, but take out the log line part until pulling it separately makes sense with two different potential streams of data. (Otel or Management API)

input/system/selfhosted/otel_handler.go
```diff
@@ -239,14 +300,23 @@ func warnAboutMultipleServers(servers []*state.Server, warnedAboutMultipleServer
 	*warnedAboutMultipleServers = true
 }
 
-func logLineFromJsonlog(record *common.KeyValueList, logParser state.LogParser) (state.LogLine, *state.LogLine) {
+// logLineFromJsonlog maps a jsonlog-style csvlog record (an OTel key/value list) to a
+// LogLine. Supabase's log drain uses the same field shape but carries the message and
+// timestamp on the OTel record itself (l) rather than in "parsed"; when l is provided
+// those take precedence over the message/log_time keys. A DETAIL, when present, is
+// returned as an additional log line.
+func logLineFromJsonlog(record *common.KeyValueList, logParser state.LogParser, l *otlpLogs.LogRecord) (state.LogLine, *state.LogLine) {
```
keiko713: I think this comment is now getting mixed up a bit. The jsonlog and csvlog are two different formats that Postgres emits, so there is not really "a jsonlog-style csvlog record" (or, if there is, it's only talking about Supabase one).
I suggest that we rename this func to `logLineFromStructuredFields`, then add the comment like below:

```
// logLineFromStructuredFields maps Postgres structured log fields (an OTel key/value
// list) to a LogLine. Postgres's jsonlog and csvlog formats share these field names:
// OTel pipelines deliver jsonlog records directly, while Supabase's log drain delivers
// csvlog-derived fields (metadata.parsed) with the message and timestamp carried on the
// OTel record itself — pass l there so EventName and TimeUnixNano fill those in.
```

keiko713 (approved):
One comment on func name + code comments, but other than that, looking good!

input/system/selfhosted/otel_handler.go
ryanbooz: Yep - definitely more clear. Thanks!

PR #855, opened 2026-08-11
🤖(amp): Accept fractional Actual Rows in PostgreSQL 18 EXPLAIN JSON

catkins:
🤖(amp):

## Summary

- accept fractional `Actual Rows` values while normalizing EXPLAIN JSON
- retain compatibility with integer row counts from earlier PostgreSQL versions
- cover PostgreSQL 18 values such as `4.60`, `1.00`, and `0.00`

## Context

PostgreSQL 18 changed `Actual Rows` in EXPLAIN output to always use two fractional digits for executed nodes. The collector currently decodes this field as `int64`, so Go rejects values such as `1.00` during query sample normalization.

Only `Actual Rows` is widened: `Actual Loops` and `Rows Removed by ...` continue to be emitted with zero fractional digits.

PostgreSQL change: https://github.com/postgres/postgres/commit/95dbd827f2edc4d10bebd7e840a0bd6782cf69b7

## Verification

- `go test ./logs/querysample/...`
- `make test`

### DIFF
--- a/logs/querysample/normalize.go
+++ b/logs/querysample/normalize.go
@@ -23,7 +23,7 @@ type planNodeGroupingSet struct {
 
 type planNode struct {
 	ActualLoops                 *int64                   `json:"Actual Loops,omitempty"`
-	ActualRows                  *int64                   `json:"Actual Rows,omitempty"`
+	ActualRows                  *float64                 `json:"Actual Rows,omitempty"` // PostgreSQL 18+ reports fractional rows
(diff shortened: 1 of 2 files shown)


lfittl (approved):
Thanks, good catch!

We should probably also add a full Postgres 18 example to the tests in the `testdata` folder, but I can take care of that separately.

catkins:
Thanks @lfittl 🙇

PR #857, opened 2026-08-14
Prevent invalid UTF-8 values in Postgres settings or other protobuf snapshot data from dropping entire snapshot

ryanbooz:
After releasing support for Supabase, a pre-existing database that monitoring had been working on stopped mysteriously submitting snapshots. Tests actually came back OK, but mixed in with the test output was "Error marshaling protocol buffers".

Eventually, with some help, I tracked it down to a GUC setting for `pgsodium`, a Supabase cryptography extension. It appears that the encryption key (maybe?) changed since I last had monitoring working and now contained invalid UTF-8 data. Therefore, the protobuf data couldn't be marshalled.

After a few iterations, this is the solution I settled on. At the very least, we need to guard against non-UTF8 data in settings (it appears to also have happened with backend types for Azure previously). But, potentially overkill, I also added a "catch all" to attempt and scrub data if the protobuf fails to marshall other data.

For settings, I thought following a similar principle to something like "\<truncated query\>" or "\<missing query text\>" was appropriate since we don't really know what the data is anyway and printing it has little value in my opinion (initially at least).

Replacing it in the snapshot also has the added benefit of not triggering any "settings have changed" logic that we might be planning if data like this changes periodically but the user has no control over it.

image

### DIFF
--- a/output/transform/postgres.go
+++ b/output/transform/postgres.go
@@ -2,11 +2,25 @@ package transform
 
 import (
 	"time"
+	"unicode/utf8"
 
 	snapshot "github.com/pganalyze/collector/output/pganalyze_collector"
 	"github.com/pganalyze/collector/state"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
+
+// when a setting value isn't valid UTF-8 (originally hit on pgsodium "boot_val") and therefore
+// can't be marshalled in a protobuf string field, replace with this consistent placeholder value
+// so that the snapshot can still be sent and we don't trigger value change logic every time
+// the non-UTF8 value changes.
+const settingValueNonUTF8 = "<non-utf8 value>"
+
+func settingValue(v string) string {
+	if !utf8.ValidString(v) {
+		return settingValueNonUTF8
+	}
+	return v
+}
 
 type OidToIdx map[state.Oid]int32
 
@@ -123,16 +137,16 @@ func transformPostgresConfig(s snapshot.FullSnapshot, transientState state.Trans
 		info := snapshot.Setting{Name: setting.Name}
 
 		if setting.CurrentValue.Valid {
-			info.CurrentValue = setting.CurrentValue.String
+			info.CurrentValue = settingValue(setting.CurrentValue.String)
 		}
 		if setting.Unit.Valid {
 			info.Unit = &snapshot.NullString{Valid: true, Value: setting.Unit.String}
 		}
 		if setting.BootValue.Valid {
-			info.BootValue = &snapshot.NullString{Valid: true, Value: setting.BootValue.String}
+			info.BootValue = &snapshot.NullString{Valid: true, Value: settingValue(setting.BootValue.String)}
 		}
 		if setting.ResetValue.Valid {
-			info.ResetValue = &snapshot.NullString{Valid: true, Value: setting.ResetValue.String}
+			info.ResetValue = &snapshot.NullString{Valid: true, Value: settingValue(setting.ResetValue.String)}
 		}
 		if setting.Source.Valid {
 			info.Source = &snapshot.NullString{Valid: true, Value: setting.Source.String}
--- a/output/transform/postgres.go
+++ b/output/transform/postgres.go
@@ -9,13 +9,26 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-// when a setting value isn't valid UTF-8 (originally hit on pgsodium "boot_val") and therefore
-// can't be marshalled in a protobuf string field, replace with this consistent placeholder value
-// so that the snapshot can still be sent and we don't trigger value change logic every time
-// the non-UTF8 value changes.
+// maskedSettingNames lists GUCs whose values are known to be unstable or not
+// meaningfully representable; their values are always replaced with a constant,
+// regardless of the current bytes. Masking by name keeps the value from flipping
+// snapshot-to-snapshot (which would churn config-change tracking) and avoids
+// shipping unstable data and increased noise. Add others here as we run into them.
+var maskedSettingNames = map[string]bool{
+	"pgsodium.getkey_script": true,
+}
+
+const settingValueMasked = "<masked value>"
+
+// settingValueNonUTF8 is the fallback for any *other* setting value that isn't valid
+// UTF-8 (and so can't be put in a protobuf string field), so the snapshot can still be
+// sent rather than dropped.
 const settingValueNonUTF8 = "<non-utf8 value>"
 
-func settingValue(v string) string {
+func settingValue(name, v string) string {
+	if maskedSettingNames[name] {
+		return settingValueMasked
+	}
 	if !utf8.ValidString(v) {
 		return settingValueNonUTF8
 	}
@@ -137,16 +150,16 @@ func transformPostgresConfig(s snapshot.FullSnapshot, transientState state.Trans
 		info := snapshot.Setting{Name: setting.Name}
 
 		if setting.CurrentValue.Valid {
-			info.CurrentValue = settingValue(setting.CurrentValue.String)
+			info.CurrentValue = settingValue(setting.Name, setting.CurrentValue.String)
 		}
 		if setting.Unit.Valid {
 			info.Unit = &snapshot.NullString{Valid: true, Value: setting.Unit.String}
 		}
 		if setting.BootValue.Valid {
-			info.BootValue = &snapshot.NullString{Valid: true, Value: settingValue(setting.BootValue.String)}
+			info.BootValue = &snapshot.NullString{Valid: true, Value: settingValue(setting.Name, setting.BootValue.String)}
 		}
 		if setting.ResetValue.Valid {
-			info.ResetValue = &snapshot.NullString{Valid: true, Value: settingValue(setting.ResetValue.String)}
+			info.ResetValue = &snapshot.NullString{Valid: true, Value: settingValue(setting.Name, setting.ResetValue.String)}
 		}
 		if setting.Source.Valid {
 			info.Source = &snapshot.NullString{Valid: true, Value: setting.Source.String}
--- a/output/upload.go
+++ b/output/upload.go
@@ -7,11 +7,14 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/pganalyze/collector/state"
 	"github.com/pganalyze/collector/util"
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
 )
 
 func SetupSnapshotUploadForAllServers(ctx context.Context, servers []*state.Server, opts state.CollectionOpts, logger *util.Logger) {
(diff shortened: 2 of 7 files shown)


output/transform/postgres.go
```diff
@@ -2,12 +2,26 @@ package transform
...
 
 	snapshot "github.com/pganalyze/collector/output/pganalyze_collector"
 	"github.com/pganalyze/collector/state"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+// when a setting value isn't valid UTF-8 (originally hit on pgsodium "boot_val") and therefore
+// can't be marshalled in a protobuf string field, replace with this consistent placeholder value
+// so that the snapshot can still be sent and we don't trigger value change logic every time
+// the non-UTF8 value changes.
```
msakrejda: Why do we want to avoid value change logic every time the non-UTF8 value changes? It's presumably not just changing randomly and our tracking semantics are still useful even if we can't represent the value.

Separately, looking into this, I think pgsodium is actually failing to preserve the memory backing the value correctly: it seems to be initialized in PostmasterContext, which is reset after a backend forks, so the pointer ends up pointing at random memory (which explains why we get invalid UTF8 out of it). In practice, this means this could point at something different for every backend. Sometimes it will happen to be valid UTF8, sometimes not. It could differ (nonsensically) snapshot to snapshot. I wonder if we shouldn't just... always mask the value?

output/upload.go
```diff
@@ -76,6 +79,98 @@ func snapshotUploadForServer(ctx context.Context, server *state.Server, logger *
...
+		case fd.IsList():
+			sanitizeUTF8List(fd, v.List())
+		case fd.IsMap():
+			sanitizeUTF8Map(fd, v.Map())
+		case fd.Kind() == protoreflect.MessageKind || fd.Kind() == protoreflect.GroupKind:
+			sanitizeInvalidUTF8(v.Message())
+		case fd.Kind() == protoreflect.StringKind:
+			if s := v.String(); !utf8.ValidString(s) {
+				fixFds = append(fixFds, fd)
+				fixVals = append(fixVals, strings.ToValidUTF8(s, utf8Replacement))
```
msakrejda, line 120: If we do keep this, `input/postgres/backends.go` and `input/postgres/backend_counts.go` call `.ToValidUTF8` with an empty string. Why something different here? It probably makes more sense to have a  replacement character than to blank out invalid bytes, but I don't see why these should be different.

output/upload.go
```diff
@@ -31,9 +34,9 @@ func snapshotUploadForServer(ctx context.Context, server *state.Server, logger *
 		case <-ctx.Done():
 			return
 		case s := <-server.FullSnapshotUpload:
-			data, err := proto.Marshal(s)
+			data, err := marshalSnapshot(s, logger)
```
msakrejda, line 37: I'm not convinced doing this preemptively for the whole snapshot is a good idea: we're masking potential problems (no one will see verbose messages outside of a test in practice) and adding a relatively expensive walk of the entire snapshot. I see it only happens if we fail a regular marshal, but I don't think it's the right trade-off. I'd rather we identify specific problems and address them tactically.

I concede this does have the advantage of solving this class of problems once and for all, but if we run into _diffferent_ problems that also manifest as invalid UTF8 (e.g., a processing error in the collector), tracking them down will be much harder with this in place.

output/transform/postgres.go
msakrejda: I reported this upstream: https://github.com/michelp/pgsodium/pull/126

output/upload.go
lfittl: Yeah, agreed, maybe this is not necessary, it seems like a lot of extra code.

output/transform/postgres.go
msakrejda: @lfittl and I talked about this, and the fact that it could vary snapshot-to-snapshot means we may actually want to check the GUC name against a list, and report "<bad value>" (not the right string, but you get the idea) for any GUC that may behave this way. Otherwise, this could be pretty confusing if it happens to be a valid UTF8 string.

ryanbooz: To make sure I'm tracking (and it makes sense I think) - one snapshot the value is fine, the next it's invalid, etc. - therefore switching between "good value"/"<non-UTF8 value>"/"good value", over and over, which isn't helpful. But displaying non-UTF8 data is likely not helpful either for most folks.

So, instead, if the GUC is in a list we maintain, replace it with a common string always and move on?

Ultimately I understand the value of that. The choice of trying to make a blanket "fix" was simply because it took a very long time without the walk functions to figure out which value was bad (you don't get that back from the marshall, obviously).

I know there are comments further down about that, but I think it's related. As a user, my snapshots won't work and I'm dead in the water until I report it, we figure out the field to add to the list, and then provide an update. Rare, for sure, but not great UX.

output/upload.go
ryanbooz: So yes, this only happens if the marshall fails. So...

On the flip side, it's not trivial to figure out which field was failing. We don't have any tooling or logging to get that out, and if we did, it would essentially have to be the walk of the snapshot.

Is it better UX for a customer that their snapshots keep going, or that they have to wait for us to identify it and push an update? Is there a path in between somehow?

Again, I think this is rare (we haven't hit it much from what I can tell), but when it happens users won't get data until we fix it.

output/transform/postgres.go
msakrejda: >To make sure I'm tracking (and it makes sense I think) - one snapshot the value is fine, the next it's invalid, etc. - therefore switching between "good value"/""/"good value", over and over, which isn't helpful. But displaying non-UTF8 data is likely not helpful either for most folks.

Almost—sorry I wasn't clear. I think what will happen in practice is that it will switch between different garbage values, _some_ of which will be invalid utf8. So instead, I'm suggesting we check the setting name, and for this specific setting (and any others we run into in the future), we just mask it as "`<unreadable>`" or something consistently.

>The choice of trying to make a blanket "fix" was simply because it took a very long time without the walk functions to figure out which value was bad (you don't get that back from the marshall, obviously).

That's fair. Maybe we can keep the code to walk the snapshot looking for invalid UTF8 just to give a better error?

output/upload.go
msakrejda: >Is it better UX for a customer that their snapshots keep going, or that they have to wait for us to identify it and push an update? Is there a path in between somehow?

This is a philosophical point to some extent, but long-term, yeah, I think so: it's better for the customer if we understand what's going on rather than sweeping a big class of potential problems under the rug.

ryanbooz: So, for now, I chose to keep this in case the marshal fails, but log the column. Generally I think this will be very rare based on more of our conversation, but will allow us to see that it happened as part of logging with test output, etc. Almost any user that sees it (particularly as part of onboarding/PoC) is very likely to point it out/ask IMO.

I realize it's additional code, so I'm still happy to bail and stick with the specific GUC list for now.

cc: @lfittl

output/upload.go
ryanbooz: done

output/transform/postgres.go
```diff
@@ -9,13 +9,26 @@ import (
...
-// can't be marshalled in a protobuf string field, replace with this consistent placeholder value
-// so that the snapshot can still be sent and we don't trigger value change logic every time
-// the non-UTF8 value changes.
+// maskedSettingNames lists GUCs whose values are known to be unstable or not
+// meaningfully representable; their values are always replaced with a constant,
+// regardless of the current bytes. Masking by name keeps the value from flipping
+// snapshot-to-snapshot (which would churn config-change tracking) and avoids
+// shipping unstable data and increased noise. Add others here as we run into them.
+var maskedSettingNames = map[string]bool{
+	"pgsodium.getkey_script": true,
```
msakrejda, line 18: It might be good to reference https://github.com/michelp/pgsodium/pull/126 in a comment.

output/upload.go
msakrejda: So I still disagree philosophically, but I reviewed it and it looked good. Then I had Claude review it and it found a few things (lightly edited for clarity):

---

- Map key collision drops data. Two bad keys that scrub to the same valid key collapse into one entry. The second write silently overwrites the first.
- Retry runs on every marshal error, not only UTF-8 ones. Cheap fix is to retry only when len(fixed) > 0. The error text is not a typed error, so string matching is the only other option.
- Logged paths include map key contents via %v and %q. It's confusing to include values here.
- The debug JSON paths in full.go and compact.go call protojson.Marshal before this and will still fail on the same data without indication of where the failure is.
- utf8Replacement in upload.go is now just an alias for util.InvalidUTF8Replacement. Use the util constant directly.

---

But it also suggested a simpler way to walk the snapshot and just log the offending paths in the error (which is roughly what I was suggesting above, albeit I didn't realize we could use this simpler implementation). I tweaked that a bit and pushed it up [here](https://github.com/pganalyze/collector/compare/fix-invalid-utf8-snapshot-values...fix-invalid-utf8-snapshot-values-msakrejda?expand=1). (This also handles the dry run case.) Thoughts?

In either case, I'll leave it to @lfittl to approve since he had the same concerns about the snapshot-walking approach.

lfittl: Yeah, I discussed this with Ryan briefly in our 1:1 last week - basically I'm -0.5 on having this (replacing the bad contents) mainly because of code complexity, and because I don't think its actually that common to have this issue.

I do like the idea of being more specific about which part of the protobuf is a problem, since that would significant cut down on fixing root causes if this does happen again. It seems your implementation is indeed simpler, so I think that direction makes sense to me if Ryan agrees.

keiko713: I chatted about this a bit with Lukas in our 1:1, and here's my take.

I'm personally not comfortable with either the auto-repair or the snapshot-walk version, as both of them look pretty complex and affect core parts.
Looks like we've had a handful of support tickets with `Error marshaling protocol buffers` over time, but I'd say that it's still rare enough that we can look into each case as it comes up.

It is, however, a problem that it is hard to debug once we hit this error. I think we can take some baby steps here as a stopgap for debugging.
For example, `--dry-run` converts the snapshot with `protojson.Marshal`, and its error names the message type and field. With the pgsodium value, it says `proto: field pganalyze.collector.NullString.value contains invalid UTF-8`. Well, sadly this isn't super useful for the pgsodium case, since `NullString` is a wrapper type used in several places, but for most fields it names the actual type/field (e.g. `proto: field pganalyze.collector.Backend.backend_type contains invalid UTF-8`).

So my suggestion for this PR:

1. Keep the `pgsodium.getkey_script` mask in `transformPostgresConfig`, masked by name as discussed, with a comment linking michelp/pgsodium#126 (still feels a bit optional, as it is not blocking anyone at the moment).
2.  Run the `backend_type` cleanup in `backends.go` and `backend_counts.go` on every platform instead of only Azure, keeping `""` as the replacement (I think this one is safe to expand).
3. In `upload.go`, print the error and point to `--dry-run`:

   ```diff
    data, err := proto.Marshal(s)
    if err != nil {
   -	logger.PrintError("Error marshaling protocol buffers")
   +	logger.PrintError("Error marshaling protocol buffers: %s (run with --dry-run to see the affected field)", err)
    	continue
    }
   ```

   The log line would then look like this:

   ```
   Error marshaling protocol buffers: string field contains invalid UTF-8 (run with --dry-run to see the affected field)
   ```

   This should give a good starting point for where to look (with `--dry-run`).

PR #859, opened 2026-08-21
Avoid disrupting snapshot collection when uploads are slow/delayed

lfittl:
This is mainly focused on the legacy HTTP upload mechanism (see first two commits) but also includes a change for shared snapshot processing path to make the snapshot channel buffered (with a small retention of max 2 full snapshots, 10 compact snapshots) instead of unbuffered, and explicitly drop the snapshot if we block beyond that, which is very unlikely. We likely want to integrate that last change with the work in #822 / consider the totality of those two changes together.

### DIFF
--- a/output/upload.go
+++ b/output/upload.go
@@ -108,7 +108,7 @@ func uploadViaWebsocketOrHttp(ctx context.Context, server *state.Server, logger 
 		if err != nil {
 			return err
 		}
-		submitSnapshot(ctx, server, opts, logger, s3Location, collectedAt, compactSnapshot)
+		return submitSnapshot(ctx, server, opts, logger, s3Location, collectedAt, compactSnapshot)
 	}
 	return nil
 }
--- a/output/upload.go
+++ b/output/upload.go
@@ -23,6 +23,12 @@ func SetupSnapshotUploadForAllServers(ctx context.Context, servers []*state.Serv
 	}
 }
 
+// Maximum time a legacy HTTP snapshot upload may take before it is abandoned,
+// so a slow or unreachable server does not indefinitely stall the per-server
+// upload queue (and with it all snapshot types, since uploads run sequentially)
+const fullSnapshotUploadTimeout = 1 * time.Minute
+const compactSnapshotUploadTimeout = 20 * time.Second
+
 func snapshotUploadForServer(ctx context.Context, server *state.Server, logger *util.Logger, opts state.CollectionOpts) {
 	var compactLogTime time.Time
 	compactLogStats := make(map[string]uint8)
@@ -37,7 +43,7 @@ func snapshotUploadForServer(ctx context.Context, server *state.Server, logger *
 				continue
 			}
 
-			err = uploadViaWebsocketOrHttp(ctx, server, logger, opts, data, s.SnapshotUuid, s.CollectedAt.AsTime(), false)
+			err = uploadViaWebsocketOrHttp(ctx, server, logger, opts, data, s.SnapshotUuid, s.CollectedAt.AsTime(), false, fullSnapshotUploadTimeout)
 			if err != nil {
 				logger.PrintError("Error uploading snapshot: %s", err)
 			} else if !opts.TestRun {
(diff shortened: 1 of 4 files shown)


output/upload.go
```diff
@@ -92,7 +98,7 @@ func summarizeCounts(counts map[string]uint8) string {
 	return details
 }
 
-func uploadViaWebsocketOrHttp(ctx context.Context, server *state.Server, logger *util.Logger, opts state.CollectionOpts, data []byte, snapshotUUID string, collectedAt time.Time, compactSnapshot bool) error {
+func uploadViaWebsocketOrHttp(ctx context.Context, server *state.Server, logger *util.Logger, opts state.CollectionOpts, data []byte, snapshotUUID string, collectedAt time.Time, compactSnapshot bool, httpUploadTimeout time.Duration) error {
```
keiko713, line 100:
```suggestion
func uploadViaWebsocketOrHttp(ctx context.Context, server *state.Server, logger *util.Logger, opts state.CollectionOpts, data []byte, snapshotUUID string, collectedAt time.Time, httpUploadTimeout time.Duration) error {
```

Since we're touching here, it's the best to actually retire the `compactSnapshot bool`, which is currently no longer used.

state/state.go
```diff
@@ -340,8 +340,11 @@ func MakeServer(config config.ServerConfig, testRun bool) *Server {
 		ActivityStateMutex:    &sync.Mutex{},
 		HighFreqStateMutex:    &sync.Mutex{},
 		CollectionStatusMutex: &sync.Mutex{},
-		FullSnapshotUpload:    make(chan *pganalyze_collector.FullSnapshot),
-		CompactSnapshotUpload: make(chan *pganalyze_collector.CompactSnapshot),
+		// Buffered so a slow upload does not immediately block the collection
+		// runners (and with them the collection schedulers); when the buffer is
+		// full, new snapshots are discarded with an error instead of blocking
+		FullSnapshotUpload:    make(chan *pganalyze_collector.FullSnapshot, 2),
+		CompactSnapshotUpload: make(chan *pganalyze_collector.CompactSnapshot, 10),
```
keiko713: Since the compact snapshot is used by activity, logs, system, query_runs, feels like 10 is a bit short. Given this will be applied for the websocket too, maybe we could bump it to 20 or so at least?

output/upload.go
```diff
@@ -23,6 +23,12 @@ func SetupSnapshotUploadForAllServers(ctx context.Context, servers []*state.Serv
 	}
 }
 
+// Maximum time a legacy HTTP snapshot upload may take before it is abandoned,
+// so a slow or unreachable server does not indefinitely stall the per-server
+// upload queue (and with it all snapshot types, since uploads run sequentially)
+const fullSnapshotUploadTimeout = 1 * time.Minute
+const compactSnapshotUploadTimeout = 20 * time.Second
```
keiko713: This 20 seconds has to cover TCP connect, TLS handshake, multipart body transfer, the server reading the body and enqueueing, the response. THEN, the entire submitSnapshot round trip (which, we actually should be refactoring this to avoid doing this when it's not test - it's a very much wasting the resource to get a simple 200 OK from the pganalyze app) plus every retryablehttp attempt and its backoff sleeps.

Raising this time to 1 minute will save any requests that are taking between 20-60 seconds (which would have been completely thrown away otherwise), still better than the current cap (~20 mins, with RetryMax 4 × 120s per-request timeout plus backoff, across two sequential HTTP calls), and bonus point, it will simplify the code as we will have a single limit of 1 minute.

On the other hand, I think it makes sense to have the number difference for the buffers in state/state.go - full snapshots doesn't need 10 buffers. and compact snapshots does definitely need more than 2 buffers (and these numbers look good to me).

keiko713 (approved):
Some feedback on the timeout number, but otherwise looks good!

output/upload.go
lfittl: I don't think that's right - `compactSnapshot` still gets passed to `submitSnapshot` which then decides to either call `/v2/snapshots` or `/v2/snapshots/compact` based on that. Those URLs may not actually make a difference in practice today on the server side, but that seems an independent change to me, so I'll leave as-is for now.

keiko713: Totally fine to leave as is for this PR, I think we actually should be doing the whole refactoring I mentioned in the other comment (rather than deleting one small arg in this PR):

> THEN, the entire submitSnapshot round trip (which, we actually should be refactoring this to avoid doing this when it's not test - it's a very much wasting the resource to get a simple 200 OK from the pganalyze app)

Maybe I should make a PR for it.

PR #860, opened 2026-08-21
WebSocket: Detect dead connections via write deadlines and ping/pong

lfittl:
Previously the WebSocket had no write deadline and no keepalive, so a silently dropped TCP session (e.g. by a NAT or firewall) would block the writer in WriteMessage until the kernel's TCP retransmission timeout (possibly 15+ minutes), stalling snapshot uploads while the connection still reported as connected.

Set a 30 second write deadline on all writes, send a ping every 30 seconds when idle, and require any message from the server within 70 seconds (the pganalyze server sends a keepalive Pong every 20 seconds, which now also refreshes this deadline).

Effectively this causes a dead connection to be closed within about a minute, letting uploads fall back to legacy HTTP and the existing reconnect logic re-establish the WebSocket.

### DIFF
--- a/util/reconnecting_socket.go
+++ b/util/reconnecting_socket.go
@@ -11,9 +11,12 @@ import (
 )
 
 type ReconnectingSocket struct {
-	// Channels shared with the caller
-	Read  chan []byte
-	Write chan []byte
+	// Channel shared with the caller for incoming messages
+	Read chan []byte
+
+	// Channel for handing outgoing messages to the writer goroutine
+	// (use WriteMessage to send out messages)
+	write chan socketWrite
 
 	// Initial arguments
 	dialer  websocket.Dialer
@@ -30,7 +33,17 @@ type ReconnectingSocket struct {
 	shutdown  chan struct{}
 }
 
+// socketWrite - A single outgoing message, together with the channel used to
+// hand the write's outcome back to the caller
+type socketWrite struct {
+	data []byte
+	// Must be buffered (capacity 1), so the writer goroutine never blocks
+	// handing back the result to a caller that has stopped waiting
+	result chan error
+}
+
 var ErrorConnectRateLimited = errors.New("Skipping connection attempt because of previous 4XX error")
+var ErrorWriteNotAccepted = errors.New("Timeout waiting for websocket connection to accept message")
 
 // Timeouts to detect dead connections (e.g. a NAT/firewall silently dropping
 // the TCP session), which would otherwise only fail after the kernel's TCP
@@ -51,7 +64,7 @@ const (
 func NewReconnectingSocket(ctx context.Context, logger *Logger, dialer websocket.Dialer, url string, headers map[string][]string, reconnectInterval time.Duration, clientErrorTimeout time.Duration) *ReconnectingSocket {
 	w := &ReconnectingSocket{
 		Read:      make(chan []byte),
-		Write:     make(chan []byte),
+		write:     make(chan socketWrite),
 		ctx:       ctx,
 		dialer:    dialer,
 		url:       url,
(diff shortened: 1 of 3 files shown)


util/reconnecting_socket.go
```diff
@@ -151,22 +164,37 @@ func (w *ReconnectingSocket) connect(ctx context.Context) (int, error) {
...
+		defer ticker.Stop()
 		for {
 			select {
 			case <-connCtx.Done():
 				w.closeConnection()
 				return
 			case data := <-w.Write:
-				err = conn.WriteMessage(websocket.BinaryMessage, data)
+				conn.SetWriteDeadline(time.Now().Add(socketWriteTimeout))
+				err := conn.WriteMessage(websocket.BinaryMessage, data)
```
keiko713: So, if we reach the deadline here, this will still report "Submitted full snapshot successfully", which is kinda false positive (reaching other deadline won't cause the submit failure, so only here is a problem).
This is kinda pre-existing issue too, and I bet this was not helping us to debug with the collector logs in the past.

For example, this can happen:

```
  t=0.000  writer takes data off the channel  ← send unblocks
  t=0.000  uploadViaWebsocketOrHttp returns nil
  t=0.001  "Submitted full snapshot successfully"     ← success logged here
  t=0.001  writer: WriteMessage(...)
  t=0.002  "Error writing to websocket: with some error"     ← error logged here
```

With this code, this can happen:

```
  t=0.000  writer takes data off the channel  ← send unblocks
  t=0.000  uploadViaWebsocketOrHttp returns nil
  t=0.001  "Submitted full snapshot successfully"     ← success logged here
  t=0.001  writer: SetWriteDeadline(+30s), WriteMessage(...)
  t=30.00  "Error writing to websocket: ... i/o timeout"   ← logged 30s later
```

Here is something I sketched with Claude Code to prevent this from happening.

```diff
diff --git a/output/upload.go b/output/upload.go
index 0c7b0c45..d7eabc51 100644
--- a/output/upload.go
+++ b/output/upload.go
@@ -100,7 +100,11 @@ func uploadViaWebsocketOrHttp(ctx context.Context, server *state.Server, logger
 
 	if server.WebSocket.Connected() {
 		logger.PrintVerbose("Uploading snapshot to websocket")
-		server.WebSocket.Write <- compressedData.Bytes()
+		result := make(chan error, 1)
+		server.WebSocket.Write <- util.SocketWrite{Data: compressedData.Bytes(), Result: result}
+		if err := <-result; err != nil {
+			return fmt.Errorf("websocket write failed: %s", err)
+		}
 	} else if server.Config.APIRequireWebsocket {
 		return errors.New("Error uploading snapshot: WebSocket not connected")
 	} else {
diff --git a/util/reconnecting_socket.go b/util/reconnecting_socket.go
index c69e8bf4..04c900f9 100644
--- a/util/reconnecting_socket.go
+++ b/util/reconnecting_socket.go
@@ -13,7 +13,7 @@ import (
 type ReconnectingSocket struct {
 	// Channels shared with the caller
 	Read  chan []byte
-	Write chan []byte
+	Write chan SocketWrite
 
 	// Initial arguments
 	dialer  websocket.Dialer
@@ -30,6 +30,11 @@ type ReconnectingSocket struct {
 	shutdown  chan struct{}
 }
 
+type SocketWrite struct {
+	Data   []byte
+	Result chan error // must be buffered (capacity 1); nil for fire-and-forget
+}
+
 var ErrorConnectRateLimited = errors.New("Skipping connection attempt because of previous 4XX error")
 
 // Timeouts to detect dead connections (e.g. a NAT/firewall silently dropping
@@ -51,7 +56,7 @@ const (
 func NewReconnectingSocket(ctx context.Context, logger *Logger, dialer websocket.Dialer, url string, headers map[string][]string, reconnectInterval time.Duration, clientErrorTimeout time.Duration) *ReconnectingSocket {
 	w := &ReconnectingSocket{
 		Read:      make(chan []byte),
-		Write:     make(chan []byte),
+		Write:     make(chan SocketWrite),
 		ctx:       ctx,
 		dialer:    dialer,
 		url:       url,
@@ -171,9 +176,12 @@ func (w *ReconnectingSocket) connect(ctx context.Context) (int, error) {
 			case <-connCtx.Done():
 				w.closeConnection()
 				return
-			case data := <-w.Write:
+			case msg := <-w.Write:
 				conn.SetWriteDeadline(time.Now().Add(socketWriteTimeout))
-				err := conn.WriteMessage(websocket.BinaryMessage, data)
+				err := conn.WriteMessage(websocket.BinaryMessage, msg.Data)
+				if msg.Result != nil {
+					msg.Result <- err
+				}
 				if err != nil {
 					w.logger.PrintError("Error writing to websocket: %s", err)
 					w.closeConnection()

```

keiko713 (approved):
Having timeouts on the websocket side as well sounds good to me 👍

seanlinsley (approved):

util/reconnecting_socket.go
lfittl: Makes sense, I've incorporated that.

PR #861, opened 2026-08-22
Share DNS resolution between a server's snapshot connections

msakrejda:
This is a follow-up to #846, broken out from that commit and tweaked a
bit. This is pretty much Claude's work and very lightly reviewed at
this point; I'm only including it here for reference. I plan to focus
on other work and won't pursue this PR.

---

Each snapshot type opens its own connection, so each one resolves
db_host independently. However, Aurora reader endpoints follow
failovers, so consecutive connections can reach different instances.
Within a single minute, the query statistics connection and the full
snapshot connection can disagree.

Resolve the hostname once per server and share the result, so the
connections stay separate but talk to the same instance. The address is
supplied to lib/pq through a custom dialer, because pq takes the dial
address from the dialer but reads the TLS server name straight out of
the connection parameters. Passing it to the dialer leaves hostname
verification intact.

The pinned address is the one a connection last actually succeeded to,
not the first address of a lookup: pinning the lookup result could
wedge collection permanently on a host whose lookup orders an
unreachable address first (IPv6 on an IPv4-only network). The rest of
the latest answer is kept as dial fallbacks with net.Dialer-style
partial deadlines, so a dead pinned instance costs one failed attempt
inside a single dial rather than a failed snapshot, and the pin moves
to whatever connected. The dialer consults the shared resolution at
dial time, so long-lived pools (the log follower) reconnect through the
current pin as well.

The pin is kept as long as the hostname keeps resolving to it: the
hostname is re-resolved at most once a minute, and the pin is evicted
once ten consecutive answers omit it. That catches the case where the
pinned instance still accepts connections but is no longer what the
hostname points at - after an Aurora failover, the old writer keeps
serving connections in its new reader role - within about ten minutes,
while endpoints with a stable answer re-confirm the pin on every lookup
and are never churned. Requiring several consecutive misses keeps
rotating multi-reader endpoints, where a valid pin is missing from most
single answers, from evicting a healthy pin needlessly. A failed lookup
leaves the pin alone, so a resolver outage cannot take down a working
pin. A failed connection drops the pin only for network-level errors: a
bad password or a missing database says nothing about the address, and
must not push the next snapshot connection onto a different instance.

Hosts written as an IP, Unix sockets, and the Cloud SQL and AlloyDB
drivers (which are given an instance name rather than a hostname) are
left alone. This reduces how often the instance changes under us, but
the reference point is still checked against the instance identity.

### DIFF
--- a/input/postgres/establish_connection.go
+++ b/input/postgres/establish_connection.go
@@ -3,23 +3,33 @@ package postgres
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
+	"net"
 
-	"github.com/pganalyze/collector/config"
+	"github.com/lib/pq"
 	"github.com/pganalyze/collector/state"
 	"github.com/pganalyze/collector/util"
 )
 
 func EstablishConnection(ctx context.Context, server *state.Server, logger *util.Logger, opts state.CollectionOpts, databaseName string) (connection *sql.DB, err error) {
-	connection, err = connectToDb(ctx, server.Config, opts, databaseName)
+	connection, err = connectToDb(ctx, server, logger, opts, databaseName)
 	if err != nil {
 		if err.Error() == "pq: SSL is not enabled on the server" && (server.Config.DbSslMode == "prefer" || server.Config.DbSslMode == "") {
 			server.Config.DbSslModePreferFailed = true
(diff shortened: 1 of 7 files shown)


msakrejda:
Closing as I don't plan to purse this. I wanted to leave the PR as a record of the problem and a potential solution.

PR #862, opened 2026-08-30
Retire submitSnapshot for http upload

keiko713:
A follow-up of https://github.com/pganalyze/collector/pull/859

While I was reviewing it, I noticed that we're doing an extra HTTP request with `submitSnapshot`. This only has the meaning when `opts.TestRun` is on, so let's only make an extra HTTP request in that case.

We could also do further cleanups, like the following:

* now we don't really use the `s3Location` from `uploadSnapshot`, we can retire things around there too
* `/v2/snapshots/test` is not doing much on the server side (only updating `last_test_snapshot_at`), also we don't do this with websocket case, so we can revisit what we want to do it

### DIFF
--- a/output/upload.go
+++ b/output/upload.go
@@ -116,7 +116,7 @@ func uploadViaWebsocketOrHttp(ctx context.Context, server *state.Server, logger 
 			return err
 		}
 		if opts.TestRun {
-			return markTestRun(uploadCtx, server, opts)
+			return markTestRun(uploadCtx, server, opts, logger)
 		}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


msakrejda (commented):
Doesn't this lose the per-server ` {"message": "Test snapshot successful", "server_url": ...}` messages, since we ignore test snapshot submission bodies on 200? Is that okay?

Other than that, this looks fine.

keiko713:
Ah good catch, brought it back.

msakrejda (approved):

PR #863, opened 2026-08-31
Cache RDS describe results per AWS account and cache API sessions

### LINKED ISSUE #577: Inefficient usage of AWS DescribeDBInstances and DescribeDBClusters API
Currently, the collector calls [FindRdsInstance](https://github.com/pganalyze/collector/blob/db03b2758711d0abb028feea243f633c9677a6d8/input/system/rds/system.go#L34-L39) once per full snapshot if [CollectSystemInformation](https://github.com/pganalyze/collector/blob/db03b2758711d0abb028feea243f633c9677a6d8/input/full.go#L154C26-L154C50) is configured. This leads to at least one AWS API call per database every collection period (multiple if using Aurora). 

When the collector is configured to collect on hundreds of DBs in an AWS account, the O(N) AWS API calls all happening one after another can lead to throttling:
```
Rds/System: Encountered error when looking for instance: Throttling: Rate exceeded
	status code: 400, request id: <REDACTED>
```

This leads to missed system state snapshots and affects other tools trying to call the same APIs within the account.

AWS's DescribeDBInstances and DescribeDBClusters APIs actually allow describing up to 100 instances or clusters at a time in one API call by the use of filters see [here](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBInstances.html). What are our thoughts about modifying the collector logic to handle multiple system snapshots at once to use fewer AWS API calls?

lfittl:
Introduce a shared cache for AWS sessions as well as results of rds:DescribeDBInstances and rds:DescribeDBClusters to avoid rate limit issues when monitoring hundreds of database servers within the same AWS account.

Cached clients for specific services can either be used directly, or for RDS calls the special-purpose GetRdsInstance/GetRdsCluster/etc are available, which also cache the result set for 60 seconds. Note this effectively means reaction to failover events is delayed, so a configured cluster endpoint might continue pointing to the prior target for up to 60 seconds for log downloads.

An edge case is worth calling out: If the configured IAM policy restricts which servers are accessible, the Describe* calls may fail, since they are no longer limited to specific identifiers/cluster IDs. In that case we fall back to doing an individual look up, and log a verbose notice.

The cache is maintained globally and reset on config reloads, and first initialized when a particular per-account AWS API session is initialized.

In passing this replaces the previous TTL map mechanism (used for caching some error cases) with a different TTL cache that is now used by this global AWS cache. The use case where a wrong instance name is configured is now dealt with separately: Assuming we are able to retrieve the full list of instances in an account, we can conclude from the list that an instance does not exist, and return without incurring any extra API calls.

Fixes #577


input/system/rds/logs.go
```diff
@@ -147,19 +144,12 @@ func getAwsDbInstanceID(ctx context.Context, config config.ServerConfig, awsCfg
...
-	// we want to react quickly to failover events.
-	cachedError := DescribeDBClustersErrorCache.Get(config.AwsDbClusterID)
-	if cachedError != "" {
-		return "", errors.New(cachedError)
-	}
-
-	instance, err := awsutil.FindRdsInstance(ctx, config, awsCfg)
+	// Errors are cached (briefly) by the account's describe cache, so a failing
+	// DescribeDBClusters call is not repeated on every log download, while
+	// still reacting quickly to failover events
```
msakrejda: Is this the right place for this comment now that caching is encapsulated?

util/awsutil/account.go
```diff
@@ -0,0 +1,227 @@
...
+	"github.com/pganalyze/collector/config"
+	"github.com/pganalyze/collector/util"
+)
+
+// AccountKey identifies a distinct AWS API client configuration. Server
+// configurations that agree on all these fields can safely share credentials
+// and API clients (and their cached API results).
+//
+// Note that this does not include "Account ID" because we typically don't know
+// that directly in the config, but instead its inferred from the context
```
msakrejda:
```suggestion
// that directly in the config, but instead it's inferred from the context
```

util/awsutil/rds.go
```diff
@@ -14,61 +14,32 @@ import (
...
 		}
 	}
 	return nil, fmt.Errorf("Failed to find RDS instance using endpoint-based search for host \"%s\" and port %d", host, port)
 }
 
 // FindRdsInstance finds and returns an RDS DBInstance for the given server config.
-func FindRdsInstance(ctx context.Context, serverCfg config.ServerConfig, awsCfg aws.Config) (*rdstypes.DBInstance, error) {
-	client := NewRdsClient(awsCfg, serverCfg)
-
+// The result is shared with other callers and must not be modified.
```
msakrejda, line 35: Is it worth copying the result to avoid future bugs here?

util/awsutil/ttl_cache.go
```diff
@@ -0,0 +1,98 @@
...
+	"context"
+	"fmt"
+	"sync"
+	"time"
+)
+
+// How long cached results are reused before refetching. This bounds both how
+// quickly we react to changes on the AWS side (e.g. failovers) and how often
+// many servers in the same account hit the (account-wide) API rate limits.
+const cacheTTL = 60 * time.Second
```
msakrejda, line 13: Should this be configurable? It determines the balance between API usage and failover responsiveness, and I'm not sure if everyone will want the same trade-off.

util/awsutil/ttl_cache.go
```diff
@@ -0,0 +1,98 @@
...
+
+// Backstop timeout for a single shared fetch. This bounds how long a hung
+// fetch can stall snapshots / log downloads for servers in the account.
+const cacheFetchTimeout = 60 * time.Second
```
msakrejda, line 17: Should this be configurable? I think it's probably okay not to expose it for now.

util/awsutil/ttl_cache_test.go
```diff
@@ -0,0 +1,228 @@
...
+	close(fetchFinish)
+
+	value, err := cache.get(context.Background(), "key", func(ctx context.Context) (string, error) {
+		atomic.AddInt32(&fetches, 1)
+		return "other", nil
+	})
+	if err != nil || value != "value" {
+		t.Errorf("expected \"value\" despite fetching caller's cancellation, got %q (err: %v)", value, err)
+	}
+	if fetches != 1 {
```
msakrejda, line 182: I don't think it matters here, but 
```suggestion
	if atomic.LoadInt32(&fetches) != 1 {
```

util/awsutil/amazon.go
msakrejda, line 1: Should we note that anything read in `GetAwsConfig` must be added to `AccountKey` to avoid caching mix-ups in the future?

util/awsutil/ttl_cache.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	entry, ok := c.entries[key]
+	if ok {
+		select {
+		case <-entry.done: // Completed fetch, check for expiry
+			if time.Since(entry.fetchedAt) < cacheTTL {
+				c.mutex.Unlock()
+				return entry.value, entry.err
+			}
+			// Expired, fall through and fetch again
```
msakrejda, line 60: If the subsequent fetch fails with a transient error, we replace a stale value with an error. Is that the right trade-off?

util/awsutil/ttl_cache.go
msakrejda, line 1: Nice, this is clear and thorough.

msakrejda (approved):
Looks good. We could also add caching to `GetRdsParameter`, but that can wait.

This can also be an increase in API calls if you're monitoring one instance in an account with a large number of instances (since `DescribeDBInstances` will be paginated incurring a call per page). That seems pretty unlikely, but it could be relevant in a trial situation. We should at least note it somewhere.

util/awsutil/amazon.go
lfittl: Good point - I think we can actually make this more clear by having `GetAwsConfig` take an `AccountKey` - that way its enforced.

util/awsutil/ttl_cache.go
lfittl: I think if we find additional callers we could make it configurable, but it didn't seem worth the trouble for now.

util/awsutil/ttl_cache.go
lfittl: Yeah, this one feels less important than the other one in terms of being able to configure it, and I don't think we need to make it configurable without a direct reason to do so.

input/system/rds/logs.go
lfittl: Yeah, I'll drop this, I think its not needed anymore to note here.

util/awsutil/rds.go
lfittl: I think its unlikely the caller would scribble on the result (at least none of the current callers did), and its not free to copy here (though the hit might be acceptable). For now I think the comment is the better trade-off, though I really wished Go had a better concept of constness for results.

util/awsutil/ttl_cache.go
lfittl: I think its what would have happened previously (you would have seen a failure, not a stale result), and I'm slightly worried that stale results introduce an unknown kind of problem that is harder to spot, e.g. when trying to figure out why a primary target didn't update after failover. For now I'll keep as is, and we could reconsider if errors are common enough that this would be helpful.

PR #864, opened 2026-09-02
Make Aurora max storage either 128 or 256 TB based on Aurora version

### LINKED ISSUE #858: AuroraMaxStorage is inaccurate
AuroraMaxStorage value is no longer inaccurate and leads to bad data when storage surpasses the fixed value.
 
The value is currently hardcoded to 128TB (https://github.com/pganalyze/collector/blob/9c83e1deb5a0a06d1c7ff1ba7479e1a77cc477e8/input/system/rds/system.go#L19)  however Aurora now allows more storage depending on pg engine version. 

From the [Aurora docs](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/CHAP_Limits.html#RDS_Limits.FileSize.Aurora)
```
256 TiB maximum:
  Aurora PostgreSQL versions:
  17.5 and all higher versions
  16.9 and higher
  15.13 and higher

128 TiB maximum:
  All earlier Aurora PostgreSQL versions
```

lfittl:
Fixes #858

### DIFF
--- a/input/system/rds/system.go
+++ b/input/system/rds/system.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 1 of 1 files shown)


keiko713 (approved):
It would be nice to have a test but looking good!

lfittl:
> It would be nice to have a test but looking good!

I feel like the logic here is simple enough that a test is unnecessary, so I'll skip for now.

PR #867, opened 2026-09-02
Fix some concurrency hazards

msakrejda:
See individual commits:

- Reconnecting socket: Don't block in Disconnect after context cancellation
- Scheduler: Enforce the run time limit for secondary schedules too
- Query runs: Read the query run map while holding the mutex


lfittl (approved):
👍 Changes make sense, also on doing a second pass.

PR #868, opened 2026-09-03
Introduce a new way to run Docker image release

keiko713:
Currently, docker image release step is pretty manual - launch the EC2 instance and do some prep, then wait for it to finish.
While I was trying to release 0.73.0, I was having a problem with this, where the `make docker_release` command hung due to some deadlock and never finish. This `docker_release` was also using the emulator to build the arm image, which was taking extra time too.

To make this step better, have the current release GH action build the docker images (triggered when we push the tag like `v0.73.0`), then the operator (aka some folks from pganalyze) will publish that image to our quay.io repo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ ifeq ($(shell uname), Darwin)
 endif
 PROTOC_URL := $(PROTOC_BASE_URL)v$(PROTOC_VERSION_NEEDED)/$(PROTOC_FILENAME)
 
-.PHONY: default build build_dist vendor test docker_release packages integration_test
+.PHONY: default build build_dist vendor test docker_image_amd64 docker_image_arm64 docker_release_staging docker_release_manifest packages integration_test
 
 default: build test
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -68,15 +68,51 @@ packages:
...
+	$(call docker_build_oci,arm64,$(DOCKER_OCI_ARM64))
+
+# Pushes the OCI archives downloaded from the GitHub release into the
+# collector-staging repository in quay.io, which is where docker_release_manifest
+# reads them from.
+docker_release_staging:
+	@test -n "$(DOCKER_RELEASE_TAG)" || (echo "ERROR: DOCKER_RELEASE_TAG is not set, make sure you are on a git release tag or override by setting DOCKER_RELEASE_TAG" ; exit 1)
+	@test -f $(DOCKER_OCI_AMD64) || (echo "ERROR: $(DOCKER_OCI_AMD64) not found, download it with: gh release download $(DOCKER_RELEASE_TAG) -p '*.oci.tar'" ; exit 1)
+	@test -f $(DOCKER_OCI_ARM64) || (echo "ERROR: $(DOCKER_OCI_ARM64) not found, download it with: gh release download $(DOCKER_RELEASE_TAG) -p '*.oci.tar'" ; exit 1)
+	skopeo copy --multi-arch all oci-archive:$(DOCKER_OCI_AMD64) docker://$(DOCKER_STAGING_IMAGE):$(DOCKER_RELEASE_TAG)-amd64
```
keiko713, line 102: Note: skopeo is used to push OCI to the collector-staging repo in quay.io. Note this `--multi-arch` is needed to copy every manifest in the index and preserves the index itself, so the host's own architecture (e.g. my local machine's arch) stops mattering. This would also help to keep the provenance between CI and Quay.

Makefile
```diff
@@ -68,15 +68,51 @@ packages:
...
+endef
+
+docker_image_amd64:
+	$(call docker_build_oci,amd64,$(DOCKER_OCI_AMD64))
+
+docker_image_arm64:
+	$(call docker_build_oci,arm64,$(DOCKER_OCI_ARM64))
+
+# Pushes the OCI archives downloaded from the GitHub release into the
+# collector-staging repository in quay.io, which is where docker_release_manifest
```
keiko713, line 96: Pushing to the collector-staging repo (TODO: create this) will help us not to push these interim images in the main collector repo. We could safely remove these images from the collector-staging repo also later on if we wish so.

lfittl: I'm a little bit confused - can you clarify why we need the staging repository? It seems that we're only pushing them up in the `docker_release_staging` step to then pull them down again in `docker_release_manifest`. Couldn't we tell skopeo to tag the image but not push it?

lfittl (commented):
Overall this looks good, but I'm a little confused as to the need of the staging repository (see my comment)

Makefile
keiko713: Thanks for pushing back! Digging into it, I can avoid the use of the staging repo. I've pushed a change to drop it.
The change is not really tell skopeo about it, since the previous way used `docker buildx imagetools create`, and that only accepts registry references as input. It can't read a local OCI archive, so both per-arch images had to exist somewhere in quay before the manifest list could be created (which is why we had to push them to quay.io, which I used collector-staging repo).

So instead of trying to tweak things around there, I dug more and found `podman manifest` can assemble the index locally from the two `.oci.tar` files, so we push once, straight to the release tags.

Can you please take another look, especially at the Makefile target and the updated CONTRIBUTING steps?

Makefile
```diff
@@ -68,15 +68,50 @@ packages:
...
+define docker_build_oci
+docker buildx inspect $(DOCKER_BUILDER) > /dev/null 2>&1 || docker buildx create --name $(DOCKER_BUILDER) --driver docker-container
+	docker buildx build --platform linux/$(1) --builder $(DOCKER_BUILDER) \
+	--provenance=true \
+	--output type=oci,dest=$(2) \
+	.
+	docker buildx rm $(DOCKER_BUILDER)
+endef
+
+docker_image_amd64:
```
lfittl, line 89: Similarly, I wonder if we should call these "container_image_..." and  rename the existing task from "docker_release" to "container_release".

Makefile
```diff
@@ -68,15 +68,50 @@ packages:
 	make -C packages
 
 DOCKER_RELEASE_TAG := $(shell git describe --tags --exact-match --abbrev=0 2> /dev/null)
+DOCKER_IMAGE := quay.io/pganalyze/collector
+DOCKER_BUILDER := collector-build
+DOCKER_OCI_AMD64 := pganalyze-collector-image-linux-amd64.oci.tar
+DOCKER_OCI_ARM64 := pganalyze-collector-image-linux-arm64.oci.tar
+DOCKER_MANIFEST := pganalyze-collector-release
```
lfittl, line 75: I wonder if we should rename all these variables to start with "CONTAINER_" instead of "DOCKER_", since its really about building a container image, not calling a docker command.

Makefile
```diff
@@ -68,15 +68,50 @@ packages:
...
+	@test -f $(DOCKER_OCI_ARM64) || (echo "ERROR: $(DOCKER_OCI_ARM64) not found, download it with: gh release download $(DOCKER_RELEASE_TAG) -p '*.oci.tar'" ; exit 1)
+	-podman manifest rm $(DOCKER_MANIFEST) 2> /dev/null
+	podman manifest create $(DOCKER_MANIFEST)
+	podman manifest add --all $(DOCKER_MANIFEST) oci-archive:$(DOCKER_OCI_AMD64)
+	podman manifest add --all $(DOCKER_MANIFEST) oci-archive:$(DOCKER_OCI_ARM64)
+	# Expect four entries: one image and one attestation per architecture
+	podman manifest inspect $(DOCKER_MANIFEST)
+	podman manifest push --all $(DOCKER_MANIFEST) docker://$(DOCKER_IMAGE):$(DOCKER_RELEASE_TAG)
+	podman manifest push --all $(DOCKER_MANIFEST) docker://$(DOCKER_IMAGE):latest
+	podman manifest push --all $(DOCKER_MANIFEST) docker://$(DOCKER_IMAGE):stable
```
lfittl, line 113: We can do this as a follow-up (and after the next release), but it'd be helpful to think through how we could call this workflow with a custom tag to be uploaded - I've often built custom container images of a pre-release commit to fix a customer's issue by manually doing the docker release steps - but now a few more steps will be involved.

lfittl (approved):
Looks good!

Makefile
lfittl: Nice!

Makefile
keiko713: Soooo I thought about this but I'd rather keep `DOCKER_`. It's the prefix we use in a lot of places (e.g. `integration_test/Makefile`, `packages/*/Makefile`), so switching just these would leave us with two prefixes for the same thing (the same goes with `docker_*`).
Then, we're gonna face the problem of... how much we're going to replace `docker` to `container`, which I kinda don't wanna deal with in this PR. I think lots of people use docker/container interchangeably and it is _not wrong_ that we're making the docker image.

I thought about renaming above 3 to `OCI_AMD64` / `OCI_ARM64` / `RELEASE_MANIFEST` but then it is kinda nice to have the same prefix also 🤔 What do you think?

Makefile
keiko713: Ah, very good call. Uhmm, it is a problem indeed - the current flow is very much tag/release focused, and built OCI tar is attached to the release, so... it's not straight forward to ask CI to build images for us and we can push this manifest to the specific tag.
I'll start thinking about this.

Makefile
lfittl: Yeah, I feel like we could do that separately and just use DOCKER_ here for now.

PR #869, opened 2026-09-08
Stop supporting RHEL7, Debian 11 and Ubuntu 20.04 and raise minimum glibc

lfittl:
This drops support for distributions that are now end of life, and that we can no longer support due to build processes failing:

- Amazon Linux 2
- RHEL7
- Debian Bullseye (11)
- Ubuntu Focal (20.04)

Note that some of these distributions still receive updates on specific paid support plans (e.g. RHEL7 until May 2029 with ELS), but the effort to continue supporting them does not seem warranted at this point.

This also raises the glibc requirement to 2.28 (RHEL-based) and 2.34 (Debian-based) and adds an explicit check for glibc compatibility in the package build process.


contrib/install.sh
```diff
@@ -102,7 +96,7 @@ then
   pkg=deb
   distribution=ubuntu
   version=$(grep VERSION_CODENAME /etc/os-release | cut -d= -f2)
-  if [ "$version" != noble ] && [ "$version" != jammy ] && [ "$version" != focal ];
+  if [ "$version" != noble ] && [ "$version" != jammy ];
```
keiko713, line 99: Can be with a different PR but we could add the support for `resolute` at some point too 🦝

lfittl: Yeah, I think that can be a separate PR - we should also add Amazon Linux 2027 which just started being in preview.

packages/repo/Dockerfile.repo-rpm
```diff
@@ -1,4 +1,4 @@
-FROM amazonlinux:2
+FROM rockylinux:8
```
keiko713, line 1: My cool friend (aka Claude) told me that it might be better to point to `rockylinux/rockylinux:8` instead.
https://hub.docker.com/_/rockylinux says:

> The Docker team curates the Official Images program, and there are currently some technical constraints preventing Rocky Linux from publishing updates here. For the most up-to-date container images, please refer to the [Rocky Linux Docker Hub repository](https://hub.docker.com/r/rockylinux/rockylinux) for now.

And that "Rocky Linux Docker Hub repository" is `rockylinux/rockylinux`.

```suggestion
FROM rockylinux/rockylinux:8
```

keiko713: Uhmm, but looks like we do use `rockylinux:8` several other places so it's okay to leave as is.

keiko713: Ok, so the other thing my cool friend told me is, this will change the RPM signing container from amazonlinux:2 to rockylinux:8, the used gpg will be from 2.0 to 2.2.

The `--passphrase-fd 3` in below line 18 and the `--edit-key` batch script in `sync_rpm.sh` both date from the gpg 2.0 era.

Starting from GnuPG 2.1, the secret-key handling moved out of gpg and into gpg-agent. As a consequence, gpg stopped honoring the `--passphrase-fd` / `--passphrase-file` / `--passphrase` options. So it reads the fd, then ignores the content and asks pinentry instead.

In fact, I actually tested if it fails with the current command locally with some minimal repro docker image with `--passphrase-fd 3`, and it failed with the following since rpm ≥ 4.12 handed passphrase handling to gpg-agent and stopped opening fd 3.

```
gpg: Fatal: passphrase-fd is invalid: Bad file descriptor
error: gpg exec failed (2)
```

We actually should be able to simplify the line 17 part to something like the following:

```
RUN echo "%_gpg_digest_algo sha256" >> /root/.rpmmacros
```

Anyways, maybe better for you to look into this by yourself rather than me trying to understand things here and suggest some options. Happy to discuss further if you need my inputs but I don't have much knowledge around here so maybe better for you to chat with Claude 👼

lfittl: Good catch with the passphrase - I didn't think of that (or think of prompting Claude the right way I suppose 😉) - let me look into that.

packages/repo/Dockerfile.repo-rpm
```diff
@@ -8,18 +8,16 @@ ENV RPM_DIR /rpm
...
 RUN mkdir -p $RPM_DIR/systemd
 
-RUN yum install -y -q rpm-sign createrepo
+RUN dnf install -y -q rpm-sign createrepo_c gnupg2
 
 RUN echo "%_gpg_name team@pganalyze.com" > /root/.rpmmacros
 
-# Switch signature algorithm from SHA1 to SHA256 to support RHEL9 and newer
+# Sign with SHA256 instead of SHA1 to support RHEL9 and newer
 # See https://www.redhat.com/en/blog/rhel-security-sha-1-package-signatures-distrusted-rhel-9
```
keiko713: Now this link might be just confusing because it's talking about the workaround command we had before. So, maybe something like:

```
# Pin SHA256 digests, since RHEL 9 and newer reject SHA-1 signed packages. gpg 2.2 already
# defaults to SHA256, so this only makes the requirement explicit -- unlike the older
# %__gpg_sign_cmd override in the post below, which was needed with RHEL 7-era gpg 2.0.
# See https://www.redhat.com/en/blog/rhel-security-sha-1-package-signatures-distrusted-rhel-9
```

Or even dropping the link:

```
# Pin SHA256 digests, since RHEL 9 and newer reject SHA-1 signed packages. gpg 2.2 already
# defaults to SHA256, so this only makes the requirement explicit
```

keiko713 (approved):
Once comment on comment but otherwise looking good!

PR #872, opened 2026-09-10
Reconnecting socket: Bound connection attempts by the caller's context

lfittl:
EnsureGrant blocked on WebSocket.Connect() until the dial finished, and the dial was only bounded by the socket's process-wide context. With the dialer's handshake timeout in place (#871) this is capped at 30 seconds per attempt, but a grant call can still wait for a background reconnect attempt to finish before its own is even started, and the run's own deadline can't interrupt either of them. For an unreachable WebSocket endpoint that means every full snapshot and test run stalls for up to a minute before the first database query, even though the HTTP fallback could serve the grant right away.

Connect() now takes a context that bounds the connection attempt. Each request carries its own context and result channel, replacing the shared start/startWait pair, so callers can't receive each other's results and a caller that gives up doesn't leave a stale result behind. An established connection keeps living on the socket's context. EnsureGrant uses this to wait at most 10 seconds for the WebSocket when api_require_websocket is off, then falls back to the HTTP-based grant while the reconnect loop keeps trying in the background.

In passing fix the 4xx back-off never engaging: connect() returned status 0 on every error, so the 8 minute delay after a 4xx handshake response was dead code. The status is now returned on error and included in the warning.

### DIFF
--- a/util/reconnecting_socket.go
+++ b/util/reconnecting_socket.go
@@ -28,8 +28,7 @@ type ReconnectingSocket struct {
 	ctx       context.Context
 	requested atomic.Bool
 	conn      atomic.Pointer[websocket.Conn]
-	start     chan struct{}
-	startWait chan error
+	start     chan connectRequest
 	shutdown  chan struct{}
 }
 
@@ -38,6 +37,16 @@ type ReconnectingSocket struct {
 type socketWrite struct {
 	data []byte
 	// Must be buffered (capacity 1), so the writer goroutine never blocks
+	// handing back the result to a caller that has stopped waiting
+	result chan error
+}
+
+// connectRequest - A single request to establish the connection, together with
+// the channel used to hand the outcome back to the caller
+type connectRequest struct {
+	// Bounds the connection attempt made on behalf of this request
+	ctx context.Context
+	// Must be buffered (capacity 1), so the manager goroutine never blocks
 	// handing back the result to a caller that has stopped waiting
 	result chan error
 }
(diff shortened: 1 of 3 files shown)


util/reconnecting_socket.go
```diff
@@ -200,19 +222,37 @@ func (w *ReconnectingSocket) Disconnect() {
...
 	if err != nil {
-		cancelConn()
-		w.logger.PrintWarning("Error starting websocket: %s %v", err, response)
-		return 0, err
+		if response != nil {
+			w.logger.PrintWarning("Error starting websocket: %s (HTTP status %d)", err, connectStatus)
+		} else {
+			w.logger.PrintWarning("Error starting websocket: %s", err)
+		}
+		return connectStatus, err
```
keiko713, line 249: > In passing fix the 4xx back-off never engaging: connect() returned status 0 on every error, so the 8 minute delay after a 4xx handshake response was dead code. The status is now returned on error and included in the warning.

I believe that now this is actually returning some status so it can enforce the `skipConnectUntil` for 8 mins.

My understanding is that the 8 minutes was meant for the reconnect loop: if the WebSocket handshake comes back 4xx, don't keep retrying every 60 seconds. This is likely aiming for the full-snapshot, and the next full snapshot re-dials and recovers. That part makes sense.

But `skipConnectUntil` is checked in the manager goroutine, which also serves caller-initiated `Connect()`, so this now applies to activity snapshots (every 10s) and log sends too. This means one 4xx aborts every snapshot (including activity/logs) for the full 8 mins, instead of previous generous 60 seconds redial attempts.
This would be only true when `api_require_websocket: true` (otherwise it'll fallback to HTTP), but as we wanna move to websockets more near future, just wanted to double check if this is what we really want.

That said, I think the common 4xx case is actually the invalid API key use (rather than some server side hiccup that might be fixed promptly), so maybe it's okay to wait for 8 mins for everything for that case.

keiko713 (approved):
One minor comment (but while I was writing, I kinda concluded that we're okay with it), looking good!

util/reconnecting_socket.go
lfittl: Yeah, I think the argument for doing it this way is that you shouldn't be getting a 4xx for server side or network issues (that'd be a 5xx) - 4xx should only happen when the client sent a bad request, i.e. bad authentication data in practice. Since we now have the status code in the warning, it should be easy to see if this becomes a problem in practice (I don't think it will).

PR #878, opened 2026-09-21
Support db_use_iam_auth for Azure Database for PostgreSQL

ryanbooz:
(Sorry - branch says PoC (it was originally) but I didn't feel it was worth renaming for this small change.)

Add an azure_database case to getIamConnectionParams that fetches an Entra ID (AAD) access token via the same credential chain already used for Event Hub/ARM (client secret, client certificate, workload identity, or managed identity) and uses it as the connection password, scoped to https://ossrdbms-aad.database.windows.net/.default.

Tested against newly spun up Azure Flexible server instance with Managed Identity login mapped as a non-admin user.


input/system/azure/iam.go
```diff
@@ -0,0 +1,37 @@
...
+// server and flexible server) Entra ID (AAD) authentication.
+//
+// See https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/how-to-configure-sign-in-azure-ad-authentication
+const dbAuthScope = "https://ossrdbms-aad.database.windows.net/.default"
+
+// GetDbAuthToken fetches an Entra ID (AAD) access token to use as the
+// password when connecting to Azure Database for PostgreSQL with IAM auth
+// (i.e. db_use_iam_auth). This supports the same credential sources as our
+// Event Hub / ARM integrations: client secret, client certificate, workload
+// identity, or (the common case for PoC/testing) VM/AKS managed identity.
```
keiko713: This comment can be improved a bit maybe. I think the second sentence is not adding much so maybe it can be simplified to:

```
// GetDbAuthToken fetches an Entra ID (AAD) access token to use as the
// password when connecting to Azure Database for PostgreSQL with IAM auth.
```

keiko713 (approved):

input/system/azure/iam.go
ryanbooz:
```suggestion
// GetDbAuthToken fetches an Entra ID (AAD) access token to use as the
// password when connecting to Azure Database for PostgreSQL with IAM auth.
```

